Issue #32, opened 2018-10-04
Support multiple OpenFOAM versions with the same source


MakisH:
Often users are bound to specific OpenFOAM variants (e.g. openfoam.com, openfoam.org, foam-extend) and versions (not always the latest, sometimes very old). However, often small or not-so-small differences will give errors during compilation. Therefore, we need a system to be able to support different versions at compile time.

Implementing such a system would help a lot dealing with issues such as #21, #8, #9, #26. A list of currently supported versions can be found in our [wiki](https://github.com/precice/openfoam-adapter/wiki/Notes-on-OpenFOAM).

Fortunately, all the three main variants set the environment variables `WM_PROJECT` and `WM_PROJECT_VERSION` (see also #29 ).

We currently use the WMake build system, developped together with OpenFOAM. This is a quite primitive system, but it is intrinsic to OpenFOAM. On top of it, I have added the [Allwmake](https://github.com/precice/openfoam-adapter/blob/master/Allwmake) bash script to automate the procedure even more (also a common way in OpenFOAM-derived projects).

A few ideas on how to go on:

1. (ugly) Support different versions in different branches. The current state of the branch works with the latest version, Git tags/other branches for older versions.
    * This is how we support multiple [SU](https://github.com/precice/su2-adapter) and [CalculiX](https://github.com/precice/calculix-adapter) versions.
    * Difficult to maintain, but easier than other options (e.g. different directories)
    * Here we have multiple "latest versions".

2. (ugly) Use pre-processor commands to check the version and activate different parts of the code.
    * A bit ugly and maybe difficult to maintain
    * Some changes (e.g. required in #8 and #9 ) are much bigger and spread throughout the code to treat with small patches

3. (ugly) Have source files like `Adapter_of6_dev.C` and `Adapter_of4_5.C`  and let our script pick the correct ones.
    * Difficult to maintain
    * Keep the files kind of clean

4. __Use a proper configuration system (e.g. CMake)__
    * Ok, but how to solve our problem with it?
    * Do we need to replace WMake completely or can it work on top of it?

I would very much value your feedback on this.

derekrisseeuw:
Hi Makis,

In my opinion the ideal openfoam adapter should be as portable as possible and easy to use. Therefore, the first solutions does not really work, because the user has to select the right branch. 

The fourth option is the most elegant solution, but has a major flaw in my opinion. If the user wishes to support multiple openfoam versions, this approach does not work anymore.

[...]

Issue #21, opened 2018-06-26
Port Foam::Time::setDeltaT to new signature


MakisH:
On February 5, 2018, [this commit](https://github.com/OpenFOAM/OpenFOAM-dev/commit/139ffcc112ae49d60d8d9f8b0c8c2d7b6315268e) to openfoam-dev broke the compatibility with the OpenFOAM adapter, as the method `Foam::Time::setDeltaT` now uses only one argument instead of two, changing a bit its logic.

We need to find a way to support both the old and the new signature and to check how the logic changes affect the adapter.

MakisH:
This is treated in #33, will not be merged anytime soon (but treated in #32 )

PR #33, opened 2018-10-04
[Will not merge] Compatibility with OpenFOAM 6

MakisH:
This is work in progress, related to #21.

- [x] Replaced setDeltaT(..., false) with the new method setDeltaTNoAdjust(...).
- [x] [OK] Test with fixed time step and serial-explicit
- [x] [OK] Test with fixed time step and serial-implicit
- [x] [PROBLEM] Test with adjustable time step and serial-explicit --> it does not output the final time (1s) for Fluid (it does for Solid) with neither `writeControl runTime` nor `writeControl adjustableRunTime` (regression from OpenFOAM 5.x)
- [x] [PROBLEM] Test with adjustable time step and serial-implicit --> Same problem as with explicit
- [ ] Test for corner cases and try to break it
- [ ] Find a way to integrate it in the main branch without breaking backwards-compatibility (see #32)


MakisH:
I just merged the latest state of `develop` (so, including FSI) and it compiles&runs with OpenFOAM6. I have not yet looked into the aforementioned timestep-related issue, though.

PR #36, opened 2018-10-05
[Will not merge] Compatibility with OpenFOAM-dev

MakisH:
This pull request keeps track of changes required for OpenFOAM-dev. ~~It will be merged when OpenFOAM-dev becomes the next stable version (probably OpenFOAM7).~~

Edit: this PR will remain open and will always track the changes in OpenFOAM-dev, until we find a good way to support multiple versions with the same source #32.


MakisH:
I just merged the latest state of `develop` (so, including FSI) and it complies&runs successfully with OpenFOAM-dev `20190109`.

PR #47, opened 2018-11-11
Update to the FSI branch

derekrisseeuw:
A large pull request for the FSI part of the openfoam adapter. Implemented is: 

- removed the obsolete velocity class
- checkpointing for all <vol,surface,point<scalar,vector,tensor>field, and the volSymmTensor field. 
- checkpointing for the old fields (oldTime()) for correct time integration.
- checkpointing for the mesh flux

Small part on a proposal for some of parts of them mesh which will be required when subcycling the fluid solver.


MakisH:
The target branch of this should actually be `precice:FSI`, but I am not sure if I can change it...

derekrisseeuw:
You are right, I changed it.

PR #51, opened 2018-12-14
FSI: Fix devRhoReff and mu to support turbulence

efirvida:
This is based on the libforce functionObject, also I put the tutorials from here: precice/tutorials#14


MakisH:
We now support force calculations for compressible solvers as implemented at the same time in #64, which was a bit smaller and easier to merge. However, thank you again very much for submitting this! If there is something more here, please open new, smaller Pull Requests with a few more details in the description and we would be happy to discuss them!

Issue #57, opened 2018-12-20
FSI: Extend Force to turbulent flows


MakisH:
Currently, in `FSI/Force.C`, we only support laminar flows (see `devRhoReff()`).

No warning is currently implemented.

More details to follow.

See also #42. Relevant to #7.

derekrisseeuw:
Check out the fork of [efirvida ](https://github.com/efirvida/openfoam-adapter), I think he implemented most of the different types of forces already.

MakisH:
This has been closed through #64.

Issue #42, opened 2018-10-31
FSI: Support compressible solvers


MakisH:
At the moment, we compute forces in `FSI/Force.C` (at the `FSI` branch for now) assuming a constant density and viscosity. In order to support compressible solvers as well, we need to abstract these in a way similar to the [KappaEffective](https://github.com/precice/openfoam-adapter/blob/master/CHT/KappaEffective.C) for CHT.

An open question here is how to validate the compressible FSI, as I am not aware of any literature on this at the moment. In other words, we are looking for something similar to the Turek-Hron FSI benchmark for compressible solvers.

We already had some discussion on this with @JSeuffert. Do you maybe have anything to add here?

This is relevant to #7.

JSeuffert:
Thanks for making this an issue and I totally agree with you @MakisH .

The compressible solvers should all have a volScalarField called “thermo:mu” for the dynamic viscosity. So in my opinion the best would be to read this at the registry. Also the density should be read from the registry and not from the dictionary. 
I dont know the best solution how to structure it but maybe we can do it like the heat flux in the CHT module by implementing Force_incompessible and Force_compressible?

Regarding the verification of the compatibility of the module to different solver types, we could firstly use the Hron-Turek benchmark and solve it quasi-incompressible with a compressible solver and even with the multiphase solvers. 
At the moment I also do not know a specific compressible FSI benchmark, but I will do some more research.

derekrisseeuw:
At the moment the code in the force class is very similar to the one in the openfoam function, but only the incompressible parts from that code are extracted. I think it would be quite straightforward to implement more of these code. 

See the Openfoam [forces.c](https://github.com/OpenFOAM/OpenFOAM-5.x/blob/7f7d351b741bf6406366a043cac98de56d2d44dd/src/functionObjects/forces/forces/forces.C)

JSeuffert:
You are right @derekrisseeuw. That could be much easier. Thanks for this information!

MakisH:
This is now supported on master.

PR #64, opened 2019-01-22
Forces calculation for compressible, multiphase and turbulence

JSeuffert:
Concerning issues #57 #42 
Maybe in parts the same as pull request #51 

Forces calculation is implemented as in the forces functionObject.

I tested the correct implementation with: pimpleDyMFoam, rhoPimpleDyMFoam, interDyMFoam, compressibleInterDyMFoam; laminar flow and RAS turbulence model
Tested with Openfoam 4 (https://github.com/OpenFOAM/OpenFOAM-4.x), precice 1.3.0, Calculix 2.13

Forces calculation itself is not verified!


derekrisseeuw:
Incredible!! 
The verification of the incompressible variant of the forces is done in my thesis, which will be available shortly.

MakisH:
Thank you very much, @JSeuffert! Is this ready to review? I will try to have a look next week.

JSeuffert:
I just deleted my last two commits that should not be a part of this pull request. 
It's now ready to review!

FSI/Force.H
```diff
@@ -5,6 +5,13 @@
 
 #include "fvCFD.H"
 
+// TEMPORARY
```
MakisH: Maybe check?

PR #65, opened 2019-01-28
Add travis integration

shkodm:
As  described in detail in this  [pull request]( https://github.com/precice/systemtests/pull/22 ) for systemtests, it is now possible for an adapter to trigger custom set of tests on the [systemtests](https://github.com/precice/systemtests) repository. Therefore I add simple Travis set up that will test OpenFOAM-OpenFOAM coupling and OpenFOAM-CalculiX coupling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+sudo: true
+dist: trusty
+language: python
+python:
+  - "3.5"
+jobs:
+  include:
+    # trigger systemtests build only when pushing to master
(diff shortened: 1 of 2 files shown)


shkodm:
Travis fails for now, but  this is because of the `TRAVIS_ACCESS_TOKEN`, that was not added when the PR was opened. I added it now, but somehow it is still not picked up in the PR build. But I think it is safe to merge, the master build should succeed.

PR #88, opened 2019-07-01
WIP: Fluid-Fluid module with cell sets

martacamps:
Adds Fluids module, which allows to exchange velocity and pressure / pressure gradient.

It reads from / writes to patches and previously defined cellSets. Both the name of the patches and cellSets are set in the adapter .yml configuration file.

It also corrects the x component of the read velocity so that the mass flow rate is the one set in the .yml configuration file.

Note: the code needs cleaning and it is currently tailored to one concrete application.


martacamps:
Is the same code I did the pull request for during Coupled2019 conference. I just renamed the branch and I realized that the previous pull request was automatically closed.

MakisH:
Thank you once more for contributing, I hope this has already been helpful to others! As this is now a bit behind to be easily merged, I will close it for now to keep a better overview, but feel free to revive it later. :smiley:

PR #96, opened 2019-08-14
Draft for 2D coupling with OpenFOAM 

### LINKED ISSUE #94: Coupling OpenFOAM to 2D solvers could be easier
# Context

We currently have several users that couple OpenFOAM (which only works in 3D) with 2D solvers. We usually treat this case as follows:
* Use a 3D solver interface: `<solver-interface dimensions="3">`.
* In the 3D solver, use only one cell in the out-of-plane direction (pseudo-2D) and define the interface as points on a line.
* In the 2D solver, give zero values for the out-of-plane field components (e.g. force in z-direction).

Examples are e.g.:
* [Tutorial for FSI with deal.II and OpenFOAM](https://github.com/precice/precice/wiki/Tutorial-for-FSI-with-deal.II-and-OpenFOAM)
* [Tutorial for CHT: Flow over a heated plate with OpenFOAM and FEniCS](https://github.com/precice/precice/wiki/Tutorial-for-CHT:-Flow-over-a-heated-plate-with-OpenFOAM-and-FEniCS)

# Problem

This approach has the following problems:
* For a face-nodes-based mesh, one point of the 2D solver has two equally-distanced neighbors on the 3D mesh (in most cases). This makes the nearest-neighbor mapping to act arbitrarily.
* RBF mapping, which is the usual workaround to the equal-distance issue, [needs to be configured to set a dead dimension](https://github.com/precice/precice/wiki/Mapping-Configuration#radial-basis-function-mapping).
* Further numerical issues with the coupling schemes (@uekerman).
* If a solver needs to assume the dimension of its partner (potentially scaling any received data), then the black-box assumption of preCICE weakens.

# Proposed solution

We are working on making preCICE compatible with mixed dimensions, but this feature will not be available anytime soon. Also, OpenFOAM is clearly the solver of choice of our users for the role of the Fluid participant, right now.

As an easier solution, @uekerman suggested that we have an option in the OpenFOAM adapter to define 3D or 2D preCICE interfaces. This would require multiple changes in the code, for which @DavidSCN already did some research.

Related to #30.

DavidSCN:
Remaining questions:
How to proceed with forces scaling?

Do we need to support meshConnectivity?
Current use case for 2D coupling is FSI, where meshConnectivity is not needed and implemented on the Fluid side.

Closes #94

### DIFF
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -24,7 +24,7 @@ preciceAdapter::CHT::HeatFlux::HeatFlux
     dataType_ = scalar;
 }
 
-void preciceAdapter::CHT::HeatFlux::write(double * buffer, bool meshConnectivity)
+void preciceAdapter::CHT::HeatFlux::write(double * buffer, bool meshConnectivity, const unsigned int dim)
 {
     int bufferIndex = 0;
 
@@ -79,7 +79,7 @@ void preciceAdapter::CHT::HeatFlux::write(double * buffer, bool meshConnectivity
     }
 }
 
-void preciceAdapter::CHT::HeatFlux::read(double * buffer)
+void preciceAdapter::CHT::HeatFlux::read(double * buffer, const unsigned int dim)
 {
     int bufferIndex = 0;
 
--- a/CHT/HeatFlux.H
+++ b/CHT/HeatFlux.H
@@ -35,11 +35,11 @@ public:
 
     //- Compute heat flux values from the temperature field
     //  and write them into the buffer
-    virtual void write(double * buffer, bool meshConnectivity);
+    virtual void write(double * buffer, bool meshConnectivity, const unsigned int dim);
 
     //- Read heat flux values from the buffer and assign them to
     //  the gradient of the temperature field
-    virtual void read(double * buffer);
+    virtual void read(double * buffer, const unsigned int dim);
 
     //- Destructor
     virtual ~HeatFlux(){};
--- a/CHT/HeatTransferCoefficient.C
+++ b/CHT/HeatTransferCoefficient.C
@@ -26,7 +26,7 @@ preciceAdapter::CHT::HeatTransferCoefficient::HeatTransferCoefficient
 }
 
 
-void preciceAdapter::CHT::HeatTransferCoefficient::write(double * buffer, bool meshConnectivity)
+void preciceAdapter::CHT::HeatTransferCoefficient::write(double * buffer, bool meshConnectivity, const unsigned int dim)
 {
     int bufferIndex = 0;
 
@@ -77,7 +77,7 @@ void preciceAdapter::CHT::HeatTransferCoefficient::write(double * buffer, bool m
 }
 
 
-void preciceAdapter::CHT::HeatTransferCoefficient::read(double * buffer)
+void preciceAdapter::CHT::HeatTransferCoefficient::read(double * buffer, const unsigned int dim)
 {
     int bufferIndex = 0;
 
--- a/CHT/HeatTransferCoefficient.H
+++ b/CHT/HeatTransferCoefficient.H
@@ -39,10 +39,10 @@ public:
     );
 
     //- Write the heat transfer coefficient values into the buffer
-    virtual void write(double * buffer, bool meshConnectivity);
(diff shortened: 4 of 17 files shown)


MakisH:
After discussion, we agree on the following:

* We want to have both a 2D domain and 2D fields.
* We don't provide mesh connectivity for FSI in 3D, we are not going to provide it in 2D either.
* We always assume that z is the out-of-plane dimension.
* For CHT, we have two options:
   * Not provide mesh connectivity, since the edges connecting the mesh centers do not exist in the mesh and could be confusing
   * Provide mesh connectivity but inform the user that these are artificial edges.
   Since already we create artificial edges when we triangulate the mesh, we decide to go with this option.
* We are not going to scale the forces in the adapter, but ask the user to take care of setting an appropriate width (either unit or same as the one modeled in the other domain).

We also decide to write a tutorial about this. It is a very common use case.

DavidSCN:
> We also decide to write a tutorial about this. It is a very common use case.

Can you give more information about the tutorial? Which case and solvers?

> Provide mesh connectivity but inform the user that these are artificial edges.
Since already we create artificial edges when we triangulate the mesh, we decide to go with this option.

Probably, we need to restrict this case to a real 3D case with `z-axis-dead=true`.

Apart from meshConnectivity (still discussing), this PR is ready.

MakisH:
> > We also decide to write a tutorial about this. It is a very common use case.
> 
> Can you give more information about the tutorial? Which case and solvers?

I think the most common would be an FSI case. I would use either deal-ii or FEniCS as a structure solver (whatever is easier) and OpenFOAM as a fluid. For deal-ii you already have experience, for FEniCS @richahert and @VojtechKubac were working on a structure solver.

As a test case I would reuse the perpendicular flap we already have.

Later we could also provide a CHT case to also demonstrate the mesh connectivity feature.

> > Provide mesh connectivity but inform the user that these are artificial edges.
> > Since already we create artificial edges when we triangulate the mesh, we decide to go with this option.
> 
> Probably, we need to restrict this case to a real 3D case with `z-axis-dead=true`.

Why not implement the edges differently for 2D meshes? It will complicate the adapter more, for sure, but depends if we actually need it.

I would say let's just ignore the mesh connectivity in this PR and revisit it later.

DavidSCN:
> Why not implement the edges differently for 2D meshes? It will complicate the adapter more, for sure, but depends if we actually need it.

That's another option of course.

> I would say let's just ignore the mesh connectivity in this PR and revisit it later.

Ok, should we add a warning somewhere?

MakisH:
> > I would say let's just ignore the mesh connectivity in this PR and revisit it later.
> 
> Ok, should we add a warning somewhere?

Definitely yes.

BenjaminRueth:
@richahert implemented the 2D-3D mapping for his experiments directly in the adapter:

https://github.com/precice/fenics-adapter/blob/721dac5a282bf833af436356355a131da84f8663/fenicsadapter/fenicsadapter.py#L326-L336

If the OF solver is solving a 2D problem. Only the x and y dimension are considered. All values in z direction are set to zero.

uekerman:
> > > I would say let's just ignore the mesh connectivity in this PR and revisit it later.
> > 
> > 
> > Ok, should we add a warning somewhere?
> 
> Definitely yes.

preCICE now throws warnings if you use NP without mesh connectivity. So maybe no warning needed here. 

https://github.com/precice/precice/blob/51542b11e39f67cb55e60a7a7d0041af96bf0308/src/mapping/NearestProjectionMapping.cpp#L77

MakisH:
> preCICE now throws warnings if you use NP without mesh connectivity. So maybe no warning needed here.
> 
> https://github.com/precice/precice/blob/51542b11e39f67cb55e60a7a7d0041af96bf0308/src/mapping/NearestProjectionMapping.cpp#L77

That's good, but it actually a step further. In the adapter, we currently have a switch for providing or not providing mesh connectivity (as this needs to triangulate the mesh first and we don't know if it can fail in some complicated cases). When a usere enables mesh connectivity and also asks for a 2D behavior, then we should have an early error in the configuration setp.

MakisH:
This should have been targetted develop, thus the revert. But at the end, I decided to eitherway update master.

PR #97, opened 2019-08-20
WIP: Add volume coupling of temperature

JaegerSt:
openFOAM volume coupling is implemented for the temperature field located in the AITS directory

### DIFF
--- a/AITS/TemperatureAIT.C
+++ b/AITS/TemperatureAIT.C
@@ -1,0 +1,82 @@
+#include "TemperatureAIT.H"
+
+using namespace Foam;
+
+preciceAdapter::AITS::Temperature::Temperature
+(
+    const Foam::fvMesh& mesh,
+    const std::string nameT
+    )
+:
+T_(
+    const_cast<volScalarField*>
+    (
+        &mesh.lookupObject<volScalarField>(nameT)
+    )
+    ),
+    mesh_(mesh)
+{
+    dataType_ = scalar;
+}
+
+void preciceAdapter::AITS::Temperature::write(double * buffer, bool meshConnectivity)      //Bernhard Version 2
+{
+    int bufferIndex = 0;
+
+    const volScalarField & Temprtr_ = *T_;
+
+    // For every cell of the mesh
+    forAll(Temprtr_, i)
+    {
+        // Copy the temperature into the buffer
+        buffer[bufferIndex++]
+        =
+        Temprtr_[i];
+    }
+
+    for (uint j = 0; j < patchIDs_.size(); j++ )
+    {
+        int patchID = patchIDs_.at(j);
+
+        // For every cell of the patch
+        forAll(T_->boundaryFieldRef()[patchID], i)
+        {
+            // Copy the temperature into the buffer
+            buffer[bufferIndex++]
+            =
+            T_->boundaryFieldRef()[patchID][i];
+        }
+    }
+}
+
+void preciceAdapter::AITS::Temperature::read(double * buffer)     // Bernhard Version 2
+{
+    int bufferIndex = 0;
+
+    volScalarField &Temprtr_ = *T_;
+
+    // For every cell of the mesh
+    forAll(Temprtr_, i)
+    {
+        // Set the temperature as the buffer value
+        Temprtr_[i]
+        =
+        buffer[bufferIndex++];
+    }
+    *T_ = Temprtr_;
+
+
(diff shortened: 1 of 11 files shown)


MakisH:
Thanks for the effort, @JaegerSt! This is something that people have already asked a few times for. I will have a look into it.

In the meantime, if you have any additional information/description or (more crucial) a simple test case, they would be very welcome.

MakisH:
During the preCICE Workshop, it was mentioned that this works well, even for different scenarios. More details will probably follow soonish.

AITS/TemperatureAIT.C
```diff
@@ -0,0 +1,82 @@
+#include "TemperatureAIT.H"
```
MakisH: Together with the comment about the name `AIT`, this file does not need to have a different name.

It could either be `Temperature` in a namespace that implies volume coupling, or something like `TemperatureVolume`.

AITS/TemperatureAIT.C
```diff
@@ -0,0 +1,82 @@
...
+    (
+        &mesh.lookupObject<volScalarField>(nameT)
+    )
+    ),
+    mesh_(mesh)
+{
+    dataType_ = scalar;
+}
+
+void preciceAdapter::AITS::Temperature::write(double * buffer, bool meshConnectivity)      //Bernhard Version 2
```
MakisH: I am not sure why this comment is here, but it does not look necessary.
```suggestion
void preciceAdapter::AITS::Temperature::write(double * buffer, bool meshConnectivity)
```

AITS/TemperatureAIT.C
```diff
@@ -0,0 +1,82 @@
...
+{
+    int bufferIndex = 0;
+
+    const volScalarField & Temprtr_ = *T_;
```
MakisH: Was this maybe meant to be `Tempptr`? A comment would help.

AITS/TemperatureAIT.H
```diff
@@ -0,0 +1,50 @@
+#ifndef AITS_TEMPERATURE_H
+#define AITS_TEMPERATURE_H
+
+#include "CouplingDataUser.H"
+#include "fvCFD.H"
+
+namespace preciceAdapter
+{
+namespace AITS
```
MakisH: I am not sure I understand why we need a separate module, instead of implementing the same functionality as a different writer/reader of the CHT module. I think we could save some code duplication.

Interface.C
```diff
@@ -1,6 +1,9 @@
 #include "Interface.H"
 #include "Utilities.H"
 #include "faceTriangulation.H"
+#include <vector>
```
MakisH: I don't see where we are using these libraries. We could probably remove these lines.

Interface.C
```diff
@@ -203,7 +206,69 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+        numDataLocations_ = mesh.C().size();
+
+        // Count the data locations for all the patches
+        // and add the those to the previously determined number of mesh points in the volume
+        for (uint j = 0; j < patchIDs_.size(); j++)
+        {
+            numDataLocations_ +=
+                    mesh.boundaryMesh()[patchIDs_.at(j)].faceCentres().size();
+        }
+        DEBUG(adapterInfo("Number of face centres: " + std::to_string(numDataLocations_)));
```
MakisH: Although OpenFOAM uses British English and we check for both, we currently use US English in the log:
```suggestion
        DEBUG(adapterInfo("Number of face centers: " + std::to_string(numDataLocations_)));
```

MeshHandle.cpp
```diff
@@ -0,0 +1,390 @@
+#include "precice/MeshHandle.hpp"
```
MakisH: Why is this file here?

MeshHandle.hpp
```diff
@@ -0,0 +1,330 @@
+#pragma once
```
MakisH: Why is this file here?

JaegerSt:
Sorry for the delayed response. This work was up to one of our students who already finished. We will hopefully manage it soon to solve the conflicts.

MakisH:
Some news on picking up the development of volume coupling for the OpenFOAM adapter: https://github.com/precice/openfoam-adapter/pull/183#issuecomment-1125754941

MakisH:
Closing in favor of #255, which implements all review points, offers the same features, and more. Will be there in the next release of the adapter (most probably named v1.3.0).

PR #107, opened 2019-12-19
Update OpenFOAM-OpenFOAM tutorials according to major preCICE config changes

MakisH:
This PR applies the same changes as https://github.com/precice/tutorials/pull/60, namely:
- Remove the `precice-config_parallel.xml` files. We don't need `mpi-single` to run parallel cases anymore. Furthermore, `distribution-type` is removed from preCICE (breaking change).
- Rename `precice-config_serial.xml` to `precice-config.xml`, replacing the respective link.
- Remove the link creation mechanism in `runFluid`, `runSolid`, `Allrun`, `Allrun_parallel`.

Depends on https://github.com/precice/precice/pull/572.


tutorials/CHT/flow-over-plate/buoyantPimpleFoam-laplacianFoam/precice-config.xml
```diff
@@ -48,13 +48,13 @@
         <exchange data="Heat-Flux" mesh="Solid-Mesh" from="Solid" to="Fluid"/>
         <relative-convergence-measure limit="1.0e-6" data="Temperature" mesh="Fluid-Mesh"/>
         <extrapolation-order value="2"/>
-        <post-processing:IQN-ILS>
+        <acceleration:IQN-ILS>
```
MakisH: This corrensponds to a diferent change in preCICE, but since in the tutorials repository they are treated as one, it's ok to treat it as one here as well.

MakisH: However, we need to update this for every tutorial.

BenjaminRueth: https://github.com/precice/tutorials/pull/60 already contains this change. E.g. [here](https://github.com/precice/tutorials/pull/60/files#diff-850e1be720a942764ab616149b633307R50-R56). Sorry, if I am causing confusion. I thought that https://github.com/precice/openfoam-adapter/pull/107 is just the OpenFOAM version of https://github.com/precice/tutorials/pull/60. (And it is more convenient for our current testing problems, if the breaking changes are not scattered)

MakisH: Yes, but we have more tutorials here (i.e. the nearest-projection variant). Just let me take care of this.

BenjaminRueth: > And it is more convenient for our current testing problems, if the breaking changes are not scattered

I think I can hack around this problem in my testing branch (see https://github.com/precice/systemtests/commit/2460f9afeae217eca4e21ae3b8ccba819d706229). Therefore, you can ignore this argument. If you want to keep things separated, feel free to `git revert` 1e826a8

MakisH:
This is ready to merge, but the system tests fail with unclear reason.

Eder-K:
The error in this tests seems to be the same one that is currently being worked on in https://github.com/precice/systemtests/pull/150. When triggering a test here, we should therefore have it run on the branch of the PR.

https://github.com/precice/openfoam-adapter/blob/eb94d5a67be2a72f9b0591e16d021c566823026f/.travis.yml#L11

However, the triggering script used here is taken from the master branch of systemtests. That script version in turn is quite old and only runs tests on the systemtests/master branch, which doesn't contain the aforementioned fixes from https://github.com/precice/systemtests/pull/150. The functionality for triggering on different branches is currently in https://github.com/precice/systemtests/pull/143.

So, to run the test, we could alter the .travis.yml file here to first pull the new script version from PR 143 and then run on PR 150:
```
      script: 
        - curl -LO --retry 3 https://raw.githubusercontent.com/precice/systemtests/i_141/trigger_systemtests.py
        - travis_wait 60 python trigger_systemtests.py --adapter openfoam --wait --branch BenjaminRueth-patch-of-of
```

Issue #4, opened 2017-11-27
Evaluating the boundaries after reading a checkpoint fails for some fields


MakisH:
In OpenFOAM, each cell has a value at the center, but it can also have values at its faces, for example for boundary cells.

It has been found that re-evaluating the faces of the interface cells after reading a checkpoint can help the stability of the solution. This is now done for (almost) all the checkpointed fields of type `volScalarField` and `volVectorField`, by default.

However, it looks like the way this is implemented causes problems for some fields. In particular for turbulent solvers that register the `epsilon` field, the field is checkpointed, but when trying to evaluate the boundaries, it cannot find it. More strangely, it requests another field:

```
--> FOAM FATAL ERROR: 

    request for volScalarField::Internal kEpsilon:G from objectRegistry region0 failed
    available objects of type volScalarField::Internal are

21
(
thermo:mu
thermo:psi
nut
K
K_0
h
p_rgh_0
rho
k
h_0
p_rgh
dpdt
gh
alphat
rho_0
p
T
p_0
thermo:rho
epsilon
thermo:alpha
)


    From function const Type& Foam::objectRegistry::lookupObject(const Foam::word&) const [with Type = Foam::DimensionedField<double, Foam::volMesh>]
    in file /home/ubuntu/OpenFOAM/OpenFOAM-5.x/src/OpenFOAM/lnInclude/objectRegistryTemplates.C at line 193.
```

The methods involved are the `preciceAdapter::Adapter::readCheckpoint()` and the `preciceAdapter::Adapter::setupCheckpointing()`, which are implemented in the [Adapter.C](https://github.com/precice/openfoam-adapter/blob/master/Adapter.C).

The object `epsilon` is a match when executing `mesh_.lookupClass<volScalarField>().toc()`.

Currently, no object named `epsilon` is re-evaluated, in order to prevent unexpected errors. A warning is shown when adding fields to the list of checkpointed fields.

**In order to reproduce this bug**, do the following:
1. In the [Adapter.C](https://github.com/precice/openfoam-adapter/blob/master/Adapter.C), in `preciceAdapter::Adapter::readCheckpoint()`, comment the `if ("epsilon" != volScalarFields_.at(i)->name())` and the two brackets.
[...]

Issue #92, opened 2019-07-08
buoyantSimpleFoam crashing with implicit coupling


MakisH:
@KyleDavisSA noticed that, if buoyant**S**impleFoam (steady-state solver) is used with implicit coupling, it will crash while reading a checkpoint.

Example output, when ran with the [shell-and-tubes heat exchanger tutorial](https://github.com/precice/precice/wiki/Tutorial-for-CHT-with-OpenFOAM-and-CalculiX):

```
---[preciceAdapter] [DEBUG] Advancing preCICE...
(0) 19:48:54 [cplscheme::BaseCouplingScheme]:673 in measureConvergence: [0mrelative convergence measure: two-norm diff = 1812.58, relative limit = 55.0467, conv = false
(0) 19:48:54 [impl::SolverInterfaceImpl]:382 in advance: [0mit 2 of 50 | dt# 1 | t 0 of 500 | dt 1 | max dt 1 | ongoing yes | dt complete no | read-iteration-checkpoint | 
---[preciceAdapter] [DEBUG] Reading a checkpoint...
---[preciceAdapter] [DEBUG] Reloaded time value t = 0.000000


--> FOAM FATAL ERROR: 
updateCoeffs(const scalarField& snGradp) MUST be called before updateCoeffs() or evaluate() to set the boundary gradient.

    From function virtual void Foam::fixedFluxPressureFvPatchScalarField::updateCoeffs()
    in file fields/fvPatchFields/derived/fixedFluxPressure/fixedFluxPressureFvPatchScalarField.C at line 157.

FOAM exiting
```

In this case, the solver Inner-Fluid is configured with:

```
write-data : 
  Sink-Temperature-Inner-Fluid
  Heat-Transfer-Coefficient-Inner-Fluid
read-data : 
  Sink-Temperature-Solid
  Heat-Transfer-Coefficient-Solid
```

For this I used OpenFOAM 5, the adapter from the latest develop (https://github.com/precice/openfoam-adapter/commit/88dfbbae21a0dbdc9f21357a2a1d5d1954646a61) and parallel-implicit coupling with constant under-relaxation. 

Configuration file: [precice-config_serial.xml.txt](https://github.com/precice/openfoam-adapter/files/3369640/precice-config_serial.xml.txt)

DavidSCN:
This is issue has already been opened in #15

MakisH:
I hadn't noticed that the same occurs there as well, thanks! Now we have more clues to find and fix the problem. ;)

Issue #15, opened 2018-02-14
fixedFluxPressure boundary fails


jaydeshpande:
[https://1drv.ms/u/s!AvEMDnpUbHI2ggCtaUp6pvMLnj5O
](url)
In the above case - change p_rgh wall boundary from zeroGradient to fixedFluxPressure. 

ONLY when run through PreCICE - it throws following error:
`updateCoeffs(const scalarField& snGradp) MUST be called before updateCoeffs() or evaluate() to set the boundary gradient.
`

Simply running buoyantPimpleFoam works.

jaydeshpande:
Patch - 

Noticed that it evaluates the boundary again at preciceAdapter::Adapter::readCheckpoint()

modified the evaluateBoundaries_ similar to epsilon : 
```
                if ("p_rgh" != volScalarFields_.at(i)->name())
                {
                    volScalarFields_.at(i)->correctBoundaryConditions();
                }
```

MakisH:
Thank you very much for the bug report. Looks similar to #4 and it is useful to find more cases where this happens. I leave it open for now, until we find the root cause of the problem.

PR #110, opened 2020-01-27
Remove boundary evaluation for checkpointing

DavidSCN:
First workaround for #4

Using these few lines of code, we could circumvent the error, which is triggered by `correctBoundaryConditions()` on the epsilonWallFunction bc type. This is a first step and not the final solution, since there are also other issues,(#92 #15) which have a similar problem and we want to find a more general solution. 
We still don't know in detail, if we need the function call (=what it does).


MakisH:
Is this ready to review?

DavidSCN:
Yes. Although it basically removes the boundary evaluation and corresponding warnings.

The only thing we still need to adapt (remove) is the option in the OF dict.

MakisH:
> The only thing we still need to adapt (remove) is the option in the OF dict.

Could you please also prepare this in the same PR?

Also, could you please give a more descriptive title and remove the "WIP"?

DavidSCN:
Done so far. There are a few lines with a removed spacing, which appear in the commit, sorry for that.

This PR would close #4, #15 and #92.

PR #121, opened 2020-02-20
FSI: OF solid mechanics and OF fluid mechanics

JSeuffert:
This is a first attempt to couple an OF fluid solver to an OF solid mechanics solver. 
You can only use it with some limitations:
- You can only couple total displacements (calculated by solid) to total displacements (used to move fluid) **or** incremental displacements (calculated by solid) to incremental displacements (used to move fluid). No combination of these is possible.
- Displacements are only coupled at the vertices (pointDisplacement)
- You need to write a boundary condition for the solid mechanics solver: e.g. ForceDisplacement, that is like the traction ones but reads the "Force" field that is calculated in the adapter.


MakisH:
Thank you once again for the contribution, @JSeuffert! Do you maybe have some tutorial case as well, so that I can try it?

MakisH:
@JSeuffert why did you decide to close this? A brief documentation of the decision could help in the future. :)

JSeuffert:
Oh I was not aware that there still was this open pull request. I just wanted to clean my repository... 
I will open it again, then also based on the current adapter version.

PR #124, opened 2020-03-15
Determine solver type by pressure dimension

### LINKED ISSUE #31: Check for (in)compressible solver type with Pressure dimensions
Currently, we check the solver type based on the registered IOobjects. This is a very indirect way to check for compressible vs incompressible vs basic solvers, but it is maybe a not-so-bad way to check for the available models (classes we can access).

https://github.com/precice/openfoam-adapter/blob/c0f28814c4d8da073a73cd076dc4e1dfd5da7e74/CHT/CHT.C#L109-L196

However, this is a complicated way and maybe it would be easier to just check for the  dimensions of Pressure to distinguish between compressible and incompresible solver. After this check, we can check if it is a basic solver (e.g. laplacianFoam) or an of an unknown type.

For compresible solvers, the dimensions of Pressure are `[1 -1 -2 0 0 0 0]`, while for incompressible `[0 2 -2 0 0 0 0]`.

DavidSCN:
This PR determines the solver type via the pressure dimension. 

Here a few comments, following from our discussion in #31:

The type 'basic' is assumed instead of naming it 'other'. So we raise a warning and tell the user, that he might get issues with this settings. I preferred now 'basic', since this is an actual OpenFOAM type and the user might (or not) be fine with the settings. If we tell the user, that we calculate with type 'other', no one knows what this means without looking in the code.

I added the checker in the module classes and decided not to pass it in the adapter class, because we first read the configuration in the specific module and start afterwards to determine the type 'automatically' and we cannot access (without passing the whole class) the adapter class in the modules.
EDIT: We could determine it in the adapter class and overwrite it in the modules. 

However, there are just some conceptual questions. The general ingredients should be fine.

Closes #31.

### DIFF
--- a/CHT/CHT.C
+++ b/CHT/CHT.C
@@ -101,10 +101,13 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
     }
 
     if (solverType == "unknown")
-      adapterInfo("Failed to determine the solver type. This might lead to "
-                  "issues during your simulation. Using 'basic' as solver type. "
-                  "Please specify the solver type in the preciceDict file.",
-                  "warning");
+      adapterInfo("Failed to determine the solver type. "
+                  "Please specify your solver type in the preciceDict. "
+                  "Known solver types are: "
+                  "Incompressible"
+                  "Compressible"
+                  "Basic",
+                  "error");
 
     return solverType;
 }
--- a/CHT/CHT.C
+++ b/CHT/CHT.C
@@ -85,22 +85,22 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
     // NOTE: When coupling a different variable, you may want to
     // add more cases here. Or you may provide the solverType in the config.
 
-    std::string solverType = "basic";
+    std::string solverType = "unknown";
 
-    dimensionSet compressible_pressure_dimension(1, -1, -2, 0, 0, 0, 0);
-    dimensionSet incompressible_pressure_dimension(0, 2, -2, 0, 0, 0, 0);
+    dimensionSet pressureDimensionsCompressible(1, -1, -2, 0, 0, 0, 0);
+    dimensionSet pressureDimensionsIncompressible(0, 2, -2, 0, 0, 0, 0);
 
     if (mesh_.foundObject<volScalarField>("p"))
     {
       volScalarField p_ = mesh_.lookupObject<volScalarField>("p");
 
-      if (p_.dimensions() == compressible_pressure_dimension)
+      if (p_.dimensions() == pressureDimensionsCompressible)
         solverType = "compressible";
-      else if (p_.dimensions() == incompressible_pressure_dimension)
+      else if (p_.dimensions() == pressureDimensionsIncompressible)
         solverType = "incompressible";
     }
 
-    if (solverType == "basic")
+    if (solverType == "unknown")
       adapterInfo("Failed to determine the solver type. This might lead to "
                   "issues during your simulation. Using 'basic' as solver type. "
                   "Please specify the solver type in the preciceDict file.",
--- a/CHT/CHT.C
+++ b/CHT/CHT.C
@@ -102,11 +102,13 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
 
     if (solverType == "unknown")
       adapterInfo("Failed to determine the solver type. "
-                  "Please specify your solver type in the preciceDict. "
-                  "Known solver types for CHT are: "
-                  "Incompressible"
-                  "Compressible",
+                  "Please specify your solver type in the CHT section of the "
+                  "preciceDict. Known solver types for CHT are: "
+                  "basic, incompressible and "
+                  "compressible",
                   "error");
+
+    DEBUG(adapterInfo("Automatically determined solver type : " + solverType));
 
     return solverType;
 }
--- a/CHT/CHT.C
+++ b/CHT/CHT.C
@@ -103,7 +103,7 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
     if (solverType == "unknown")
       adapterInfo("Failed to determine the solver type. "
                   "Please specify your solver type in the preciceDict. "
-                  "Known solver types are: "
+                  "Known solver types for CHT are: "
                   "Incompressible"
                   "Compressible"
                   "Basic",
--- a/CHT/CHT.C
+++ b/CHT/CHT.C
@@ -105,8 +105,7 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
                   "Please specify your solver type in the preciceDict. "
                   "Known solver types for CHT are: "
                   "Incompressible"
-                  "Compressible"
-                  "Basic",
+                  "Compressible",
                   "error");
 
     return solverType;
--- a/CHT/CHT.C
+++ b/CHT/CHT.C
@@ -48,7 +48,7 @@ bool preciceAdapter::CHT::ConjugateHeatTransfer::configure(const IOdictionary& a
 bool preciceAdapter::CHT::ConjugateHeatTransfer::readConfig(const IOdictionary& adapterConfig)
 {
     const dictionary CHTdict = adapterConfig.subOrEmptyDict("CHT");
-  
+
     // Read the solver type (if not specified, it is determined automatically)
     solverType_ = CHTdict.lookupOrDefault<word>("solverType", "");
     DEBUG(adapterInfo("    user-defined solver type : " + solverType_));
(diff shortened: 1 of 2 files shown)


DavidSCN:
Sorry, my base branch is now OF 6. The committed changes are the same.

Adapter.C
```diff
@@ -611,7 +611,7 @@ void preciceAdapter::Adapter::adjustSolverTimeStep()
     // Update the solver's timestep (but don't trigger the adjustDeltaT(),
     // which also triggers the functionObject's adjustTimeStep())
     // TODO: Keep this in mind if any relevant problem appears.
-    const_cast<Time&>(runTime_).setDeltaT(timestepSolver_, false);
+    const_cast<Time&>(runTime_).setDeltaTNoAdjust(timestepSolver_);
```
MakisH: TODO: Update for target branch

Adapter.H
```diff
@@ -378,7 +378,7 @@ public:
         //- Called by the functionObject's execute()
         void execute();
 
-        //- Called by the functionObject's adjustTimeStep()
+        //- Called by the functionObject's setTimeStep()
```
MakisH: TODO: Update for target branch

CHT/CHT.C
```diff
@@ -85,86 +85,26 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
...
 
-    std::string solverType;
-
-    // Determine the solver type: Compressible, Incompressible or Basic.
-    // Look for the files transportProperties, turbulenceProperties,
-    // and thermophysicalProperties
-    bool transportPropertiesExists = false;
-    bool turbulencePropertiesExists = false;
-    bool thermophysicalPropertiesExists = false;
+    std::string solverType = "basic";
```
MakisH: > The type 'basic' is assumed instead of naming it 'other'. So we raise a warning and tell the user, that he might get issues with this settings. I preferred now 'basic', since this is an actual OpenFOAM type and the user might (or not) be fine with the settings. If we tell the user, that we calculate with type 'other', no one knows what this means without looking in the code.

What if we assume `unknown` here? The difference would be that `other` expects to read something from the config, while `unknown` will only appear if everything goes wrong.

> I preferred now 'basic', since this is an actual OpenFOAM type

Keep in mind that some users define their own solvers, which may be quite exotic (and complex, thus not `basic`). In OpenFOAM, `basic` is a small collection of solvers that do not have much in common that we can take advantage of. We could still support both `basic` and `other`, but then we cannot really distinguish between these two.

CHT/CHT.C
```diff
@@ -85,86 +85,26 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
...
-    if (mesh_.foundObject<IOdictionary>("transportProperties"))
-    {
-        transportPropertiesExists = true;
-        DEBUG(adapterInfo("Found the transportProperties dictionary."));
-    }
-    else
-    {
-        DEBUG(adapterInfo("Did not find the transportProperties dictionary."));
-    }
+    dimensionSet compressible_pressure_dimension(1, -1, -2, 0, 0, 0, 0);
```
MakisH: Why `snake_case` here? I would prefer something like `pressureDimensionsCompressible` and `pressureDimensionsIncompressible`.

CHT/CHT.C
```diff
@@ -85,86 +85,26 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
...
-        {
-            adapterInfo("Could not determine the solver type, or this is not a "
-            "compatible solver: neither transport, nor turbulence properties "
-            "are provided.",
-            "error");
-        }
-    }
+    if (solverType == "basic")
+      adapterInfo("Failed to determine the solver type. This might lead to "
+                  "issued during your simulation. Using 'basic' as solver type. "
```
MakisH:
```suggestion
                  "issues during your simulation. Using 'basic' as solver type. "
```

preciceAdapterFunctionObject.C
```diff
@@ -102,7 +102,7 @@ bool Foam::functionObjects::preciceAdapterFunctionObject::write()
     return true;
 }
 
-bool Foam::functionObjects::preciceAdapterFunctionObject::adjustTimeStep()
+bool Foam::functionObjects::preciceAdapterFunctionObject::setTimeStep()
```
MakisH: TODO: Update for target branch

preciceAdapterFunctionObject.H
```diff
@@ -128,7 +128,7 @@ public:
         virtual bool write();
 
         //- Called at the end of Time::adjustDeltaT() if adjustTime is true
-        virtual bool adjustTimeStep();
+        virtual bool setTimeStep();
```
MakisH: TODO: Update for target branch

CHT/CHT.C
DavidSCN: Yes, of course!

Adapter.C
DavidSCN: Are these TODOs for me? I don't understand

CHT/CHT.C
DavidSCN: > What if we assume unknown here? The difference would be that other expects to read something from the config, while unknown will only appear if everything goes wrong.

I'm fine with it. But what should the module do afterwards? Crash? Or select the 'basic' implementation?

DavidSCN:
> Afterwards, we could also try moving it into `Adapter`, but now the respective code is much much shorter that I am not sure if it would solve more problems than it would create. A user-defined module may implement additional types (e.g. linear model vs non-linear model for XYZ) and may want to combine their module with an existing one.

Yes! I think propagating it in the end to the proper place will not save any line of code.

CHT/CHT.C
DavidSCN: The workflow is now as follows:
Read from the dict: if (type specified) -> fine -> select appropriate type and calculate
                               else (determine) -> either we get inco or compr or unknown type-> how to proceed here

Adapter.C
MakisH: These (TODO) changes come from the OpenFOAM6 branch, which we don't want to bring into develop/master. If you rebase on develop, they will get fixed automatically. ;-) I just added these comments so that we don't forget it.

CHT/CHT.C
MakisH: If one of the two gets determined (i.e. there is a pressure field), let it be whatever gets determined. In any case, we should inform the user that we are determining the solver type automatically.

If none of these can be determined (i.e. there is no pressure field), we can either:
1. say "unknown solver type, please define it in preciceDict", or (not as important)
2. check if there is no pressure but there is temperature. In that case, we have our current "basic" type (and hopefully nothing else). If neither pressure not temperature is found, throw an error similar to (1).

DavidSCN: > we should inform the user that we are determining the solver type automatically.

We are informing the user [in a debug message](https://github.com/precice/openfoam-adapter/blob/11db960fdfa0534eaeeb847b402ececa8a8bcb68/CHT/CHT.C#L36)

Ok, then I vote for option one and I will throw an error

DavidSCN:
TODO: Update FSI.C

FSI/FSI.C
```diff
@@ -72,88 +72,29 @@ std::string preciceAdapter::FSI::FluidStructureInteraction::determineSolverType(
...
-        }
-    }
+    if (solverType == "unknown")
+      adapterInfo("Failed to determine the solver type. "
+                  "Please specify your solver type in the preciceDict. "
+                  "Known solver types are: "
+                  "Incompressible"
+                  "Compressible"
+                  "Basic",
+                  "error");
```
MakisH: How are we treating the `error` solver type? Is this something that we want to show to the user?

DavidSCN: Hm this was my understanding of your proposal in the discussion above:
> If none of these can be determined (i.e. there is no pressure field), we can either:
say "unknown solver type, please define it in preciceDict"

MakisH: I just realized that `error` is just the usual argument to `adapterInfo()`. I got confused by the different lines.

MakisH: Looks good then!

DavidSCN: Alright, then I will test everything and merge, if it is fine.

FSI/FSI.C
```diff
@@ -72,88 +72,29 @@ std::string preciceAdapter::FSI::FluidStructureInteraction::determineSolverType(
...
-            "error");
-        }
-    }
+    if (solverType == "unknown")
+      adapterInfo("Failed to determine the solver type. "
+                  "Please specify your solver type in the preciceDict. "
+                  "Known solver types are: "
+                  "Incompressible"
+                  "Compressible"
+                  "Basic",
```
MakisH: In the FSI module we don't do anything with `Basic` solvers, right? We could remove it then.

FSI/FSI.C
```diff
@@ -72,88 +72,29 @@ std::string preciceAdapter::FSI::FluidStructureInteraction::determineSolverType(
...
-            adapterInfo("Could not determine the solver type, or this is not a "
-            "compatible solver: neither transport, nor turbulence properties "
-            "are provided.",
-            "error");
-        }
-    }
+    if (solverType == "unknown")
+      adapterInfo("Failed to determine the solver type. "
+                  "Please specify your solver type in the preciceDict. "
+                  "Known solver types are: "
```
MakisH:
```suggestion
                  "Known solver types for FSI are: "
```

CHT/CHT.C
```diff
@@ -85,86 +85,29 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
...
-            adapterInfo("Could not determine the solver type, or this is not a "
-            "compatible solver: neither transport, nor turbulence properties "
-            "are provided.",
-            "error");
-        }
-    }
+    if (solverType == "unknown")
+      adapterInfo("Failed to determine the solver type. "
+                  "Please specify your solver type in the preciceDict. "
+                  "Known solver types are: "
```
MakisH:
```suggestion
                  "Known solver types for CHT are: "
```

FSI/FSI.C
DavidSCN: Done, but the user is still able to define it, have a look at the CHT configure function. Anyway, I removed it from the message.

DavidSCN: Ah sorry, now I get it.

DavidSCN:
Tested and works as expected!

PR #125, opened 2020-03-16
Support stress data

### LINKED ISSUE #41: FSI: Add an option for stress mapping
At the moment, we map forces. However, it would be very nice to also be able to map stresses, which would be easy to implement, having already implemented `Force.C`.

A limitation here is that we cannot use stresses in CalculiX at the moment. So, a prerequisite for this would be to either:

A. Implement stresses reading also in the [calculix-adapter](https://github.com/precice/calculix-adapter),
B. Add connectivity information in the openfoam-adapter, so that we use a [preCICE action](https://github.com/precice/precice/wiki/Actions-Configuration) to convert the stresses to forces (see also #2 ).
C. Use another structure solver that already supports reading stresses (is there one already?).

This is relevant to #7 .

DavidSCN:
This PR adds an option for users to map stresses instead of forces.

 I tried to balance the similarity between the existing forces module by inheriting it and reuse some existing functionalities there Therefore, it still has its own class (as usual for each data module). The stress calculation is very similar: The multiplication by the face are is skipped and a normal vector is used instead.

The code runs, but it is not testes yet.

Closes #41.

### DIFF
--- a/FSI/Force.C
+++ b/FSI/Force.C
@@ -178,7 +178,7 @@ Foam::tmp<Foam::volScalarField> preciceAdapter::FSI::Force::mu() const
     }
 }
 
-void preciceAdapter::FSI::Force::write(double * buffer, bool /*meshConnectivity*/, const unsigned int dim)
+void preciceAdapter::FSI::Force::write(double * buffer, bool meshConnectivity, const unsigned int dim)
 {
     // Compute forces. See the Forces function object.
 
@@ -255,7 +255,7 @@ void preciceAdapter::FSI::Force::write(double * buffer, bool /*meshConnectivity*
     }
 }
 
-void preciceAdapter::FSI::Force::read(double * /*buffer*/, const unsigned int /*dim*/)
+void preciceAdapter::FSI::Force::read(double * buffer, const unsigned int dim)
 {
     /* TODO: Implement
     * We need two nested for-loops for each patch,
--- a/FSI/Force.C
+++ b/FSI/Force.C
@@ -4,6 +4,17 @@ using namespace Foam;
 
 preciceAdapter::FSI::Force::Force
 (
+        const Foam::fvMesh& mesh,
+        const std::string solverType
+)
+:
+mesh_(mesh),solverType_(solverType)
+{}
+
+
+
+preciceAdapter::FSI::Force::Force
+(
     const Foam::fvMesh& mesh,
     const fileName& timeName,
     const std::string solverType
@@ -167,7 +178,7 @@ Foam::tmp<Foam::volScalarField> preciceAdapter::FSI::Force::mu() const
     }
 }
 
-void preciceAdapter::FSI::Force::write(double * buffer, bool meshConnectivity, const unsigned int dim)
+void preciceAdapter::FSI::Force::write(double * buffer, bool /*meshConnectivity*/, const unsigned int dim)
 {
     // Compute forces. See the Forces function object.
 
@@ -244,7 +255,7 @@ void preciceAdapter::FSI::Force::write(double * buffer, bool meshConnectivity, c
     }
 }
 
-void preciceAdapter::FSI::Force::read(double * buffer, const unsigned int dim)
+void preciceAdapter::FSI::Force::read(double * /*buffer*/, const unsigned int /*dim*/)
 {
     /* TODO: Implement
     * We need two nested for-loops for each patch,
--- a/FSI/Force.C
+++ b/FSI/Force.C
@@ -12,7 +12,6 @@ mesh_(mesh),solverType_(solverType)
 {}
 
 
-
 preciceAdapter::FSI::Force::Force
 (
     const Foam::fvMesh& mesh,
@@ -24,8 +23,7 @@ preciceAdapter::FSI::Force::Force
     */
 )
 :
-mesh_(mesh),
-solverType_(solverType)
+Force(mesh, solverType)
 {
     //What about type "basic"?
     if (solverType_.compare("incompressible") != 0 && solverType_.compare("compressible") != 0) 
--- a/FSI/Force.H
+++ b/FSI/Force.H
@@ -5,7 +5,6 @@
 
 #include "fvCFD.H"
 
-// TEMPORARY
 #include "pointFields.H"
 #include "vectorField.H"
 #include "immiscibleIncompressibleTwoPhaseMixture.H"
--- a/FSI/Force.H
+++ b/FSI/Force.H
@@ -64,10 +64,10 @@ public:
     );
 
     //- Write the displacement values into the buffer
-    virtual void write(double * buffer, bool meshConnectivity, const unsigned int dim);
+    void write(double * buffer, bool meshConnectivity, const unsigned int dim);
 
     //- Read the displacement values from the buffer
-    virtual void read(double * buffer, const unsigned int dim);
+    void read(double * buffer, const unsigned int dim);
 
     //- Destructor
     ~Force();
--- a/FSI/Force.H
+++ b/FSI/Force.H
@@ -62,10 +62,10 @@ public:
         const std::string solverType
     );
 
-    //- Write the displacement values into the buffer
+    //- Write the forces values into the buffer
     void write(double * buffer, bool meshConnectivity, const unsigned int dim);
 
-    //- Read the displacement values from the buffer
+    //- Read the forces values from the buffer
     void read(double * buffer, const unsigned int dim);
 
     //- Destructor
--- a/FSI/Stress.C
+++ b/FSI/Stress.C
@@ -1,0 +1,142 @@
+#include "Stress.H"
+
+using namespace Foam;
+
+preciceAdapter::FSI::Stress::Stress
+(
+    const Foam::fvMesh& mesh,
+    const fileName& timeName,
+    const std::string solverType
+    /* TODO: We should add any required field names here.
+    /  They would need to be vector fields.
+    /  See FSI/Temperature.C for details.
+    */
+)
+:
+Force(mesh,solverType),
+mesh_(mesh),
+solverType_(solverType)
+{
+    //What about type "basic"?
+    if (solverType_.compare("incompressible") != 0 && solverType_.compare("compressible") != 0) 
+    {
+        FatalErrorInFunction
+            << "Forces calculation does only support "
+            << "compressible or incompressible solver type."
+            << exit(FatalError);
+    }
+    
+    dataType_ = vector;
+
+    Stress_ = new volVectorField
+    (
+        IOobject
+        (
+            "Stress",
+            timeName,
+            mesh,
+            IOobject::NO_READ,
+            IOobject::AUTO_WRITE
+        ),
+        mesh,
+        dimensionedVector
+        (
+            "pdim",
+            dimensionSet(1,-1,-2,0,0,0,0),
+            Foam::vector::zero
+        )
+    );
+}
+
+void preciceAdapter::FSI::Stress::write(double * buffer, bool /*meshConnectivity*/, const unsigned int dim)
+{
+    // Compute stress. See the Forces function object.
+
+    // Stress tensor boundary field
+    tmp<volSymmTensorField> tdevRhoReff = this->devRhoReff();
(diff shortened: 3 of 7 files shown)


FSI/Stress.C
```diff
@@ -0,0 +1,142 @@
...
+    /  They would need to be vector fields.
+    /  See FSI/Temperature.C for details.
+    */
+)
+:
+Force(mesh,solverType),
+mesh_(mesh),
+solverType_(solverType)
+{
+    //What about type "basic"?
```
MakisH: What would be an example? I think the error message is enough.

FSI/Force.C
```diff
@@ -167,7 +178,7 @@ Foam::tmp<Foam::volScalarField> preciceAdapter::FSI::Force::mu() const
     }
 }
 
-void preciceAdapter::FSI::Force::write(double * buffer, bool meshConnectivity, const unsigned int dim)
+void preciceAdapter::FSI::Force::write(double * buffer, bool /*meshConnectivity*/, const unsigned int dim)
```
MakisH: Why commenting the names out?

FSI/Force.H
```diff
@@ -31,7 +31,7 @@ private:
     //- Force field
     Foam::volVectorField * Force_;
 
-
+protected:
```
MakisH: :+1:

FSI/Stress.C
DavidSCN: This is a copied comment from the force file, not mine. I think it can be deleted.

FSI/Force.H
```diff
@@ -56,11 +56,18 @@ public:
         */
     );
 
+    //- Constructor
+    Force
```
MakisH: Why do we need an additional constructor without timeName? We do have `timeName` available when we do `Stress : Force`.

FSI/Force.H
```diff
@@ -56,11 +56,18 @@ public:
...
+    (
+        const Foam::fvMesh& mesh,
+        const std::string solverType
+    );
+
     //- Write the displacement values into the buffer
-    void write(double * buffer, bool meshConnectivity, const unsigned int dim);
+    virtual void write(double * buffer, bool meshConnectivity, const unsigned int dim);
```
MakisH: I am not sure if we actually need virtual here. A subclass can just redefine a method, even when it is not virtual. We should still explicitly use a `Stress` object to get the stresses, not a `Force` one.

FSI/Stress.C
```diff
@@ -0,0 +1,142 @@
...
+    if (solverType_.compare("incompressible") != 0 && solverType_.compare("compressible") != 0) 
+    {
+        FatalErrorInFunction
+            << "Forces calculation does only support "
```
MakisH:
```suggestion
            << "Stresses calculation only supports "
```

FSI/Stress.C
```diff
@@ -0,0 +1,142 @@
...
+        )
+    );
+}
+
+void preciceAdapter::FSI::Stress::write(double * buffer, bool /*meshConnectivity*/, const unsigned int dim)
+{
+    // Compute stress. See the Forces function object.
+
+    // Stress tensor boundary field
+    tmp<volSymmTensorField> tdevRhoReff = this->devRhoReff();
```
MakisH: I understand that by deriving from `Force`, you want to reduce duplication, although currently we do duplicate a lot. A good idea would be to further extract some of the duplicate parts into separate protected methods (both for `Force` and `Stress`).

FSI/Stress.H
```diff
@@ -0,0 +1,67 @@
...
+
+#include "CouplingDataUser.H"
+#include "Force.H"
+
+#include "fvCFD.H"
+
+// TEMPORARY
+#include "pointFields.H"
+#include "vectorField.H"
+#include "immiscibleIncompressibleTwoPhaseMixture.H"
```
MakisH: Why is this needed? Looks too specific.

FSI/Force.C
DavidSCN: Depending on your compiler configuration, this might lead to an unused warning. Wmake doesn't complain,  I will revert it.

FSI/Stress.H
DavidSCN: Copied from the forces module.

FSI/Force.H
DavidSCN: Correct, my fault

DavidSCN: Wait, the previous constructor also creates the force field and defines the dataType. The new constructor does nothing and is just for the access we need. I would like to keep it.

FSI/Force.H
DavidSCN: Alright, I will remove it.

FSI/Stress.C
DavidSCN: Taking a closer look at the code, I think the duplication is here quite minimal, since the important difference in the `write` function, namely the calculation using the normal vector, is included in every calculation step. Anyway, we could maybe unify the read (not supported) function. These are the only two methods, which are present here. 
I would propose we keep it like this at the moment. What do you think?

FSI/Force.C
MakisH: WMake still triggers the same compiler and there should be a warning (unless explicitly disabled).

The warning gives us a signal that we maybe need to change something in the design: `meshConnectivity` seems to be "a different kind of animal" than `buffer` and `dim`, which we explicitly need to know for writing. So, we should either overload `write` to have one version with and one without `meshConnectivity`, or select a different one using a [strategy pattern](https://sourcemaking.com/design_patterns/strategy). Or maybe move `meshConnectivity` to some kind of "repository" (if I remember correctly, that's the "repository pattern"), so that we don't need to propagate it all the way.

FSI/Force.H
MakisH: Then we could also make the 3-argument constructor call the 2-argument constructor, to avoid duplication. This is also a common pattern in the OpenFOAM source.

FSI/Stress.C
MakisH: This is also my feeling, let's keep it like this. :+1:

FSI/Force.C
DavidSCN: The current implementation is consistent with the other modules. Hence, these changes would affect the whole adapter and require a separate PR. However, (what I meant and it is just a guess) configuring the code with e.g. cmake would probably result in a compiler warning, whereas wmake doesn't trigger any warning. 
Removing/Commenting the variable is in any case a clean way to avoid warnings.

MakisH: It would definitely be a bigger work package, yes. Let's open an issue for this before resolving this conversation.

Regarding the specific PR: I am fine doing this, as long as there is a comment next to it explaining why we do it.

DavidSCN: Alright, keep in mind that we call this methods unified through the CouplingDataUser and in some other functions, the `meshConnectivity` is needed. So, all arguments are passed in to the function. This was also the reason, why I added them here, although they are not needed. I am not familiar with pattern methods. Do you open up the issue? Then, you could maybe directly describe, what you have in mind.

FSI/Force.H
DavidSCN: Like this?

MakisH: I am not sure what you are refering to, could you add the code here? In `Force.C` I see that the 3-argument constructor does everything and the 2-argument constructor does nothing.

FSI/Stress.H
MakisH: Still, let's check if it is actually needed and remove it if not. Or at least remove the `// TEMPORARY`.

FSI/Force.H
DavidSCN: But the 3 argument constructor calls the 2 argument constructor. The 2 argument constructor initializes the const variables. That is indeed not very much, but exactly what we want for the inheritance. Do I understand correctly?

MakisH: I guess I was looking at an outdated version of the file (clicking on the code here). This looks good: https://github.com/precice/openfoam-adapter/pull/125/commits/e48b212ddc9334a377c541b8b62cdd9661d53c82

FSI/Stress.H
DavidSCN: Ok this is needed in the calculation of `mu` in case of mixture,  I will remove the temporary

DavidSCN:
This is now ready to merge. We should maybe test it before or add some tests for the module.

MakisH:
> We should maybe test it before or add some tests for the module.

Could you modify the existing OpenFOAM-CalculiX flap_perp tutorial with comments on how to switch to stresses? We could then use this as a test.

Instructions in the wiki should be enough, unless you find that the changes are quite extensive. Then it would be nice to have a comparison between Force-Displacement and Stress-Displacement, to see at least that the coupling makes sense.

DavidSCN:
In order to test we would need to modify the solid solver, i.e. it needs to work with stress instead of forces. I cannot modify the calculix adapter, if that is your question.
Assuming we have such a solid solver, a user needs to select a consistent instead of a conservative mapping, this should be sufficient.
Does the calculix adapter support stress data?

MakisH:
> Does the calculix adapter support stress data?

Indeed, it does not support stresses right now, but we could either:
- Compute the stresses using a python action in preCICE
- Prepare an example in deal.II? I assume that would be easier.

Otherwise, to not leave this PR open forever, we could merge it but document that it has not yet been tested and needs some validation. In that case, let's throw a warning when using this writer.

DavidSCN:
Alright, I will make a test in deal.II

DavidSCN:
Tested and works fine. I added a comment in the source code. This is now a scaled variant of the force vector, which is commonly used in FE applications. If you resolve the discussion above, this can be merged.

MakisH:
Could you maybe contribute the test as a tutorial in the tutorials repository? It would be a useful example.

uekerman:
Quick remark: To transform stresses to forces in preCICE, you don't need a Python action. There are already implemented actions doing this. `multiply-by-area` and `divide-by-area`. [The wiki gives more information](https://github.com/precice/precice/wiki/Actions-Configuration#basics-and-pre-implemented-actions). Would be quite a nice test (and tutorial) actually if it also works with OpenFOAM (stresses) - preCICE (conversion) - CalculiX (forces). I guess users don't understand this feature yet as there are no tutorials for it. Depending on which side you do the conversion you could either map consistently or conservatively. It's quite a powerful feature.

DavidSCN:
But one needs to be careful using this method, since preCICE always scales according to the initial mesh configuration, right? The forces usually provided by OF scale with the deformed configuration. I am not sure, which one is needed by CalculiX, but this could result in some differences. 
I think the stress mapping here needs to be consistent instead of conservative, still not completely sure about it, but I will test it.

uekerman:
> But one needs to be careful using this method, since preCICE always scales according to the initial mesh configuration, right? The forces usually provided by OF scale with the deformed configuration. I am not sure, which one is needed by CalculiX, but this could result in some differences.

Yes, preCICE scales with the initial (reference) mesh. This issue with reference vs. physical mesh is really sth we need to investigate better. But beyond this PR, I guess. 

> I think the stress mapping here needs to be consistent instead of conservative, still not completely sure about it, but I will test it.

Yes, stresses need to be mapped consistently. But with the scaling actions, you still have all freedom in the configuration: 
* write stresses, map stresses consistently, communicates stresses, scale stresses to forces, read forces; or
* write stresses, scale stresses to forces, map forces conservatively, communicate forces, read forces.

FSI/Stress.H
```diff
@@ -0,0 +1,64 @@
...
+        const fileName& timeName,
+        const std::string solverType
+        // We create an IOobject and we need the time directory
+        /* TODO: We should add any required field names here.
+        */
+    );
+
+    //- Write the stress values into the buffer
+    void write
+    (double * buffer, bool meshConnectivity, const unsigned int dim);
```
MakisH, line 53:
```suggestion
    void write(double * buffer, bool meshConnectivity, const unsigned int dim);
```

DavidSCN:
> Looks nice and clean, I have not tested it with any example yet.

I have https://github.com/precice/openfoam-adapter/pull/125#issuecomment-601942511 and tutorials will be in https://github.com/precice/tutorials/pull/79.

PR #91, opened 2019-07-08
[Will not merge] Compatibility with OpenFOAM 7

MakisH:
Main change from OpenFOAM 6: `fileName::DIRECTORY` was changed to `fileType::directory`.

We are looking for a [nice way to support multiple versions at the same time](https://github.com/precice/openfoam-adapter/issues/32). Any ideas?

Issue #52, opened 2018-12-20
Releases and versioning


MakisH:
We should make proper releases of the adapter, similarly to preCICE. This way, a user can directly let us know which version she is using, we can announce release changes, thank contributors etc.

Questions:
* Which version number do we start from? I vote for `0.1.0` or similar. If we gave `1.0`, it would give the false impression that the project is as mature as preCICE itself.
* Should we just release different variants of the code for each supported OpenFOAM version? If yes, how should they be named? See also #32.
* If we wanted provide packages for each version instead, what could they be? There is already a Spack package with OpenFOAM v1806. OpenFOAM usually looks at its own paths for libraries, we should be able to write on them, or edit these paths.

MakisH:
I lately converge to the following idea:
- release the adapter every time there is a new OpenFOAM.com version
- name the releases with dates, similarly to the OpenFOAM versions, and assume that all versions up to this work (.com only).
- keep the OpenFOAM.org versions in different branches, but do not refer the user to them. Instead, release additional archives for .org, with a suffix in the version.

Semantic versioning does not make much sense here, since we do not have any specific API to maintain. We do not plan to change the configuration file significantly from now on, the main dependency that makes versioning complicated is OpenFOAM.

We do not expect any versioning issues coming from changes in preCICE.

An important question is what happens if we greatly refactor the adapter and introduce more dependencies. This could come by pull requests such as #150.

uekerman:
Just discussed with @MakisH, I like option 3.

uekerman:
> We do not expect any versioning issues coming from changes in preCICE

What do we do for a breaking preCICE release? A major release of the adapter, I guess?

MakisH:
> > We do not expect any versioning issues coming from changes in preCICE
> 
> What do we do for a breaking preCICE release? A major release of the adapter, I guess?

If nothing changes for the user, then it should be a minor release for the adapter, I guess. One example I can think of is Atom and Electron: releasing Atom `1.55 -> 1.56`, they upgraded Electron `6.1.12 -> 9.3.5` ([details](https://github.com/atom/atom/releases/tag/v1.56.0)).

I am not sure how to define a "stable API" for an adapter, thus (as I wrote above), a standard semantic versioning seems to be non-fitting. Similarly, I can see that OpenFOAM cannot define that, thus OpenCFD chose the chronological format and the OpenFOAM Foundation increases the (now single) first digit with every release.

[...]

PR #128, opened 2020-05-19
Version fusion

### LINKED ISSUE #32: Support multiple OpenFOAM versions with the same source

MakisH:
This pull request isolates version-specific changes into separate files included at compile-time (we could even say "configuration-time"), trying to unify all the different variants of the adapter we are currently offering as branches. See #32 for a discussion of the problem (closes #32).

Motivation:
- Currently, with every version-specific branch we apply short patches.
- OpenFOAM already splits code into parts that it includes in the middle of the implementation using `#include "file.H"`.

Idea:
- Put the parts that are version-specific into a separate directory `variants/<variant>/version-specific/`.
- Include them in the middle of the implementation, e.g. as `#include "version-specific/file.H"`.
- Leave a comment with a concrete implementation example for readability (they must be short anyway).
- As `<variant>` we always take the `$WM_PROJECT_VERSION`. We only treat it as a string as it varies a lot among versions.
- We need a directory for every version. We define the files in the oldest supported version and create links for newer versions (to the complete directory and/or individual files).

Pros & Cons of this approach:
- (+) No need for the user to change branches, always have the latest changes in all variants
- (+) Easily extensible: Just find the closest version and copy the directory or make a link
   - Supporting a new version should usually mean "create a link to the previous version".
- (+) Avoids duplication: Only define the patch in one file, link from other versions
- (+/-) Clearly define which versions we support
- (-) For a new version we always need a new directory/link --> users should get the latest adapter or patch the change
- (-) To fix one version in a new location we need to update all the variants.
- (-) Reduced readability/refactoring capability
   - But OpenFOAM itself already uses such a preprocessor-based approach
   - and we keep a comment of the implementation
   - we will have an issue if fundamental functions we use all the time start changing (e.g. older `.boundaryField()` and `.boundaryFieldRef()`.

Todo:
- [ ] Implement a warning/error for unrecognized versions
   - Can we assume and handle a "default" variant?
- [x] Port changes from #33 (OpenFOAM 6) (closes #33)
- [x] Port changes from  #91 (OpenFOAM 7) (closes #91)
- [ ] Port changes from #36 (OpenFOAM-dev) (closes #36)
   - Open questions: are there issues with openfoam.org/dev and openfoam.com/dev?
- [ ] Implement changes from #120 (OpenFOAM 4) (closes #120)
- [ ] Check/update bundled tutorial scripts

Before merging, check that the final version builds with:
   - [ ] OpenFOAM 4.1
   - [ ] OpenFOAM 5.x
   - [ ] OpenFOAM 6
   - [ ] OpenFOAM 7
   - [ ] OpenFOAM-dev
   - [ ] OpenFOAM v1706
   - [ ] OpenFOAM v1712
   - [ ] OpenFOAM v1806
   - [ ] OpenFOAM v1812
   - [ ] OpenFOAM v1906
   - [ ] OpenFOAM v1912
   - [ ] OpenFOAM v2006

External updates:
- [ ] Check/update tutorial scripts (especially FSI tutorials)
- [ ] Update wiki page [Notes on OpenFOAM](https://github.com/precice/openfoam-adapter/wiki/Notes-on-OpenFOAM)
- [ ] Update system tests

Possible next steps:
- Automatically decide between `pimpleFoam` and `pimpleDyMFoam` in the tutorial scripts. Maybe simply apply a special case for the oldest supported versions.


MakisH:
Since we decided (#52) to keep support for the Foundation branches and release archives for each range of supported versions, I am closing this for now. It is an idea that we may revive in the future, though.

Issue #129, opened 2020-05-19
Broken in OpenFOAM-dev: TurbulenceModels library replaced by MomentumTransportModels


MakisH:
We have a problem with radical changes in OpenFOAM.org, already released in OpenFOAM-dev 20200426 and mainly originating from [this commit](https://github.com/OpenFOAM/OpenFOAM-dev/commit/de66b1be68b117f69075758e89ae08a224cc9d4b).

Trying to build the adapter currently fails with:
```
In file included from CHT/KappaEffective.C:1:0:
CHT/KappaEffective.H:4:39: fatal error: turbulentFluidThermoModel.H: No such file or directory
```
I am currently not sure how much this change affects us, but it could be anywhere from "a little" to "vastly".

As a starting points, I see this change everywhere:
```diff
- #include "turbulentFluidThermoModel.H"
+ #include "fluidThermoMomentumTransportModel.H"
```
as well as in `Make/options`:
```diff
- -I$(LIB_SRC)/TurbulenceModels/turbulenceModels/lnInclude \
- -I$(LIB_SRC)/TurbulenceModels/compressible/lnInclude \
+ -I$(LIB_SRC)/MomentumTransportModels/momentumTransportModels/lnInclude \
+ -I$(LIB_SRC)/MomentumTransportModels/compressible/lnInclude \
```
and
```diff
- -lturbulenceModels \
- -lcompressibleTurbulenceModels \
+ -lmomentumTransportModels \
+ -lfluidThermoMomentumTransportModels \
```
___
Another related error I got:
```
In file included from CHT/KappaEffective.C:1:0:
CHT/KappaEffective.H:6:37: fatal error: turbulentTransportModel.H: No such file or directory
```

Related changes:
```diff
- #include "turbulentTransportModel.H"
+ #include "kinematicMomentumTransportModel.H"
```
as well as in the `Make/options`:
```diff
- -I$(LIB_SRC)/TurbulenceModels/incompressible/lnInclude \
- -I$(LIB_SRC)/transportModels \
- -I$(LIB_SRC)/transportModels/incompressible/singlePhaseTransportModel \
+ -I$(LIB_SRC)/MomentumTransportModels/incompressible/lnInclude \
+ -I$(LIB_SRC)/transportModels/lnInclude \
```
and
```diff
- -lincompressibleTurbulenceModels \
+ -lincompressibleMomentumTransportModels \
```

___

And then, of course, we need to update how we access `kappaEff`, `nu`, etc:
```
[...]

PR #130, opened 2020-05-22
Adapt KappaEffective to new transport/thermophysical model structure (OpenFOAM 8)

MakisH:
An work-in-progress attempt to port the adapter to the latest (April/May 2020) changes in OpenFOAM-dev (openfoam.org).

This is related to #129.

### DIFF
--- a/CHT/KappaEffective.C
+++ b/CHT/KappaEffective.C
@@ -10,8 +10,8 @@ using namespace Foam;
 preciceAdapter::CHT::KappaEff_Compressible::KappaEff_Compressible(
     const Foam::fvMesh& mesh)
 : mesh_(mesh),
-  turbulence_(
-      mesh.lookupObject<compressible::turbulenceModel>(turbulenceModel::propertiesName))
+  thermophysicalTransportModel_(
+      mesh.lookupObject<thermophysicalTransportModel>(thermophysicalTransportModel::typeName))
 {
     DEBUG(adapterInfo("Constructed KappaEff_Compressible."));
 }
@@ -24,12 +24,12 @@ void preciceAdapter::CHT::KappaEff_Compressible::extract(uint patchID, bool mesh
         primitivePatchInterpolation patchInterpolator(mesh_.boundaryMesh()[patchID]);
 
         //Interpolate kappaEff_ from centers to nodes
-        kappaEff_ = patchInterpolator.faceToPointInterpolate(turbulence_.kappaEff()().boundaryField()[patchID]);
+        kappaEff_ = patchInterpolator.faceToPointInterpolate(thermophysicalTransportModel_.kappaEff()().boundaryField()[patchID]);
     }
     else
     {
-        // Extract kappaEff_ from the turbulence model
-        kappaEff_ = turbulence_.kappaEff()().boundaryField()[patchID];
+        // Extract kappaEff_ from the thermophysical transport model
+        kappaEff_ = thermophysicalTransportModel_.kappaEff()().boundaryField()[patchID];
     }
 }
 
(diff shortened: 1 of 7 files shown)


CHT/CHT.C
```diff
@@ -104,16 +104,14 @@ std::string preciceAdapter::CHT::ConjugateHeatTransfer::determineSolverType()
         DEBUG(adapterInfo("Did not find the transportProperties dictionary."));
     }
 
-    if (mesh_.foundObject<IOdictionary>(turbulenceModel::propertiesName))
+    if (mesh_.foundObject<IOdictionary>("turbulenceProperties"))
```
MakisH, line 107: The new name is `turbulenceModel::typeName`. However, hard-coding this makes it easier to work across OpenFOAM versions.

MakisH:
A summary of the changes needed (I will probably need to edit this as I better understand the changes):
- `turbulenceModel` is now `momentumTransportModel`, again in `compressible::` and `incompressible::` forms.
- `turbulentTransportModel.H` is now `kinematicMomentumTransportModel.H`
- `turbulentFluidThermoModel.H` is now `fluidThermoMomentumTransportModel.H` + `thermophysicalTransportModel.H` (I need to check this further, but we need both files)
- (todo: more)

Current state:
- Adapter compiles, but still reports a weird error at the end (see open questions)
- CHT tutorials do not run.

Open questions/issues:
- [ ] `Allwmake` fails with an error that is not included in `wmake.log`. Running simple `wmake` ends with:
```
wmakeLnIncludeAll: running wmakeLnInclude on dependent libraries:
    wmakeLnInclude error: base directory /opt/openfoam8/src/transportModels/incompressible/ does not exist
    wmakeLnInclude error: base directory /opt/openfoam8/src/transportModels/compressible/ does not exist
    wmakeLnInclude error: base directory /opt/openfoam8/src/transportModels/twoPhaseMixture/ does not exist
    wmakeLnInclude error: base directory /opt/openfoam8/src/transportModels/interfaceProperties/ does not exist
mkdir: cannot create directory ‘’: No such file or directory
/opt/openfoam8/wmake/makefiles/general:167: recipe for target '/libpreciceAdapterFunctionObject.so' failed
make: *** [/libpreciceAdapterFunctionObject.so] Error 1
```
- [ ] Runtime error in the Fluid side:
```
lookup of thermophysicalProperties from objectRegistry region0 successful
    but it is not a thermophysicalTransport, it is a heRhoThermo<pureMixture<const<hConst<perfectGas<specie>>,sensibleEnthalpy>>>
```
- [ ] The solver type guessing does not work.
- [ ] At the end of the simulation there is a strange runtime error:
```
---[preciceAdapter] The simulation was ended by preCICE. Calling the end() methods of any functionObject explicitly.
End

Inconsistency detected by ld.so: dl-close.c: 811: _dl_close: Assertion `map->l_init_called' failed!
```
- [ ] After these changes, do we still need the turbulence-related parts?

CHT/KappaEffective.C
```diff
@@ -47,136 +47,32 @@ scalar preciceAdapter::CHT::KappaEff_Compressible::getAt(int i)
...
-        kappaEff_= patchInterpolator.faceToPointInterpolate(kappaEff_temp);
+        //Interpolate kappaEff_ from centers to nodes
+        kappaEff_ = patchInterpolator.faceToPointInterpolate(thermophysicalTransportModel_.kappaEff()().boundaryField()[patchID]);
     }
     else
     {
-        // if no interpolation
-        kappaEff_ = kappaEff_temp;
+        // Extract kappaEff_ from the thermophysical transport model
+        kappaEff_ = thermophysicalTransportModel_.kappaEff()().boundaryField()[patchID];
```
MakisH, line 74: With this change, we don't need to distinguish between incompressible and compressible solvers anymore. This raises new possibilities for simplifying the code, but also new questions on how to deal with different OpenFOAM versions (see #32 #128 ).

TEFEdotCC:
#144 adaption nedded?

DavidSCN:
I ran this case with OpenFOAM 8 and the flap_perp deal.II tutorial on my ubuntu 2004 machine. Compilation and running the case works without any error. Maybe the error was related to #134. Also, the `flow-over-plate/buoyantPimpleFoam-laplacianFoam` case runs flawlessly without any issue. 
Which system do you use/ how can I exactly reproduce the linking error?

MakisH:
> I ran this case with OpenFOAM 8 and the flap_perp deal.II tutorial on my ubuntu 2004 machine. Compilation and running the case works without any error. Maybe the error was related to #134. Also, the `flow-over-plate/buoyantPimpleFoam-laplacianFoam` case runs flawlessly without any issue.
> Which system do you use/ how can I exactly reproduce the linking error?

I got this error on Ubuntu 16.04. I will also check in 20.04 then, maybe this is compiler-related (but still very weird).

Did you run the simulation until the very end?

It should not be related to #134, as the commits are already merged in this branch.

MakisH:
I confirm that it runs smoothly on Ubuntu 20.04 and the latest build of OpenFOAM 8, for the bundled CHT tutorial.

@DavidSCN could you please have a look in the code and check if it makes sense?

DavidSCN (commented):
Seems like there changed quite a lot internally in OpenFOAM. Without looking at the OpenFOAM documentation more specifically, it is hard to say if everything works as expected. It would be nice to have passing tests for all OF related tutorial cases.

MakisH:
@DavidSCN how should we proceed from here?

DavidSCN:
Could we run some tests on it in order to compare the results exactly? IIRC there are some unit tests in progress and we should have some coupled tests on travis as well. I would suggest to reproduce the results from the tests in order to prove the functionality. Maybe we can review the code afterwards and merge it.

MakisH:
@DavidSCN the unit tests are nowhere near being usable right now. The system tests may also fail as we don't have reference data for OpenFOAM 8. I would only rely on manual tests for now.

Would you maybe have time to compare 1-2 cases (maybe heated plate and perpendicular flap, including a case in parallel) and have a look at the results? If not, I can also do it at some point.

DavidSCN:
Hm, I would rather invest the time in order to set up some tests and verify this PR using the tests since these issues will constantly appear. I've recently done something very similar for the deal.II cases.

MakisH:
I updated the branch with the latest state of `master` where I needed to resolve a few conflicts, but I have not yet compiled it. If you try this and it does not compile, please comment here.

MakisH:
I fixed some building issues introduced while resolving the conflicts and thi sbranch now builds with OpenFOAM 8.

I tried to run all of our tutorials. One case does not work:
- `elastic-tube-3d`: The CalculiX side fails early with:
  ```log
  Adapter calling advance()...
  ---[precice]  relative convergence measure: relative two-norm diff = inf, limit = 0.001, normalization = 0, conv = true
  ---[precice]  relative convergence measure: relative two-norm diff = inf, limit = 0.001, normalization = 0, conv = true
  ---[precice]  All converged
  ---[precice] WARNING:  The coupling residual equals almost zero. There is maybe something wrong in your adapter. Maybe you always write the same data or you call advance without providing new data first or you do not use available read data. Or you just converge much further than actually necessary.
  ASSERTION FAILED
    Location:          void precice::acceleration::impl::Preconditioner::apply(Eigen::MatrixXd&)
    File:              /home/makish/github/precice_develop/src/acceleration/impl/Preconditioner.hpp:107
    Rank:              0
    Failed expression: M.rows() == (int) _weights.size()
  ```
  Log files:  [fluid-openfoam.log](https://github.com/precice/openfoam-adapter/files/6363722/fluid-openfoam.log), [solid-calculix.log](https://github.com/precice/openfoam-adapter/files/6363725/solid-calculix.log). I do not get the same behavior with OpenFOAM v2006.

The following cases work and the results are visually ok:
- CHT: `flow-over-heated-plate` (`fluid-openfoam`, `solid-openfoam`)
- CHT: `flow-over-heated-plate-nearest-projection` (`fluid-openfoam`, `solid-openfoam`)
- FSI: `perpendicular-flap`
- FF: `partitioned-pipe` (`pimpleFoam`) - OpenFOAM 7 already removed sonicLiquidFoam (see [solvers in OpenFOAM 8](https://cfd.direct/openfoam/user-guide/v8-standard-solvers/))
- FSI: `quickstart`

I cannot test:
- CHT: `flow-over-heated-plate-steady-state` (starts ok, I don't have code_aster installed)
- FSI: `multiple-perpendicular-flaps` (starts ok, my deal.II adapter segfaults in the very beginning, I don't want to debug it now)
- FSI: `turek-hron-fsi3` (I don't have groovyBC installed)

At the end of every OpenFOAM simulation, I get the following error:

```log
Inconsistency detected by ld.so: dl-close.c: 811: _dl_close: Assertion `map->l_init_called' failed!
```

I tested everything with their current develop branches (preCICE v2.2.0+later).

MakisH:
I tested again the `elastic-tube-3d`. The exported VTK files don't show anything helpful, as I only get the initialization, even with  `every-iteration="true"`.

While it is CalculiX that explodes, in the first `advance()` call, it does run with OpenFOAM 7, so it has to be something with the force calculation in OpenFOAM (see file `ForceBase.C` here, @DavidSCN any idea?). Important is that I had to merge our previous implementation of `Force.C`, our current implementation with `ForceBase.C`, and the OpenFOAM8-specific changes.

@KyleDavisSA How is this case different than the `perpendicular-flap`?

| Feature | `elastic-tube-3d` | `perpendicular-flap` | Notes |
--- | --- | --- | ---
Exchanged quantities | Force + DisplacementDelta | Force + Displacement |
Meshes | 
Mapping | 2x nearest projection | 2x RBF | Tried RBF, did not help.
Coupling scheme | serial-implicit | parallel-implicit | Tried parallel-implicit and serial-explicit, did not help.
Acceleration | IQN-IMVJ with residual-sum preconditioner | same |

Any help would be welcome here.OpenFOAM 8 is very easy to install on Ubuntu: https://openfoam.org/download/8-ubuntu/

DavidSCN:
> While it is CalculiX that explodes, in the first advance() call, it does run with OpenFOAM 7, so it has to be something with the force calculation in OpenFOAM (see file ForceBase.C here, @DavidSCN any idea?). Important is that I had to merge our previous implementation of Force.C, our current implementation with ForceBase.C, and the OpenFOAM8-specific changes.

Will have a look at this.

DavidSCN:
I installed OF8 and ran the elastic tube now with this branch and deal.II and it works for me. There is only one issue: OpenFOAM 8 handles the boundary conditions (?) differently. In fact, the applied load/stress in the very first iteration is (as opposed to ESI) zero, which is probably also the reason for the fail in calculix. Since it is zero, the very first iteration converges immediately so I'm not sure if it is only the first iteration or the first complete time step. My best guess is at the moment that the definition of the boundary conditions in the pressure is not correct/up-to-date (maybe even for ESI).  My paraview version complains here also when trying to visualize time `0`.

MakisH:
> My best guess is at the moment that the definition of the boundary conditions in the pressure is not correct/up-to-date (maybe even for ESI). My paraview version complains here also when trying to visualize time `0`.

Would this mean that we only need to change something in the tutorial case, then? Or is this probably something in the adapter?

> Btw OpenFOAM 8 reminds me again why we had the `0.orig` directories sweat_smile

I did not notice that. Does it overwrite our `0/` files?

DavidSCN:
As pointed out by @KyleDavisSA our OpenFOAM 8 (and dev)  adapter branches do not compile without this PR. We should merge it or drop the branches.

MakisH:
We should definitely merge it, but what is the current state? Does this now work as expected if we rebase on `OpenFOAM8` (which I guess already includes #179)?

DavidSCN:
> We should definitely merge it, but what is the current state?

Not sure, I guess we need some manual testing as we don't have any tests, right?

> Does this now work as expected if we rebase on OpenFOAM8 (which I guess already includes #179)?

Yes, we definitely need to rebase or merge in order to incorporate #179

MakisH:
Merging this branch onto `OpenFOAM8` works. The `flow-over-heated-plate` and `elastic-tube-3d` cases complete without any modifications and their results look reasonable.

To move forward (especially to get input in the context of #199 ), let's merge this and report any potential issues separately.

Directly after merging this, I will merge `OpenFOAMdev` to `OpenFOAM8` and squash the version-specific commits, as we currently do in the other branches as well.

PR #132, opened 2020-06-29
Replace non-const boundaryFieldRef wherever possible

MakisH:
OpenFOAM provides the methods boundaryField and boundaryFieldRef to return references to the boundary of a field. Their difference is that boundaryField returns a const reference, while boundaryFieldRef returns a non-const reference.
    
There are a few parts of the code where we were using the non-const versions where the const would be enough. This replaces boundaryFieldRef by boundaryField, essentially keeping only one non-const instance at the read() method of every coupling data reader.

This depends on #131.

@DavidSCN have I forgotten anything?

### DIFF
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -33,11 +33,11 @@ void preciceAdapter::CHT::HeatFlux::write(double * buffer, bool meshConnectivity
     {
         int patchID = patchIDs_.at(j);
 
-        scalarField gradientPatch
+        const scalarField & gradientPatch
         (
-            refCast<fixedValueFvPatchScalarField>
+            refCast<const fixedValueFvPatchScalarField>
             (
-                T_->boundaryFieldRef()[patchID]
+                T_->boundaryField()[patchID]
             ).snGrad()
         );
 
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -33,9 +33,13 @@ void preciceAdapter::CHT::HeatFlux::write(double * buffer, bool meshConnectivity
     {
         int patchID = patchIDs_.at(j);
 
-        scalarField gradientPatch=refCast<fixedValueFvPatchScalarField>
-                (T_->boundaryFieldRef()[patchID]
-                 ).snGrad();
+        scalarField gradientPatch
+        (
+            refCast<fixedValueFvPatchScalarField>
+            (
+                T_->boundaryFieldRef()[patchID]
+            ).snGrad()
+        );
 
         // Extract the effective conductivity on the patch
         extractKappaEff(patchID, meshConnectivity);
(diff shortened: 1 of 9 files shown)


MakisH:
Building is ok, but there are differences in the results, which confuse me. Either I introduced some bug, or I resolved one that we didn't know. I have to look closer into this.

TEFEdotCC:
#144 solved all request's?

DavidSCN:
> Building is ok, but there are differences in the results, which confuse me. Either I introduced some bug, or I resolved one that we didn't know. I have to look closer into this.

Could you describe the steps to reproduce this issue. I ran a simple FSI case and couldn't observe any differences.

MakisH:
> > Building is ok, but there are differences in the results, which confuse me. Either I introduced some bug, or I resolved one that we didn't know. I have to look closer into this.
> 
> Could you describe the steps to reproduce this issue. I ran a simple FSI case and couldn't observe any differences.

Yes, sorry. I ran the "flow over a heated plate" pimpleFoam-laplacianFoam tutorial with this branch and with develop and then compared the results with e.g. Meld (diff tool). The overall picture still looked reasonable, but I could not explain the differences in the results.

(we definitely need unit tests)

CHT/HeatFlux.C
```diff
@@ -33,9 +33,13 @@ void preciceAdapter::CHT::HeatFlux::write(double * buffer, bool meshConnectivity
     {
         int patchID = patchIDs_.at(j);
 
-        scalarField gradientPatch=refCast<fixedValueFvPatchScalarField>
-                (T_->boundaryFieldRef()[patchID]
-                 ).snGrad();
+        const scalarField & gradientPatch
```
DavidSCN, line 36:
```suggestion
        const scalarField gradientPatch
```
You cannot take the reference here, because the `gradientPatch` is just a temporary created field. We need an explicit copy.

DavidSCN (changes requested):
I tested it now with the suggested fix and obtained (also with meld) the same results on the Fluid side for the `flow-over-heated-plate`. I work, however, with OF7. Maybe you could test it again with your OF version.

CHT/HeatFlux.C
MakisH: Very good point, thank you! Sometimes all we need is a second pair of eyes... :smile:

MakisH:
Thank you for the fix! I don't see the same problem anywhere else.

I now also confirm that it gives identical results for the `tutorials/CHT/flow-over-plate/buoyantPimpleFoam-laplacianFoam` serial case. Let's do the same check for 1-2 more cases and it should be good to merge.

DavidSCN (approved):
I checked the CHT nearest-projection case and could not observe any differences before. The FSI/stress module has successfully been checked before. This covers already most of the changes.
General question: why have you adjusted the formatting in so many places? Just coding style or do we have finally some format file?

MakisH:
> General question: why have you adjusted the formatting in so many places? Just coding style or do we have finally some format file?

The formatting was already mixed and I introduced even more changes recently. The plan is to add a clang-format file soon and reformat all branches.

MakisH:
I also checked the flap_perp/OpenFOAM-CalculiX (with forces). Thanks!

MakisH:
> > General question: why have you adjusted the formatting in so many places? Just coding style or do we have finally some format file?
> 
> The formatting was already mixed and I introduced even more changes recently. The plan is to add a clang-format file soon and reformat all branches.

actually, in this specific PR, I manually reformatted the parts I touched to make them more readable and consistent.

PR #134, opened 2020-07-09
Fix Force_ delete in Stress inheritance

### LINKED ISSUE #133: Bus error in adapter destruction
Currently working with the `Stress` data field and OpenFOAM 7, I face the following error, when the simulation has finished: 
```
---[preciceAdapter] [DEBUG] Destroying the preCICE solver interface...
---[preciceAdapter] [DEBUG] Deleting the interfaces...

Bus error (core dumped)
```

This error was reported by @Eder-K in the [recent system tests](https://travis-ci.org/github/precice/systemtests/jobs/705575276#L15435) and I can reproduce it locally. 

I still need to investigate, if other data fields are affected as well, I guess there is something wrong in the destructors.

DavidSCN:
Resolves #133

### DIFF
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -33,9 +33,13 @@ void preciceAdapter::CHT::HeatFlux::write(double * buffer, bool meshConnectivity
     {
         int patchID = patchIDs_.at(j);
 
-        scalarField gradientPatch=refCast<fixedValueFvPatchScalarField>
-                (T_->boundaryFieldRef()[patchID]
-                 ).snGrad();
+        scalarField gradientPatch
+        (
+            refCast<fixedValueFvPatchScalarField>
+            (
+                T_->boundaryFieldRef()[patchID]
+            ).snGrad()
+        );
 
         // Extract the effective conductivity on the patch
         extractKappaEff(patchID, meshConnectivity);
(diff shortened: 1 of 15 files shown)


DavidSCN:
Got it now, the problem was the delete on the force field. I introduced another constructor (when created the stress) to avoid it's allocation, since it might be large. Therefore, I added now a boolean check in the destructor.

DavidSCN:
> this is a hint that we need some change in the class hierarchy, not additional variables.

We could of course create a base class for both, Stress and Force to be fully consistent and circumvent the problem here. Also, we could introduce smart pointers, but this should be done consistently in all fields. These are all larger changes and more dedicated to general design principles, so, I don't like to spend more time working on this. However, note that the deal.II and openfoam systemtests fail due to this issue. 

You could either close this PR completely (it is linked in the issue anyway) or keep it open until a better solution is presented.

DavidSCN:
@Eder-K does it fix the systemtest?

Eder-K (approved):

DavidSCN:
@M

DavidSCN:
@MakisH  I cannot merge for some reason, so please merge

MakisH:
Somehow this was stuck in a `WIP` check. Everything looked good, so I merged it.

Eder-K:
@MakisH could you also update the develop branch to include this change?

MakisH:
@Eder-K I updated `develop`. I did not see that this was targetting `master`, and it has happened to me a few times. We need to fix this somehow. But we have a few more important things to fix first.

PR #147, opened 2020-12-08
Enable `faceCenters` for Displacement based reading

### LINKED ISSUE #146: Extend the 2D mode to only set one layer of vertices
In https://github.com/precice/openfoam-adapter/pull/96, we added support for 2D solver interfaces. In that case, the adapter can skip the third dimension in vertices and data, such as here:

https://github.com/precice/openfoam-adapter/blob/8ae5f061ab26ae059b8cdcf304089306f9ce7227/Interface.C#L142-L148

However, we still pass all the patch nodes to preCICE:

https://github.com/precice/openfoam-adapter/blob/8ae5f061ab26ae059b8cdcf304089306f9ce7227/Interface.C#L105

This works nicely, but it can still lead to the RBF mapping issue described in https://github.com/precice/tutorials/issues/112.

What we need is to distinguish the two layers of nodes and only select the one (usually "front" or "back"). Alternatives I can think of:
- maybe we can do this directly with an OpenFOAM method (I don't know if such a method exists, but this would be ideal)
- we somehow select points that belong to two patches: the interface and the front or back "empty" boundary.
- we manually find the two repeating z (or x, or y) values and add only the points we want to a list. I can imagine that this solution would be quite difficult to maintain.

In any case, we should see how each alternative affects the rest of the code. If we need to manually find the correct points every time we write/read data, that would be risky.

This is a requirement for https://github.com/precice/tutorials/issues/117.

DavidSCN:
Resolves #146 
However, it comes at the price of a repeated memory allocation for the intermediate 'center based' field in order to transform the buffer into an OpenFOAM suitable format. A variant similar to https://github.com/DavidSCN/openfoam-adapter/pull/1 would be much better.

Open question: how should we proceed in 3D? Should we stick to this variant or should be fall back to the `faceNodes`? Also, I'm still not sure about the formatting, but afaik we don't have standard yet.

### DIFF
--- a/FSI/Displacement.C
+++ b/FSI/Displacement.C
@@ -16,7 +16,7 @@ preciceAdapter::FSI::Displacement::Displacement(
     dataType_ = vector;
 }
 
-// We cannot do this steps in the constructor by design of the adapter since the information of the CouplingDataUser is
+// We cannot do this step in the constructor by design of the adapter since the information of the CouplingDataUser is
 // defined later. Hence, we call this method after the CouplingDaaUser has been configured
 void preciceAdapter::FSI::Displacement::initialize()
 {
--- a/FSI/Displacement.C
+++ b/FSI/Displacement.C
@@ -57,7 +57,7 @@ void preciceAdapter::FSI::Displacement::read(double * buffer, const unsigned int
           // the boundaryCellDisplacement is a vector and ordered according to the iterator j
           // and not according to the patchID
           // First, copy the buffer data into the center based vectorFields on each interface patch
-          for (int i = 0; i < cellDisplacement_->boundaryField()[patchID].size(); ++i) {
+          forAll (cellDisplacement_->boundaryField()[patchID], i) {
             for (unsigned int d = 0; d < dim; ++d)
               cellDisplacement_->boundaryFieldRef()[patchID][i][d] = buffer[i * dim + d];
           }
@@ -78,7 +78,7 @@ void preciceAdapter::FSI::Displacement::read(double * buffer, const unsigned int
                   pointDisplacement_->boundaryFieldRef()[patchID]));
 
           // Overwrite the nodes on the interface directly
-          for (int i = 0; i < pointDisplacement_->boundaryFieldRef()[patchID].size(); ++i) {
+          forAll (pointDisplacement_->boundaryFieldRef()[patchID], i) {
             for (unsigned int d = 0; d < dim; ++d)
               pointDisplacementFluidPatch[i][d] = buffer[i * dim + d];
           }
--- a/FSI/Displacement.C
+++ b/FSI/Displacement.C
@@ -53,30 +53,18 @@ void preciceAdapter::FSI::Displacement::read(double * buffer, const unsigned int
                 cellDisplacement[i][2] = buffer[bufferIndex++];
         }
 
-        //Interpolate from centers to nodes
-        primitivePatchInterpolation patchInterpolator(mesh_.boundaryMesh()[patchID]);
-        vectorField pointDisplacement
-        (
-            patchInterpolator.faceToPointInterpolate(cellDisplacement)
-        );
-
         // Get the displacement on the patch
-        fixedValuePointPatchVectorField& pointDisplacementFluidPatch
+        vectorField& pointDisplacementFluidPatch
         (
-            refCast<fixedValuePointPatchVectorField>
+            refCast<vectorField>
             (
                 pointDisplacement_->boundaryFieldRef()[patchID]
             )
         );
 
-        // For every cell of the patch
-        forAll(pointDisplacement_->boundaryFieldRef()[patchID], i)
-        {
-            // Set the displacement to the received one
-            pointDisplacementFluidPatch[i][0] = pointDisplacement[i][0];
-            pointDisplacementFluidPatch[i][1] = pointDisplacement[i][1];
-            if(dim ==3)
-                pointDisplacementFluidPatch[i][2] = pointDisplacement[i][2];
-        }
+        //Interpolate from centers to nodes
+        primitivePatchInterpolation patchInterpolator(mesh_.boundaryMesh()[patchID]);
+        pointDisplacementFluidPatch =
+             patchInterpolator.faceToPointInterpolate(cellDisplacement);
     }
 }
--- a/FSI/Displacement.C
+++ b/FSI/Displacement.C
@@ -19,6 +19,26 @@ mesh_(mesh)
     dataType_ = vector;
 }
 
+// We cannot do this steps in the constructor by design of the adapter since the information of the CouplingDataUser is
+// defined later. Hence, we call this method after the CouplingDaaUser has been configured
+void preciceAdapter::FSI::Displacement::initialize()
+{
+  // We use the cellDisplacement in order to copy the interface boundary fields
+  const volVectorField &cellDisplacement_ =
+      mesh_.lookupObject<volVectorField>("cellDisplacement");
+
+  // Initialize appropriate objects for each interface patch, namely the volField and the interpolation object
+  // this is only necessary for face based FSI
+  if (this->locationsType_ == "faceCenters" || this->locationsType_ == "faceCentres")
+    for (unsigned int j = 0; j < patchIDs_.size(); j++) {
+      const unsigned int patchID = patchIDs_.at(j);
+      boundaryCellDisplacement_.emplace_back(cellDisplacement_.boundaryField()[patchID]);
+      interpolationObjects_.emplace_back(new primitivePatchInterpolation(mesh_.boundaryMesh()[patchID]));
+    }
+}
+
+
+
 void preciceAdapter::FSI::Displacement::write(double * buffer, bool meshConnectivity, const unsigned int dim)
 {
     /* TODO: Implement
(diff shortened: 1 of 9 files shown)


MakisH:
> However, it comes at the price of a repeated memory allocation for the intermediate 'center based' field in order to transform the buffer into an OpenFOAM suitable format.

Maybe we could make `cellDisplacement` a member of `Displacement`, allocate it in the constructor, and update it in the `read`?

> A variant similar to [DavidSCN#1](https://github.com/DavidSCN/openfoam-adapter/pull/1) would be much better.

Could you explain how these two PRs relate to each other? It is not obvious to me why that one resolves the issue (and which exactly).

> Open question: how should we proceed in 3D? Should we stick to this variant or should be fall back to the `faceNodes`?

We are essentially changing the semantics of the `read` method: we still read Displacement, but we read it at a different location.
- Supporting both would be quite painful and without a clear benefit.
- Having 2D:cellDisplacement and 3D:cellDisplacement would be more consistent and cleaner. However, it does an additional interpolation step in the 3D case (which then makes the role of the preCICE mapping unclear).
- Having 2D:cellDisplacement and 3D:pointDisplacement would be inconsistent, but we can indicate that the special case is the 2D (already quite special for OpenFOAM). Then people know that if they e.g. design conforming meshes, the displacements are still mapped exactly. So, no surprises for the 3D FSI people.

I think we need to at least compare the results of both methods before we decide. The 3D tube could be an example to test for the 3D case.

> Also, I'm still not sure about the formatting, but afaik we don't have standard yet.

We don't have one, don't worry for now. It looks clean anyway.

FSI/DisplacementDelta.C
```diff
@@ -41,23 +42,29 @@ void preciceAdapter::FSI::DisplacementDelta::read(double * buffer, const unsigne
...
 
+        const volVectorField &tmp_field = mesh_.lookupObject<volVectorField>("cellDisplacement");
+        vectorField           cellDisplacement(tmp_field.boundaryField()[patchID]);
+        forAll(cellDisplacement, i)
+        {
+            // Set the displacement to the received one
+            cellDisplacement[i][0] = buffer[bufferIndex++];
+            cellDisplacement[i][1] = buffer[bufferIndex++];
+            if(dim ==3)
+                cellDisplacement[i][2] = buffer[bufferIndex++];
```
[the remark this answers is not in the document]
DavidSCN: Copy error.

FSI/Displacement.C
```diff
@@ -41,23 +42,29 @@ void preciceAdapter::FSI::Displacement::read(double * buffer, const unsigned int
...
-            // Set the displacement to the received one
-            pointDisplacementFluidPatch[i][0] = buffer[bufferIndex++];
-            pointDisplacementFluidPatch[i][1] = buffer[bufferIndex++];
-            if(dim ==3)
-                pointDisplacementFluidPatch[i][2] = buffer[bufferIndex++];
-        }
+        //Interpolate from centers to nodes
+        primitivePatchInterpolation patchInterpolator(mesh_.boundaryMesh()[patchID]);
+        pointDisplacementFluidPatch =
+             patchInterpolator.faceToPointInterpolate(cellDisplacement);
```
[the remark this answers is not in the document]
DavidSCN: > Do I understand this change correctly?
> 
> To read Displacement:
> 
>     * Before: get the `boundaryFieldRef` of `pointDisplacement` and assign it buffer values.

Correct. 
> 
>     * Now: get the `boundaryField` (why?) of  `cellDisplacement` and assign it buffer values. Then, interpolate the `cellDisplacement` from the face centers to `pointDisplacement` on the face nodes.
> 
It's probably a confusing choice. I just want to use an appropriate constructor for the temporary `volVectorField` in order to allocate the correct size (it's not a reference but a copy). Every other `volVectorField` would be good as well. The `cellDisplacement` was just an obvious choice. There is probably a better way of doing this.

>     * Additionally: replace the more specific type `fixedValuePointPatchVectorField` with the more general `vectorField` (why?)

Compatibility with the `primitivePatchInterpolation` object. Otherwise I need another temporary field in order to assign values from the interpolated temporary `volVectorField` to the `fixedValuePointPatchVectorField`.

DavidSCN:
> > However, it comes at the price of a repeated memory allocation for the intermediate 'center based' field in order to transform the buffer into an OpenFOAM suitable format.
> 
> Maybe we could make `cellDisplacement` a member of `Displacement`, allocate it in the constructor, and update it in the `read`?
> 
In general, allocating some field in the constructor makes sense. 

> > A variant similar to [DavidSCN#1](https://github.com/DavidSCN/openfoam-adapter/pull/1) would be much better.
> 
> Could you explain how these two PRs relate to each other? It is not obvious to me why that one resolves the issue (and which exactly).
> 

That 'one resolves the issue'? Which one do you mean? Strictly speaking, both resolve the issue of 'duplicated nodes in 2D', but with another methodology: The one presented in this PR ensures consistency with the OF methods as we use them right now. This means we set our boundary conditions in the `pointDisplacementField` (which is also the file specified in the initial condition directory). In order to handle this in the adapter, we use OF interpolation methods so that the user specifies `faceCenters` as the data location and we interpolate from the face centers to face nodes in the adapter. 
In the other method, we directly enforce the boundary condition on the `cellDisplacement` field since the mesh displacement is anyway solved based on the `cellDisplacement`. We more or less 'ignore' the `pointDisplacement` field. However, this is not the native OF way and I need to check where apparent differences come from.

> > Open question: how should we proceed in 3D? Should we stick to this variant or should be fall back to the `faceNodes`?
> 
> We are essentially changing the semantics of the `read` method: we still read Displacement, but we read it at a different location.
> 
>     * Supporting both would be quite painful and without a clear benefit.
> 
Agree.

>     * Having 2D:cellDisplacement and 3D:cellDisplacement would be more consistent and cleaner. However, it does an additional interpolation step in the 3D case (which then makes the role of the preCICE mapping unclear).
> 
I guess the main point here is just the performance drawback in case of larger 3D cases.

>     * Having 2D:cellDisplacement and 3D:pointDisplacement would be inconsistent, but we can indicate that the special case is the 2D (already quite special for OpenFOAM). Then people know that if they e.g. design conforming meshes, the displacements are still mapped exactly. So, no surprises for the 3D FSI people.
> 

Maybe we can support only in 3D both methodologies so that it is still easy to configure if you don't care about performance, but changes are required if you would like to skip the additional interpolation.

> 
> I think we need to at least compare the results of both methods before we decide. The 3D tube could be an example to test for the 3D case.

If we decide to change things for the 3D case as well, yes. Otherwise the perpendicular flap should be fine (which I cheched at least visually).

MakisH:
> > I think we need to at least compare the results of both methods before we decide. The 3D tube could be an example to test for the 3D case.
> 
> If we decide to change things for the 3D case as well, yes. Otherwise the perpendicular flap should be fine (which I cheched at least visually).

But with this PR we are adding the cellDisplacement -> pointDisplacement step in both the 2D and 3D, right?

I would favor to just keep the 3D as it is, which (as I understood) is also your opinion. But I am not sure I understand what is happening in the 3D case then.

DavidSCN:
> > > I think we need to at least compare the results of both methods before we decide. The 3D tube could be an example to test for the 3D case.
> > 
> > 
> > If we decide to change things for the 3D case as well, yes. Otherwise the perpendicular flap should be fine (which I cheched at least visually).
> 
> But with this PR we are adding the cellDisplacement -> pointDisplacement step in both the 2D and 3D, right?
> 
> I would favor to just keep the 3D as it is, which (as I understood) is also your opinion. But I am not sure I understand what is happening in the 3D case then.

Yes, I haven't adjusted the code to differ between both dimensions because I wanted to wait for your opinion.

MakisH:
Then, since we share the same opinion on:
- 2D: cellDisplacement
- 3D: pointDisplacement

from my side feel free to shape the rest of the code. :-)

DavidSCN:
So, the first clean version would be ready for a review. Now, every dimension works with centers and nodes. Maybe I will exchange the `primitivePatchPointInterpolator` by a `motionInterpolator`, but this will not change the overall structure and approach. 
It would be nice if someone could test the DisplacementDelta class, I don't have any solid solver for that at the moment.

I'm not so happy with the massive code duplication at the moment. Maybe I squash the DisplacementDelta and Displacement class together in a base class and define very small derived classes. However, this would go into another PR.

DavidSCN:
Ready to merge. I'll stick to the current methods. There might be other interpolation methods in OpenFOAM but they mostly apply the whole fields rather than single boundary fields.

FSI/Displacement.C
```diff
@@ -2,22 +2,35 @@
...
-)
+preciceAdapter::FSI::Displacement::Displacement(
+    const Foam::fvMesh &mesh,
+    const std::string   namePointDisplacement)
+    : pointDisplacement_(
+          const_cast<pointVectorField *>(
+              &mesh.lookupObject<pointVectorField>(namePointDisplacement))),
+      cellDisplacement_(
+          const_cast<volVectorField *>(
+              &mesh.lookupObject<volVectorField>("cellDisplacement"))),
```
MakisH, line 13: For consistency, we should also get the name of `cellDisplacement` from the configuration file. Even though I guess we could eventually drop this feature, I am not sure if anyone is using it.

FSI/Displacement.C
```diff
@@ -2,22 +2,35 @@
...
+      mesh_(mesh)
 {
     dataType_ = vector;
 }
 
+// We cannot do this steps in the constructor by design of the adapter since the information of the CouplingDataUser is
```
MakisH, line 19:
```suggestion
// We cannot do this step in the constructor by design of the adapter since the information of the CouplingDataUser is
```

FSI/Displacement.C
```diff
@@ -30,34 +43,45 @@ void preciceAdapter::FSI::Displacement::write(double * buffer, bool meshConnecti
...
+
 // return the displacement to use later in the velocity?
 void preciceAdapter::FSI::Displacement::read(double * buffer, const unsigned int dim)
 {
-    // For every element in the buffer
-    int bufferIndex = 0;
-
-    // For every boundary patch of the interface
-    for (uint j = 0; j < patchIDs_.size(); j++)
+    for (unsigned int j = 0; j < patchIDs_.size(); j++)
```
MakisH, line 50: we could also use `std::size_t` or `auto` here. But let's be consistent with the rest of the codebase for now.

FSI/Displacement.C
```diff
@@ -30,34 +43,45 @@ void preciceAdapter::FSI::Displacement::write(double * buffer, bool meshConnecti
...
+
+        } else if (this->locationsType_ == "faceNodes") {
+
+          // Get the displacement on the patch
+          fixedValuePointPatchVectorField &pointDisplacementFluidPatch(
+              refCast<fixedValuePointPatchVectorField>(
+                  pointDisplacement_->boundaryFieldRef()[patchID]));
+
+          // Overwrite the nodes on the interface directly
+          for (int i = 0; i < pointDisplacement_->boundaryFieldRef()[patchID].size(); ++i) {
```
MakisH, line 81: Why not using the same `forAll` construct here?

FSI/DisplacementDelta.C
```diff
@@ -2,22 +2,35 @@
...
+              &mesh.lookupObject<pointVectorField>(namePointDisplacement))),
+      cellDisplacement_(
+          const_cast<volVectorField *>(
+              &mesh.lookupObject<volVectorField>("cellDisplacement"))),
+      mesh_(mesh)
 {
     dataType_ = vector;
 }
 
+// We cannot do this steps in the constructor by design of the adapter since the information of the CouplingDataUser is
```
MakisH, line 19:
```suggestion
// We cannot do this step in the constructor by design of the adapter since the information of the CouplingDataUser is
```

MakisH (approved):
The code looks clean (see some small comments), I have not tried to run it. Please check if we get the same results for the `pointDisplacement` cases. The system tests should also catch any differences here, but anyway.

DavidSCN:
> The code looks clean (see some small comments), I have not tried to run it. Please check if we get the same results for the `pointDisplacement` cases. The system tests should also catch any differences here, but anyway.

I checked the pointDisplacement. As I said, it would be good if anyone can check the DisplacementDelta case,. I don't think that we have a systemtest for this data class.

FSI/Displacement.C
DavidSCN: Which configuration file? The preciceDict? Or should I just store the name in the `FSI.H`? Note that we actually read an already present `volVectorField` called `cellDisplacement`.

FSI/Displacement.C
DavidSCN: I think the range based solution is more recent, also in OpenFOAM itself. At this specific point, we cannot use it since we need the actual position of the iterator.

FSI/Displacement.C
DavidSCN: I don't like it that much. I replaced it now by `forAll`

FSI/Displacement.C
MakisH: The `preciceDict`. Instead of `"cellDisplacement"` we should have something like `nameCellDisplacement`, as in three lines above.

DavidSCN: Done

DavidSCN:
I don't see any systemtests passing for this PR. However, I just checked the `DisplacementDelta` case and it works (as expected). Finally ready to merge.  The plot looks a bit unusual since DisplacementDelta refers more to the velocity.

PR #148, opened 2020-12-16
Refactor Force based data sets

### LINKED ISSUE #139: Refactor FSI Stress/Force
As observed in #133 and discussed in #134 , the current design and implementation of writing forces and stresses is rather bad: The stress field inherits common functionality from the force field. Though the Force field creates its own data field, which is not required in case stress data is used. 

A temporary solution is given in #134. A much simpler and cleaner solution would be to create a force calculation base class for both fields, so that common functionality can be inherited independently in stress and force classes respectively.

DavidSCN:
I took the opportunity to clean-up the Force based data classes and split the common functionality in a base class. The PR reverts a crime introduced in #134 and partially #125, reduces code duplication and makes the whole data set much easier to maintain. Closes #139.

I hope we can run the tests on this PR.

### DIFF
--- a/FSI/Stress.C
+++ b/FSI/Stress.C
@@ -42,7 +42,8 @@ void preciceAdapter::FSI::Stress::read(double * buffer, const unsigned int dim)
 
 vectorField preciceAdapter::FSI::Stress::getFaceVectors(const unsigned int patchID) const
 {
-  return mesh_.boundary()[patchID].nf();
+    // face normal vectors
+    return mesh_.boundary()[patchID].nf();
 }
 
 preciceAdapter::FSI::Stress::~Stress()
--- a/FSI/Stress.C
+++ b/FSI/Stress.C
@@ -5,34 +5,17 @@ using namespace Foam;
 preciceAdapter::FSI::Stress::Stress
 (
     const Foam::fvMesh& mesh,
-    const fileName& timeName,
     const std::string solverType
-    /* TODO: We should add any required field names here.
-    /  They would need to be vector fields.
-    /  See FSI/Temperature.C for details.
-    */
 )
 :
-Force(mesh,solverType),
-mesh_(mesh),
-solverType_(solverType)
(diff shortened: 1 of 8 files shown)


FSI/Stress.C
```diff
@@ -49,96 +32,17 @@ solverType_(solverType)
...
-    */
-    FatalErrorInFunction
-        << "Reading stresses is not supported."
-        << exit(FatalError);
+    this->readFromBuffer(buffer);
+}
+
+vectorField preciceAdapter::FSI::Stress::getFaceVectors(const unsigned int patchID) const
+{
+  return mesh_.boundary()[patchID].nf();
```
MakisH, line 45: Maybe add a comment here on what `nf()` does, similar to the one for `Sf()`?

MakisH (approved):
Looking at the code, I like this refactoring. :+1:

- Before we had `Force` and `Stress : public Force`.
- Now we have a `ForceBase` with most functionality, and `Force` and `Stress` essentially changing the normal vectors they are using, overriding the following virtual function:

```c++
Foam::vectorField getFaceVectors(const unsigned int patchID) const override;
```
In the case of Force:
```c++
  // Normal vectors multiplied by face area
  return mesh_.boundary()[patchID].Sf();
  ```
In the case of Stress:
```c++
return mesh_.boundary()[patchID].nf();
```

This PR removes more code than adds, so this is highly encouraging! I have not tried to build or test the code, but if you already checked 1-2 cases, it should be fine.

DavidSCN:
> This PR removes more code than adds, so this is highly encouraging! I have not tried to build or test the code, but if you already checked 1-2 cases, it should be fine.

Yes. Considering that there is now more 'c++ structure' code, it removes even more code tom maintain than shown in the overview. Let me address your comment and check the result again.

FSI/Stress.C
DavidSCN: Done.

PR #155, opened 2021-03-26
Add Assertion for xy-plane for 2D cases

### LINKED ISSUE #154: Assert the correct layer of 2D computations
https://github.com/precice/openfoam-adapter/blob/8355e96ff1c9721d45c42f8d79c8e9da22425d76/Interface.C#L94-L105

We implicitly assume here that the out-of-plane layer is the z-axis and the interface is located in xy-plane. We currently don't do anything if the user violate it, but the result will be nonsense. In other places, we perform a loop over `dim` resulting in a similar problem. We need to prevent this case. Also related to precice/calculix-adapter#56.

DavidSCN:
Closes #154

### DIFF
--- a/Interface.C
+++ b/Interface.C
@@ -93,12 +93,41 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
 
             // Assign the (x,y,z) locations to the vertices
             for (int i = 0; i < faceCenters.size(); i++)
-            {
-                vertices[verticesIndex++] = faceCenters[i].x();
-                vertices[verticesIndex++] = faceCenters[i].y();
-                if(dim_ == 3)//out-of-plane axis = z
-                    vertices[verticesIndex++] = faceCenters[i].z();
+              for (unsigned int d = 0; d < dim_; ++d)
+                vertices[verticesIndex++] = faceCenters[i][d];
+
+#ifdef ADAPTER_DEBUG_MODE
+            // Check, if we are in the right layer in case of preCICE dimension 2
+            if (dim_ == 2) {
+              const pointField faceNodes =
+                  mesh.boundaryMesh()[patchIDs_.at(j)].localPoints();
+              //Allocate memory for z-coordinates
+              std::array<double, 2>  z_location({0, 0});
+              constexpr unsigned int z_axis = 2;
+
+              // Find out about the existing planes
+              z_location[0] = faceNodes[0][z_axis];
+              for (int i = 0; i < faceNodes.size(); i++) {
+                if (z_location[0] == faceNodes[i][z_axis])
+                  continue;
+                else {
+                  z_location[1] = faceNodes[i][z_axis];
+                  break;
+                }
+              }
+
+              // Check all remaining nodes
+              for (int i = 0; i < faceNodes.size(); i++)
+                if (z_location[0] == faceNodes[i][z_axis] || z_location[1] == faceNodes[i][z_axis])
+                  continue;
+                else
+                  FatalErrorInFunction
+                      << "It seems like you are using preCICE in 2D and your geometry is not located int the xy-plane. "
+                         "The OpenFOAM-adapter implementation supports preCICE 2D cases only with the z-axis as out-of-plane direction."
+                         "Please rotate your geometry so that the geometry is located in the xy-plane."
+                      << exit(FatalError);
             }
+#endif
         }
 
         // Pass the mesh vertices information to preCICE
@@ -140,12 +169,8 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
             // Assign the (x,y,z) locations to the vertices
             // TODO: Ensure consistent order when writing/reading
             for (int i = 0; i < faceNodes.size(); i++)
-            {
-                vertices[verticesIndex++] = faceNodes[i].x();
-                vertices[verticesIndex++] = faceNodes[i].y();
-                if (dim_ == 3)//out-of-plane axis = z!
-                    vertices[verticesIndex++] = faceNodes[i].z();
-            }
+                for (unsigned int d = 0; d < dim_; ++d)
+                  vertices[verticesIndex++] = faceNodes[i][d];
         }
 
         // Pass the mesh vertices information to preCICE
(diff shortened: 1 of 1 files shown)


Interface.C
```diff
@@ -93,12 +93,41 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
             for (int i = 0; i < faceCenters.size(); i++)
-            {
-                vertices[verticesIndex++] = faceCenters[i].x();
-                vertices[verticesIndex++] = faceCenters[i].y();
-                if(dim_ == 3)//out-of-plane axis = z
-                    vertices[verticesIndex++] = faceCenters[i].z();
+              for (unsigned int d = 0; d < dim_; ++d)
+                vertices[verticesIndex++] = faceCenters[i][d];
+
+#ifdef ADAPTER_DEBUG_MODE
```
DavidSCN, line 99: @MakisH Not sure if this corresponds only to messages or general DEBUG as well.

MakisH: `ADAPTER_DEBUG_MODE` is set in https://github.com/precice/openfoam-adapter/blob/8ae5f061ab26ae059b8cdcf304089306f9ce7227/Allwmake#L20 and yes, it also refers to general debug mode as well.

Interface.C
```diff
@@ -93,12 +93,41 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+              // Find out about the existing planes
+              z_location[0] = faceNodes[0][z_axis];
+              for (int i = 0; i < faceNodes.size(); i++) {
+                if (z_location[0] == faceNodes[i][z_axis])
+                  continue;
+                else {
+                  z_location[1] = faceNodes[i][z_axis];
+                  break;
+                }
+              }
```
MakisH, line 117: Could you please add a comment explaining the way of thinking here? With a quick look, I could not follow.

Interface.C
```diff
@@ -93,12 +93,41 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
-            {
-                vertices[verticesIndex++] = faceCenters[i].x();
-                vertices[verticesIndex++] = faceCenters[i].y();
-                if(dim_ == 3)//out-of-plane axis = z
-                    vertices[verticesIndex++] = faceCenters[i].z();
+              for (unsigned int d = 0; d < dim_; ++d)
+                vertices[verticesIndex++] = faceCenters[i][d];
+
+#ifdef ADAPTER_DEBUG_MODE
+            // Check, if we are in the right layer in case of preCICE dimension 2
```
MakisH, line 100:
```suggestion
            // Check if we are in the right layer in case of preCICE dimension 2
```

Interface.C
```diff
@@ -93,12 +93,41 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+              }
+
+              // Check all remaining nodes
+              for (int i = 0; i < faceNodes.size(); i++)
+                if (z_location[0] == faceNodes[i][z_axis] || z_location[1] == faceNodes[i][z_axis])
+                  continue;
+                else
+                  FatalErrorInFunction
+                      << "It seems like you are using preCICE in 2D and your geometry is not located int the xy-plane. "
+                         "The OpenFOAM-adapter implementation supports preCICE 2D cases only with the z-axis as out-of-plane direction."
```
MakisH, line 126:
```suggestion
                         "The OpenFOAM adapter implementation supports preCICE 2D cases only with the z-axis as out-of-plane direction."
```

Interface.C
```diff
@@ -140,12 +169,8 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
             // TODO: Ensure consistent order when writing/reading
             for (int i = 0; i < faceNodes.size(); i++)
-            {
-                vertices[verticesIndex++] = faceNodes[i].x();
-                vertices[verticesIndex++] = faceNodes[i].y();
-                if (dim_ == 3)//out-of-plane axis = z!
-                    vertices[verticesIndex++] = faceNodes[i].z();
-            }
+                for (unsigned int d = 0; d < dim_; ++d)
+                  vertices[verticesIndex++] = faceNodes[i][d];
```
MakisH, line 173: Two questions:
1. Is the `[i][d]` form safe? Is it used elsewhere in OpenFOAM? I am mostly worried about accidentally changing some type in the future.
2. Since we restrict anyway the user to XY, what is the motivation for changing this?

DavidSCN: I have not made any safety check. Regarding point 2: We compress 6 locs to 2 locs, it has less duplication and is easier to read.

Interface.C
DavidSCN: Done.

Interface.C
MakisH: I see the "less duplication" point, I am not so sure about the "easier to read", as in the current code we have explicit calls to `.x()`, `.y()`, `.z()`.

I guess it should be safe enough, let's use it. But then I guess we use `.x()` etc also in other parts of the adapter.

Interface.C
```diff
@@ -93,12 +93,44 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
-            {
-                vertices[verticesIndex++] = faceCenters[i].x();
-                vertices[verticesIndex++] = faceCenters[i].y();
-                if(dim_ == 3)//out-of-plane axis = z
-                    vertices[verticesIndex++] = faceCenters[i].z();
+              for (unsigned int d = 0; d < dim_; ++d)
+                vertices[verticesIndex++] = faceCenters[i][d];
+
+#ifdef ADAPTER_DEBUG_MODE
+            // Check if we are in the right layer in case of preCICE dimension 2
```
MakisH, line 100:
```suggestion
            // Check if we are in the right layer in case of preCICE dimension 2
            // If there is at least one node with a different z-coordinate, then the (2D) geometry is not on the xy-plane, as required.
```

MakisH (approved):

Interface.C
DavidSCN: It's anyway mixed, at the moment, as I use the same code already for the FSI module :sweat_smile:

Issue #137, opened 2020-08-13
empty pkg-config Cflags  despite correct libprecice.pc file


mojtabaAmiraslanpour:
While compiling the OpenFOAM 7.0 adapter on PreCICE 2.1 (installed by using binary DEB packages on Ubuntu 18.04.1), header files are not found. I found that `pkg-config --cflags libprecice` returns empty.

I used the procedure mentioned in [here](https://unix.stackexchange.com/questions/362289/why-is-pkg-config-cflags-openssl-returning-nothing-on-rhel-6-8) to change the line which returns the path to the header files in the Allwmake file:
```
export ADAPTER_PKG_CONFIG_CFLAGS="`PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 pkg-config --cflags libprecice`"
```
And this fixed the issue.

MakisH:
Thank you for reporting this, @mojtabaAmiraslanpour!

It is true that pkgconfig ignores the paths known to the system, but this shouldn't be a problem, exactly because the paths are already known.

I am currently a bit confused on why you got this problem and why this change fixed it. I am not sure if this is related, but did you also start a new terminal in the meantime?

I will try to reproduce your issue, please also let me know in case you have more details that could help.

mojtabaAmiraslanpour:
my precice header files are located inside `/usr/include/precice` folder. Is that the normal installation?
While executing the Allwmake file before the edit, I got `No such file or directory` errors for the header files.

MakisH:
> my precice header files are located inside `/usr/include/precice` folder. Is that the normal installation?

Yes, this is the normal installation, and adapters should use preCICE as e.g.:
https://github.com/precice/openfoam-adapter/blob/c49b862aab4845ae79b8b7257a20e15ba5a0019c/Adapter.H#L22

This means that looking into `/usr/include` (a standard path) should be enough.

I will look into this closer.

MakisH:
@mojtabaAmiraslanpour did you have any news or any other details regarding this that would help me to reproduce it?

MakisH:
I believe this has probably been fixed with https://github.com/precice/openfoam-adapter/pull/157

Issue #119, opened 2020-02-18
Linking fails despite correct pkg-config flags


fsimonis:
# Problem Description

When installing precice 2.0.0 in `/usr/local/` and using ldd as a linker., then running `./Allwmake` fails due to missing symbols
Setting the `LD_LIBRARY_PATH` fixes this issue.

# Potential Solution

Forward the necessary flags to the linker invocation.

quentonwu:
To elaborate, run the following command:"export LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu:$LD_LIBRARY_PATH" in the same terminal window before running ./Allwmake should allow a successful build

fsimonis:
@quentonwu Thank you for the comment! The (your?) blog post mentions preCICE version `1.3.0`. Which version are you currently using?

This issue should be fixed with https://github.com/precice/precice/pull/686 which was released with preCICE `2.1.0`. @MakisH Can you confirm this?

quentonwu:
@fsimonis No, the blog post isn't mine, I just came across it after searching the error online. I am using the latest preCICE build(2.1.1) on Ubuntu along with OpenFOAM 2012

MakisH:
> This issue should be fixed with [precice/precice#686](https://github.com/precice/precice/pull/686) which was released with preCICE `2.1.0`. @MakisH Can you confirm this?

It is not very easy for me to test this at the moment, but I understand from @quentonwu's comment that it is not fixed.

quentonwu:
I just tried to build the CalculiX adapter and I had the exact same problem as well. Could it be something from my end? Or have the library changed something? Apologises for my uncertainness, I don't have much experience building software.

MakisH:
@fsimonis @quentonwu do you still have the same issue with v1.1.0?

fsimonis:
We could use build the vm with precice installed in `/usr/local` (default).

MakisH:
> We could use build the vm with precice installed in `/usr/local` (default).

Please open an issue in the vm for that, explaining the motivation.

davidscn:
I stumbled across the same issue and dug a bit deeper. I think the problem here is more with our building script. We use the following lines to 'detect' success or failure of the linking step in our script:

https://github.com/precice/openfoam-adapter/blob/25f9798da00eb9ef38d96109fb2dab3eaec310a2/Allwmake#L109-L118

[...]

PR #157, opened 2021-03-28
More robust building scripts

MakisH:
I checked our `Allclean` and `Allwmake` scripts for common issues with shellcheck and found a few good points (these scripts are quite old). I may adjust a few more parts before merging this.

I also updated the output to point to the new documentation and to remove the "try the `tutorials/` next".

More changes:
- Remove hints for using preCICE as a static library. It should anyway not work / we don't want the user to go in this direction.
- Remove mentions and uses of `PRECICE_ROOT`. This was a relic from the early preCICE 1.x days, which we had there for backwards compatibility.
- Remove mentions and uses of `CPLUS_INCLUDE_PATH`, `LD_LIBRARY_PATH`, and `LIBRARY_PATH`. Nowadays, no user should need to use these and they also don't really work as intended.
- Add inline comments next to deprecated OpenFOAM methods, explaining why we use them.
- Always show the WMake output. It is short, expected, and shows the progress. We still keep the `wmake.log` and `ldd.log`.
- OpenFOAM was complaining that some files were not ending with a newline. This fixes it.


MakisH:
@DavidSCN this is ready from my side. Any comments?

Allwmake
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-cd ${0%/*} || exit 1    # Run from this directory
+set -e -u
```
DavidSCN, line 2: `set -e` won't work since it directly exits. For me it results in 
`./Allwmake: 51: CPLUS_INCLUDE_PATH: parameter not set`

Allwmake
```diff
@@ -123,20 +125,19 @@ log "  ADAPTER_GLOBAL_LIBRARY_PATHS    = ${ADAPTER_GLOBAL_LIBRARY_PATHS}"
 
 # Run wmake (build the adapter) and check the exit code
 log ""
-log "Building with WMake (see the wmake.log log file)..."
-wmake ${ADAPTER_WMAKE_OPTIONS} libso > wmake.log 2>&1
-log ""
-if [ $? -ne 0 ] || [ "$(grep -c -E "error:" wmake.log)" -ne 0 ]; then
+log "Building with WMake (see the wmake.log log file)...\n"
+if ! wmake "${ADAPTER_WMAKE_OPTIONS}" libso >wmake.log 2>&1 ||
```
DavidSCN, line 129: Now I have finally the chance to comment on this: I don't like the hidden compiler output and comparing it to other libraries, e.g. swak4foam, it seems to be non-default hiding it. Maybe it is also a developer problem, but could we please pipe the compiler output (additionally) to `cout` by default. If something goes wrong, one would directly see.

Allwmake
MakisH: I added some placeholders so that this error is not triggered at expected places. By the way, this is because of `-u`, not because of `-e`.

Allwmake
MakisH: I changed this to
```shell
if ! wmake "${ADAPTER_WMAKE_OPTIONS}" libso 2>&1 | tee wmake.log ||
    [ "$(grep -c -E "error:" wmake.log)" -ne 0 ]; then
```
This way, we get both the verbose output and the `wmake.log` file.

Since this throws a few known deprecation warnings to the user, I added some inline comments explaining that these are intentionally there.

MakisH:
I also removed:
- The `PRECICE_ROOT` variable (no user should need to use this anymore, ever)
- The notes about using preCICE as a static library (we don't really want the user to do that anyway). I need to update this in the docs afterwards.

Ready for another review.

DavidSCN (commented):
Still doesn't work for me, since I have not set `CPLUS_INCLUDE_PATH`. You could just remove the `set -u` if you like.
```
./Allwmake: 99: ADAPTER_GLOBAL_CPLUS_INC_PATHS: parameter not set
```
The problem here is also that I don't see the missing variable in the terminal since `Allwmake` immediately aborts without any hint. My first thought was that the building was executed in the background.

>  The PRECICE_ROOT variable (no user should need to use this anymore, ever)

I used it and building doesn't work for me now (FYI). My `include-path` in the `PKG_CONFIG` is just wrong. In order to set my `CPLUS_INCLUDE_PATH` I usually would  export my `CPATH`, since it sets the `C_INCLUDE_PATH` as well. This works only if I remove the `set -u -e` since `C_PLUS_INCLUDE_PATH` is still undefined.

Allwmake
```diff
@@ -48,38 +35,44 @@ log "Building the OpenFOAM adapter for preCICE..."
 # 5. translate the new line characters to "\", i.e. make everything again one line.
 
 # Extract the paths from the CPLUS_INCLUDE_PATH list and convert them to a sequence of "-I/path/ \"
-if [ "${CPLUS_INCLUDE_PATH}" ]; then
-    ADAPTER_GLOBAL_CPLUS_INC_PATHS=$(echo $CPLUS_INCLUDE_PATH | tr ':' '\n' |  sed '/^$/d' | sed 's/^/-I/' | tr '\n' ' \\')
+if [ "${CPLUS_INCLUDE_PATH:-}" ]; then
+    ADAPTER_GLOBAL_CPLUS_INC_PATHS=$(echo "${CPLUS_INCLUDE_PATH}" | tr ':' '\n' | sed '/^$/d' | sed 's/^/-I/' | tr '\n' ' \\ ')
+else
```
DavidSCN, line 40:
```suggestion
else if
elif [ "${CPATH:-}" ]; then
    ADAPTER_GLOBAL_CPLUS_INC_PATHS=$(echo "${CPATH}" | tr ':' '\n' | sed '/^$/d' | sed 's/^/-I/' | tr '\n' ' \\ ')
else
```

MakisH:
> Still doesn't work for me, since I have not set `CPLUS_INCLUDE_PATH`. You could just remove the `set -u` if you like.

The purpose of these options is to scream whenever there is a problem like this, so that we don't end up in undefined behavior situations. So, in my perspective, `-u` helped us here already!

> ```
> ./Allwmake: 99: ADAPTER_GLOBAL_CPLUS_INC_PATHS: parameter not set
> ```
> 
> The problem here is also that I don't see the missing variable in the terminal since `Allwmake` immediately aborts without any hint. My first thought was that the building was executed in the background.

I fixed this as well right now. Please check again.

> > The PRECICE_ROOT variable (no user should need to use this anymore, ever)
> 
> I used it and building doesn't work for me now (FYI). My `include-path` in the `PKG_CONFIG` is just wrong. In order to set my `CPLUS_INCLUDE_PATH` I usually would export my `CPATH`, since it sets the `C_INCLUDE_PATH` as well. This works only if I remove the `set -u -e` since `C_PLUS_INCLUDE_PATH` is still undefined.

- Why did you need to use `PRECICE_ROOT`? Do we still have this in any documentation page? I know that developers may still use it sometimes in some places, but I believe we should not have the adapters support it and always ask for `pkg-config`.
- To clarify: the OpenFOAM `etc/bashrc` (one of the discovery methods) sets the `LD_LIBRARY_PATH`, so we use it. The `CPLUS_INCLUDE_PATH` and `LIBRARY_PATH` I still keep in case they are used for preCICE, but maybe we should even remove those.
- What happens if you [set the `PKG_CONFIG_PATH`](https://www.precice.org/installation-linking.html#make-and-scripts-in-general)?

I am trying to test this as well, but I need to fix something on my system first.

Allwmake
MakisH: Is there a good practice reason behind this suggestion? Any citations? See my comment below, I would rather remove this altogether, if possible.

DavidSCN: `CPATH` is for `C` and `c++`

https://stackoverflow.com/questions/558803/how-to-add-a-default-include-path-for-gcc-in-linux

DavidSCN:
> I fixed this as well right now. Please check again.

Works!

>   Why did you need to use PRECICE_ROOT? Do we still have this in any documentation page? I know that developers may still use it sometimes in some places, but I believe we should not have the adapters support it and always ask for pkg-config.
    
I guess this is a legacy crime of myself. It was required in the beginning of the adapter and I never changed it. 

> To clarify: the OpenFOAM etc/bashrc (one of the discovery methods) sets the LD_LIBRARY_PATH, so we use it. The CPLUS_INCLUDE_PATH and LIBRARY_PATH I still keep in case they are used for preCICE, but maybe we should even remove those.
What happens if you set the PKG_CONFIG_PATH?

I guess the PKG_CONFIG will work, but I don't install my system preCICE in order to keep track of the actual preCICE version in the adapters.

MakisH:
> I guess the PKG_CONFIG will work, but I don't install my system preCICE in order to keep track of the actual preCICE version in the adapters.

You don't need to install. It should work by adding the local path. For me:
```
export PKG_CONFIG_PATH="/home/makish/github/precice/build/lib/pkgconfig/:$PKG_CONFIG_PATH"
export precice_DIR="/home/makish/github/precice/build/"
```

But beware there may be issues at the moment, so trying it now would be helpful (see #119 #137). We can also enforce this later.

DavidSCN:
I have essentially the same exports in my `bashrc`. But this doesn't work and it is not surprising. It only works if you adjust your install prefix accordingly. My `.pc` file gives
```
includedir=/usr/local/include
libdir=/usr/local/lib

Name: preCICE
Description: preCICE coupling library
Version: 2.2.0
URL: https://www.precice.org
Libs: -L${libdir} -lprecice
Cflags: -I${includedir}
```
which is also the preCICE default.

MakisH:
Indeed, this is also how I (don't) use it at the moment. What if you set the [prefix](https://www.precice.org/installation-source-configuration.html)?

DavidSCN:
This works :+1:

Would still be nice to have a 'manual' variant available.

MakisH:
> This works +1
 
Apart from setting the prefix, how did you fix the `include/` path? If I set `precice/build` as prefix, my `precice.pc` becomes:
```
includedir=/home/makish/github/precice/build/include
libdir=/home/makish/github/precice/build/lib
```
so I am looking for a nice way to automatically generate the correct thing (without having to install).

> Would still be nice to have a 'manual' variant available.

Then I keep the `CPLUS_INCLUDE_PATH` and `LIBRARY_PATH`.

DavidSCN:
> > This works +1
> 
> Apart from setting the prefix, how did you fix the `include/` path? If I set `precice/build` as prefix, my `precice.pc` becomes:
> 
> ```
> includedir=/home/makish/github/precice/build/include
> libdir=/home/makish/github/precice/build/lib
> ```
> 
> so I am looking for a nice way to automatically generate the correct thing (without having to install).
>

I need to admit I have simply edited the `.pc` file.
```
includedir=$HOME/precice/src
libdir=$HOME/precice/build
```
but I guess this is not a persistent solution as cmake will probably override the file if I rebuild the library.

> > Would still be nice to have a 'manual' variant available.
> 
> Then I keep the `CPLUS_INCLUDE_PATH` and `LIBRARY_PATH`.

Alright. But still don't want to add the `CPATH` right? Just two last comments: do you want to point at the end to the adapter specific variables, e.g. `ADAPTER_PKG_CONFIG_LIBS` or to the more OS like variables (as you want to keep them anyway). What happens if somebody uses it on mac (not sure if OF supports it at all)?

MakisH:
> ```
> includedir=$HOME/precice/src
> libdir=$HOME/precice/build
> ```
> but I guess this is not a persistent solution as cmake will probably override the file if I rebuild the library.
For me `wmake` works like this, but `ldd` fails (so the library is not discoverable at runtime). Does it work for you (without setting preCICE in the `LD_LIBRARY_PATH`?

> Alright. But still don't want to add the `CPATH` right?

I am not yet convinced that there is a good reason to change this. The `CPLUS_INCLUDE_PATH` clearly indicates the purpose, at least.

> Just two last comments: do you want to point at the end to the adapter specific variables, e.g. `ADAPTER_PKG_CONFIG_LIBS` or to the more OS like variables (as you want to keep them anyway).

Yes, because they are not in the same format. I first extract everything I need from the system and forward it. Does this not make sense somewhere?

> What happens if somebody uses it on mac (not sure if OF supports it at all)? 

I do not have the means to check this. In principle they technically support macOS, I think. Definitely the environment variables are different there, though, but this PR should not break anything in this direction.

DavidSCN:
Ok, then please go ahead.

Allwmake
MakisH: We have an issue here: apparently, `CPLUS_INCLUDE_PATH` does not have any effect, but `CPATH` does. I am investigating this.

DavidSCN: I thought you want to skip the `CPATH`?

DavidSCN:
I have now double-checked everything. What were we doing all the time? We can now still export the `LIBRARY_PATH` and the `CPATH`/`CPLUS_INCLUDE_PATH` and it works like a charm.

MakisH:
Yes, these should go to the compiler anyway. I don't remember anymore, but I guess I made things complicated because of SCons and the installation practices we were using back then.

DavidSCN (approved):
Awesome :tada:

MakisH (commented):
Everything also looks as intended from my side.

Issue #89, opened 2019-07-02
Reduce debug output verbosity


MakisH:
@uekerman observed that the output of the OpenFOAM adapter is too verbose in general, reducing readability. 

We provide a preprocessor flag `ADAPTER_DEBUG_MODE` to toggle the verbosity, with the goal to improve performance. In "release" mode, we get the following kind of output:

```
Starting time loop

---[preciceAdapter] The preciceAdapter was loaded.
---[preciceAdapter] Reading the adapter's YAML configuration file /home/makish/adapter/tutorials/CHT/flow-over-plate/buoyantPimpleFoam-laplacianFoam/Fluid/precice-adapter-config.yml...
(0) 18:49:18 [impl::SolverInterfaceImpl]:101 in configure: This is preCICE version 1.5.1
(0) 18:49:18 [impl::SolverInterfaceImpl]:102 in configure: Configuring preCICE with configuration: "precice-config.xml"
(0) 18:49:18 [impl::SolverInterfaceImpl]:145 in configure: Run in coupling mode
(0) 18:49:18 [impl::SolverInterfaceImpl]:214 in initialize: Setting up master communication to coupling partner/s
(0) 18:49:18 [impl::SolverInterfaceImpl]:219 in initialize: Masters are connected
(0) 18:49:18 [partition::ProvidedPartition]:36 in communicate: Gather mesh Fluid-Mesh
(0) 18:49:18 [partition::ProvidedPartition]:62 in communicate: Send global mesh Fluid-Mesh
(0) 18:49:18 [partition::ProvidedPartition]:79 in compute: Compute partition for mesh Fluid-Mesh
(0) 18:49:18 [impl::SolverInterfaceImpl]:223 in initialize: Setting up slaves communication to coupling partner/s
(0) 18:49:18 [impl::SolverInterfaceImpl]:228 in initialize: Slaves are connected
(0) 18:49:18 [impl::SolverInterfaceImpl]:254 in initialize: it 1 of 200 | dt# 1 | t 0 of 1 | dt 0.01 | max dt 0.01 | ongoing yes | dt complete no | write-iteration-checkpoint | 
(0) 18:49:18 [cplschemes::SerialCouplingSchemes]:109 in initializeData: initializeData is skipped since no data has to be initialized
---[preciceAdapter] preCICE was configured and initialized
[...]

PR #159, opened 2021-04-11
Turn off debug messages by default

MakisH:
Until now, we always had debug messages turned on by default, which maybe was a bit too much for the user. Similarly, this makes log files unnecessarily longer/noisier and potentially everything slower.

I have the impression that for a while these additional messages have not really helped us discover new problems in user-submitted log files and they can always very easily be turned on again, if needed. I suggest that we change this default.

Now it would also be a good time since we anyway massively update the tutorials and the user will notice related changes.

This does not change the compilation mode, which is the OpenFOAM default, or as set by the user on the OpenFOAM side.

Partially related to #89.

@DavidSCN please merge if you aggree.

### DIFF
--- a/Allwmake
+++ b/Allwmake
@@ -17,7 +17,7 @@ ADAPTER_PRECICE_ROOT="${PRECICE_ROOT}"
 # ADAPTER_PRECICE_DEP="${ADAPTER_PRECICE_DEP} -lpython2.7"
 
 # Optional: Preprocessor flags ("-DADAPTER_DEBUG_MODE" enables debug messages)
-ADAPTER_PREP_FLAGS="-DADAPTER_DEBUG_MODE"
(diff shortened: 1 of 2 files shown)


DavidSCN (commented):
I completely agree that there are many messages which are just useless. In particular, the messages during the actual run are spamming. I like the messages during initialization though, but since we improved the error messages in preCICE the user is usually informed about misconfigurations so that we can drop it.
However, in #155 we introduced a check in order to prevent a misuse for the new 2D mode. We should definitely enable it by default. I wrapped it around the `DEBUG_MODE` flag, but since it is anyway an initialization cost check occurring only once  we could just enable the check at anytime.

MakisH:
Thank you, I removed the debug mode from this part of `Interface.C`, please check.

DavidSCN (approved):

PR #161, opened 2021-04-12
Move documentation to develop

MakisH:
Now that we have tested sourcing the documentation from the adapter repository, let's also move it directly to develop.

We also need to update the subproject specification in the website afterwards:
- https://github.com/precice/precice.github.io/blob/master/_config.yml
- https://github.com/precice/precice.github.io/blob/master/.gitmodules

Related to #157.


DavidSCN (approved):
We addressed the documentation review in the respective PRs, so I didn't go through the whole documentation right now. Let's merge and update the tutorials submodule.

PR #162, opened 2021-04-13
Re-enable multi-core builds

DavidSCN:
The problem occurred due to the double-quotes. Shellcheck will actually complain about this. This particular exception is described here: https://github.com/koalaman/shellcheck/wiki/SC2086#exceptions
We could find an alternative by restricting the script to `bash` etc. I would propose to just go this way.

Another comment. The environment variable as described in the wiki doesn't seem to work. Does it work for you?

### DIFF
--- a/Allwmake
+++ b/Allwmake
@@ -73,7 +73,7 @@ log "  pkg-config --libs libprecice    = ${ADAPTER_PKG_CONFIG_LIBS}"
 # Run wmake (build the adapter) and check the exit code
 log ""
 log "Building with WMake (see the wmake.log log file)...\n"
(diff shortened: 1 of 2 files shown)


DavidSCN:
@MakisH I think the environment variable is out-dated. If you look into `$WM_DIR/wmake` you can look all the options. OpenFOAM handles it indeed via the option `Allwmake -j`

MakisH:
I would prefer to keep it POSIX-compliant. Let's handle the exception of shellcheck, or at least add an explicit ignore.

DavidSCN:
I like the default :+1:

PR #168, opened 2021-04-17
Adopt the "Keep a changelog" format in CHANGELOG.md

MakisH:
This rewrites the changelog to follow the structure and basic elements of https://keepachangelog.com/en/1.0.0/

This seems to be turning into a widely-adopted format and I understand that GitHub can render it nicely in the release notes, with the version labels turning into links to compare the states of the code between this and the previous version. It is being developed for at least since 5-6 years now.

Since we have not yet established a versioning system (see #52), it is a good time to formalize such things.

Notes:
- I added an exception for the markdown linter (to be added next) for this file only, as the format defines duplicate headings.
- The badge I used is not yet official, but looks not so intrusive. I got it from [this comment](https://github.com/olivierlacan/keep-a-changelog/issues/316#issuecomment-678345877).

@DavidSCN what do you think of this format? Please don't merge yet.


DavidSCN (commented):
Not sure if this 'grouping' of topics really simplifies things, but it seems to be widely used. Since I have not so much experience with changelogs I propose to just try it out.

MakisH:
One main open point is how to avoid merge conflicts in the changelog. Let's keep this nice format, but let's also adopt the "add one file per PR" workflow. I documented this in the `CONTRIBUTING.md` and added an example in this PR.

PR #169, opened 2021-04-17
Add GitHub actions to lint markdown files and check links

MakisH:
This adds:
- the [markdownlint](https://github.com/marketplace/actions/markdown-lint) action to the "Check" workflow, to check all of our markdown files for common violations.
- the [markdown-link-check](https://github.com/marketplace/actions/markdown-link-check) action to check all the links in markdown files. Since this checks too many links on GitHub, it often replies with `429 too many requests`. As this mainly concerns issues and pull requests that are not expected to disappear, I consider these links as alive.

This also fixes the currently detected violations and replaces all relative links to precice.org with static links (to the non-www version).

Additional motivation: this could find wrongly-nested lists, wrong heading levels, ... Things that are easy to miss visually. Besides, this also checks links and I already found a bad one. Moreover, this tries to enforce a bit more general markdown rules. GitHub allows a few differences (e.g. no empty lines before lists), which means that other tools (e.g. pandoc) could have trouble with out files in the future.


MakisH:
@DavidSCN this is finally ready for review! Do you see anything strange in the diff? Do you like the new actions?

Please don't merge just yet.

DavidSCN (commented):
markdown files look cleaner :+1: 
Only comment from my side: we need some instructions how to run this checks locally. Otherwise, one gets the errors when committing (:wink: ). Is there a related ubuntu package?

MakisH:
> markdown files look cleaner +1
> Only comment from my side: we need some instructions how to run this checks locally. Otherwise, one gets the errors when committing (wink ). Is there a related ubuntu package?

Very good point, I documented this in `CONTRIBUTING.md`.

CONTRIBUTING.md
```diff
@@ -8,4 +8,30 @@ and watch out for more specific details in this file.
...
+
+If you get any permission errors when trying to install from npm, [set the npm prefix to a local directory](https://stackoverflow.com/a/54170648/2254346).
+
+After installing, you can check all files in the current directory with:
+
+```bash
+markdownlint .
+```
+
+There are also extensions for many editors that automatically check Markdown files with the same system.
```
BenjaminRodenberg, line 31: Not sure if this is the right way to go. I'm locally using `act` to execute github actions. This makes sure that I actually run locally what the github action also does. I would also recommend this approach to the user, because then we don't have to keep this documentation here synchronized with what is implemented in the github actions.

Note to avoid disappointment: For this specific `act` (of course) does not work. Trying to find out why.

MakisH: Ooops, I didn't know you intended to review. I did not know about [act](https://github.com/nektos/act), thanks!

In any case, the current way is very easy to run independently, and is works directly in the editor. But this is something we could look into and document further.

BenjaminRodenberg: > Ooops, I didn't know you intended to review. I did not know about [act](https://github.com/nektos/act), thanks!

No problem. Just a random remark from my side. `act` is not working with this action. Until this is fixed, it's not helpful and the documentation above is fine.

BenjaminRodenberg: I just found the problem. I removed the check of shell scripts in https://github.com/precice/fenics-adapter/commit/de476c36fa2fa0dff5947a25dbfc3c79eb2aba0c and now act works for me. I think I don't need this check and can therefore drop it. Alternatively: If you split the two checks, one for md and one for shell scripts, I assume the md part will work with act and you don't need to document it here.

PR #170, opened 2021-04-17
Add a GitHub action for automated synchronization

DavidSCN:
_This is just a suggestion_.

By default, we provide the adapter for the ESI versions. However, we also offer the org versions and we frequently forget to synchronize them. This action would automatically open up a PR when we push to master in order to synchronize (this is a draft, now only OF7) all other branches. It won't work right now, since I have not yet created the access token. I would first like to hear your opinion @MakisH .

### DIFF
--- a/.github/workflows/sync-version.yml
+++ b/.github/workflows/sync-version.yml
@@ -1,24 +1,18 @@
-name: Sync
 on:
   push:
     branches:
-      - master
+    - master
 
 jobs:
-  sync-branches:
+  pull-request:
     runs-on: ubuntu-latest
-    name: Syncing branches
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-      - name: Set up Node
-        uses: actions/setup-node@v1
-        with:
-          node-version: 12
-      - name: Opening pull request
-        id: pull
-        uses: tretuna/sync-branches@1.2.0
-        with:
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-          FROM_BRANCH: "master"
-          TO_BRANCH: "OpenFOAM7"
+    - uses: actions/checkout@v2
+    - name: pull-request
+      uses: repo-sync/pull-request@v2
+      with:
+        source_branch: ""                                 # If blank, default: triggered branch
+        destination_branch: "OpenFOAM7"                   
+        pr_title: "Synchronize ${{ github.ref }} with OpenFOAM7" # Title of pull request
+        pr_body: "*An automated PR*: additional branches to sync: OpenFOAM4, OpenFOAM6, OpenFAOMdev"              # Full markdown support, requires pr_title to be set
+        github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


MakisH:
This is a great suggestion! Let's just keep this open for a bit, until we settle with the versioning/releasing strategy and format the code (in my list for the next days). I wanted to cleanup and synchronize these branches first.

.github/workflows/sync-version.yml
```diff
@@ -0,0 +1,18 @@
+on:
+  push:
+    branches:
+    - master
+
+jobs:
+  pull-request:
```
MakisH, line 7: I liked your previous suggestion `sync-branches`, it was less generic. I think we will easily forget what `pull-request` means.

.github/workflows/sync-version.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: pull-request
+      uses: repo-sync/pull-request@v2
+      with:
+        source_branch: ""                                 # If blank, default: triggered branch
+        destination_branch: "OpenFOAM7"                   
+        pr_title: "Synchronize ${{ github.ref }} with OpenFOAM7" # Title of pull request
+        pr_body: "*An automated PR*: additional branches to sync: OpenFOAM4, OpenFOAM6, OpenFAOMdev"              # Full markdown support, requires pr_title to be set
```
MakisH, line 17: This comment will quickly become outdated. Let's make it more generic or remove it once we implement the other jobs as well.

Actually, why not already implement everything else? It is a short copy-paste anyway. :D

.github/workflows/sync-version.yml
```diff
@@ -0,0 +1,18 @@
...
+        github_token: ${{ secrets.GITHUB_TOKEN }}
```
MakisH, line 18: I am not sure how tokens work. Would this be your token? We also have @precice-bot that we can use for this.

DavidSCN: This is a special token: https://docs.github.com/en/actions/reference/authentication-in-a-workflow (I also learned it today). I guess it's more convenient to stick to the github token.

.github/workflows/sync-version.yml
DavidSCN: PRs are also about documentation and history. I am just a bit concerned that the history will be swamped by bot PRs.

MakisH: Then, we could convert this to a single issue or comment to an existing issue to update all branches. That would be a reminder and we would do a direct merge.

DavidSCN: Why not open up a PR as a working place? Then you have directly everything at hand..

MakisH: I am just proposing this as a fix to

> I am just a bit concerned that the history will be swamped by bot PRs.

DavidSCN:
After discussing with @MakisH we decided that this feature is not worth the effort.

PR #171, opened 2021-04-21
Split checks and recommend act for local checks.

BenjaminRodenberg:
Related to discussion in https://github.com/precice/openfoam-adapter/pull/169#discussion_r617529846

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,28 +10,17 @@ Instead of directly editing `CHANGELOG.md`, please add a file `123.md`
 in `changelog-entries`, where `123` your pull request number. This helps reduce
 merge conflicts and we will merge these files at the time we release a new version.
 
-## Checking the Markdown files
-
-Every contribution is automatically checked for issues in the `.md` files.
-You can run these tests locally by installing [markdownlint-cli](https://github.com/igorshubovych/markdownlint-cli)
-from npm:
-
-```bash
-npm install -g markdownlint-cli
-```
-
-If you get any permission errors when trying to install from npm, [set the npm prefix to a local directory](https://stackoverflow.com/a/54170648/2254346).
-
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -10,28 +10,17 @@ Instead of directly editing `CHANGELOG.md`, please add a file `123.md`
...
+Every contribution is automatically checked for issues in the `.md` files.
+You can run these tests locally by installing [act](https://github.com/nektos/act).
 
-After installing, you can check all files in the current directory with:
+After installing, you can run our CI locally. Go to the root folder of this repository and run:
 
 ```bash
-markdownlint .
+act -j check_md
 ```
```
MakisH, line 26: This would be a nice contribution in the contribution guidelines of preCICE: https://precice.org/community-contribute-to-precice.html

Since act seems to be a bit complicated to install (it also needs Docker, which is not a trivial dependency), I would like to keep the previous version for the markdown-specific checks and maybe link to some general note about act at the end. Structrure-wise, act is not about Markdwon and we have more actions here.

MakisH (changes requested):
This looks reasonable, especially since the jobs can be parallelized. I started with a single job and was waiting for the point that the need for a split becomes clear. Looks like you found that point!

CONTRIBUTING.md
MakisH: I can take care of applying this change.

MakisH: Moved to the [preCICE website](https://precice.org/community-contribute-to-precice.html#optional-help-us-with-some-checks).

MakisH: I removed both sections about markdown files and links. The links part does not make sense to remain here anyway. We don't use any relative URLs anywhere here anymore.

MakisH (approved):

PR #175, opened 2021-04-29
Add curly braces to one-line if blocks

### LINKED ISSUE #174: Dangerous one-line if statements
The reformatting proposed in #173 exposes potential for future bugs: it leads to statements such as

```c++
if (conditions)
    do_stuff();
```

This is a maintainability time bomb: if one adds an additional line, such as
```c++
if (conditions)
    do_stuff();
    do_more_stuff();
```
then this will not work as intended. To prevent this, let's wrap all such blocks with `{}`.

MakisH:
Closes #174.

@DavidSCN please have a look at the diff and merge if it looks reasonable.

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -56,11 +56,19 @@ bool preciceAdapter::Adapter::configFileRead()
 
             // Set the modules switches
             if (module == "CHT")
+            {
                 CHTenabled_ = true;
+            }
+
             if (module == "FSI")
+            {
                 FSIenabled_ = true;
+            }
+
             if (module == "FF")
+            {
                 FFenabled_ = true;
+            }
         }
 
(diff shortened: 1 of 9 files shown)


DavidSCN (commented):
This is somehow necessary for`clang-format`, right? Usually clang should realize that you skipped the brackets and indentation should be correct. What am I missing here?

MakisH:
> This is somehow necessary for`clang-format`, right? Usually clang should realize that you skipped the brackets and indentation should be correct. What am I missing here?

It is not due to formatting, I just noticed that we had this maintainability problem after formatting, which could happen also without formatting. See #174 for the reasoning.

DavidSCN (approved):
I didn't went through all the remaining code, but looks good so far.

PR #176, opened 2021-04-29
Replace deprecated lookupType and subDictPtr with get and findDict

### LINKED ISSUE #172: Replace deprecated lookupType() with get()
We now agree to support the latest OpenCFD version by default and only release archives for the other versions (see #52 ).

We could then replace the deprecated `lookupType<>()` with `get<>()`, supported since at least v1812, to reduce deprecation warnings at compile-time: https://github.com/precice/openfoam-adapter/blob/master/Adapter.C#L44

MakisH:
Closes #172.

This should work with all ESI versions since ~~v1806~~ (update: since v1812). I want to check building this with that v1806 and maybe v1712 before merging.

Now, building with v2012 gives no warnings! :tada:

@DavidSCN mainly FYI, you may want to have a look at the diff.

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -37,19 +37,16 @@ bool preciceAdapter::Adapter::configFileRead()
                 IOobject::NO_WRITE));
 
         // Read and display the preCICE configuration file name
-        // NOTE: lookupType<T>("name") is deprecated in openfoam.com since v1812,
-        // which recommends get<T>("name") instead. However, get<T>("name")
-        // is not implemented in openfoam.org at the moment.
-        preciceConfigFilename_ = preciceDict.lookupType<fileName>("preciceConfig"); // We use this deprecated lookupType<>() for backwards compatibility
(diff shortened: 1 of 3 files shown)


DavidSCN:
> @DavidSCN mainly FYI, you may want to have a look at the diff.

I want to have a look at the compiler output :wink: Very cool!

MakisH:
I checked with the following versions:

| version | `.get<>()` | `.findDict()` |
---|---|---
| v1712 | :x: | :x: |
| v1806 | :heavy_check_mark: | :x: |
| v1812 | :heavy_check_mark: | :heavy_check_mark: |

DavidSCN:
Ok, then the decision is up to you.If you still want to merge it we should document the compatibility somewhere.

MakisH:
> Ok, then the decision is up to you.If you still want to merge it we should document the compatibility somewhere.

Good point, I updated the documentation page here.

PR #177, opened 2021-04-29
Update master (v1.0.0)

### LINKED ISSUE #52: Releases and versioning

MakisH:
This is the first versioned release, so it also closes #52. See the file `CHANGELOG.md` for details.

TODO:
- [x] Merge changelog
- [x] Add a version message

@DavidSCN do we also need to merge anything else first?


DavidSCN:
No, we can still merge the automated sync later on.

DavidSCN (approved):

PR #180, opened 2021-05-05
Disable execute at start for OpenFOAM8

### LINKED ISSUE #179: functionObjects executed at start in OpenFOAM8
In https://github.com/OpenFOAM/OpenFOAM-dev/commit/2ebed5ec7123ff8750931a65ab57a7bc7683aaa1, the default behavior of `functionObjects` has been changed and extended to that they are executed already before the simulation starts, i.e. the `execute` subroutine is called before we start a time step. 
In terms of our adapter, this means we write to and advance preCICE initially with the initial condition of our simulation. That might be in particular a problem (see also https://github.com/precice/openfoam-adapter/pull/130#issuecomment-831483248) if we have zero initial condition (initial field also with zero velocity) since our solid solver doesn't receive any loads during the first time window and (if everything goes well) calculates a zero displacement. I guess it's more or less undefined in terms of convergence (division by zero). However, preCICE raises a warning and emits convergence at this point (after one iteration)

```bash
---[precice]  relative convergence measure: relative two-norm diff = inf, limit = 0.005, normalization = 0, conv = true
---[precice]  relative convergence measure: relative two-norm diff = inf, limit = 0.005, normalization = 0, conv = true
---[precice]  All converged
---[precice] WARNING:  The coupling residual equals almost zero. There is maybe something wrong in your adapter. Maybe you always write the same data or you call advance without providing new data first or you do not use available read data. Or you just converge much further than actually necessary.
```

Afterwards, the simulation starts more or less as usual, we just lag one time step behind the correct time series. Still, the new default behavior is certainly not what we want and we actually have an option for writing initial data in the preCICE configuration in case this is desired by the user.

Luckily, the new default option is a public boolean: We can simply add the following line
```c++
    this->executeAtStart_ = false;
```
in the constructor of our `functionObject`

https://github.com/precice/openfoam-adapter/blob/aab511c0756f9dc50b542d2c9166502b5ee56a9d/preciceAdapterFunctionObject.C#L52-L61

Usually, the parameter is intended to be set at run time in the `controlDict` (it is one of the default parameters every functionObject reads), but we don't read these base class parameters at all, so disabling this option in the code by default should be just fine and doesn't require any changes in the configuration of the tutorial.

Any comment?

DavidSCN:
I cannot run `clang-format` 
```
YAML:22:16: error: invalid boolean
AlignOperands: AlignAfterOperator
               ^~~~~~~~~~~~~~~~~~
Error reading /home/schneidd/precice/openfoam-adapter/.clang-format: Invalid argument
```

and I cannot find the changelog-entries directory. Is the contributor supposed to create it at every time. 
Closes #179 

- <del>[ ] I updated the documentation in `docs/`<del>

### DIFF
--- a/changelog-entries/180.md
+++ b/changelog-entries/180.md
@@ -1,0 +1,1 @@
+Disabled execute at start for OpenFOAM8 (issue #179)
--- a/changelog-entries/180.md
+++ b/changelog-entries/180.md
@@ -1,1 +1,1 @@
-Disabled execute at start for OpenFOAM8 (issue #179)
+Disabled execute at start for OpenFOAM8 (PR #180, issue #179)
--- a/preciceAdapterFunctionObject.C
+++ b/preciceAdapterFunctionObject.C
@@ -23,6 +23,7 @@ License
 \*---------------------------------------------------------------------------*/
 
 #include "preciceAdapterFunctionObject.H"
+#include "Utilities.H"
 
 // OpenFOAM header files
 #include "Time.H"
@@ -72,6 +73,9 @@ Foam::functionObjects::preciceAdapterFunctionObject::~preciceAdapterFunctionObje
 
 bool Foam::functionObjects::preciceAdapterFunctionObject::read(const dictionary& dict)
 {
+    // We disable executeAtStart explicitly here and don't read the respective entry from the dictionary(fvMeshFunctionObject::read(dict)). Have a look at issue #179 for the rationale.
+    this->executeAtStart_ = false;
+    DEBUG(adapterInfo("Execute at start set to (dictionary input is ignored): " this->executeAtStart()));
     adapter_.configure();
 
     return true;


MakisH:
> I cannot run `clang-format`
> 
> ```
> YAML:22:16: error: invalid boolean
> AlignOperands: AlignAfterOperator
>                ^~~~~~~~~~~~~~~~~~
> Error reading /home/schneidd/precice/openfoam-adapter/.clang-format: Invalid argument
> ```

Which version? I used clang-format 11 here, as I wanted as many options as possible, to fit the OpenFOAM format as closely as possible.

> and I cannot find the changelog-entries directory. Is the contributor supposed to create it at every time.
> Closes #179

Only the first time after a release, to keep `master` cleaner. But we could also add a step to recreate this directory when merging back to develop and bumping the version to unreleased.

DavidSCN:
> > I cannot run `clang-format`
> > ```
> > YAML:22:16: error: invalid boolean
> > AlignOperands: AlignAfterOperator
> >                ^~~~~~~~~~~~~~~~~~
> > Error reading /home/schneidd/precice/openfoam-adapter/.clang-format: Invalid argument
> > ```
> 
> Which version? I used clang-format 11 here, as I wanted as many options as possible, to fit the OpenFOAM format as closely as possible.
> 

`clang-format version 10.0.0-4ubuntu1 `

> > and I cannot find the changelog-entries directory. Is the contributor supposed to create it at every time.
> > Closes #179
> 
> Only the first time after a release, to keep `master` cleaner. But we could also add a step to recreate this directory when merging back to develop and bumping the version to unreleased.

Ok, will add an entry and create the directory

changelog-entries/180.md
```diff
@@ -0,0 +1 @@
+Disabled execute at start for OpenFOAM8 (issue #179)
```
MakisH, line 1: Please add the PR number here, not the issue.

DavidSCN: Ok, I thought the name of the file would be sufficient.

DavidSCN: Added it.

PR #182, opened 2021-05-20
Add volume coupling in Interface class

Alphaoo1:
The PR contains the edits required for doing Volume coupling.

To keep this PR simple I did not add the usage of the volume field.
Another PR can be opened for that, specifically Generic volume scalar field and Generic volume vector field.

This is in regards to issue[ #591](https://precice.discourse.group/t/extending-openfoam-adapter-for-volumetric-coupling/591) opened on Discourse.

TODO list:

- [x] add changelog md

### DIFF
--- a/182.md
+++ b/182.md
@@ -1,0 +1,40 @@
+# ChangeLog: Volume Coupling for OpenFOAM v2012 adapter PR #182
+
+The PR contains the edits required for doing Volume coupling.
+
+Added `locationsType == "volume"`
+
+To keep this PR simple I did not add the usage of the volume field.
+Another PR can be opened for that, specifically Generic volume scalar field and Generic volume vector field.
+
+This is in regards to issue[ #591](https://precice.discourse.group/t/extending-openfoam-adapter-for-volumetric-coupling/591) opened on Discourse.
+
+## Usage: 
+
+in `Fluid/system/preciceDict`
+
+```
+
+interfaces
+{
+  Interface1
+  {
+    mesh              Fluid-Mesh-PV;
+    patches           ();
+    locations         volume;
+    
+    readData
+    (
+        Volume_field_data_to_read
+    );
+    
+    writeData
+    (
+        Volume_field_data_to_write
+    );
+  };
+};
+
+```
+
+Ofcourse the volume dataFields `Volume_field_data_to_read` and `Volume_field_data_to_write` needs to be added to a module in order to use them.
--- a/182.md
+++ b/182.md
@@ -37,4 +37,4 @@ interfaces
 
 ```
 
-Ofcourse the volume dataFields `Volume_field_data_to_read` and `Volume_field_data_to_write` needs to be added to a module in order to use them.
+Of course the volume dataFields `Volume_field_data_to_read` and `Volume_field_data_to_write` needs to be added to a module in order to use them.
--- a/182.md
+++ b/182.md
@@ -7,7 +7,7 @@ Added `locationsType == "volume"`
 To keep this PR simple I did not add the usage of the volume field.
 Another PR can be opened for that, specifically Generic volume scalar field and Generic volume vector field.
 
-This is in regards to issue[ #591](https://precice.discourse.group/t/extending-openfoam-adapter-for-volumetric-coupling/591) opened on Discourse.
+This is in regards to issue[#591](https://precice.discourse.group/t/extending-openfoam-adapter-for-volumetric-coupling/591) opened on Discourse.
 
 ## Usage: 
 
--- a/182.md
+++ b/182.md
@@ -9,12 +9,11 @@ Another PR can be opened for that, specifically Generic volume scalar field and 
 
 This is in regards to issue[#591](https://precice.discourse.group/t/extending-openfoam-adapter-for-volumetric-coupling/591) opened on Discourse.
 
-## Usage: 
+## Usage
 
 in `Fluid/system/preciceDict`
 
-```
-
+```cpp
 interfaces
 {
   Interface1
--- a/Interface.C
+++ b/Interface.C
@@ -298,7 +298,6 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
             {
                 vertices[verticesIndex++] = CellCenters[i].z();
             }
-
         }
 
         // Get the locations of the mesh vertices (here: face centers)
@@ -324,7 +323,7 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
         // Pass the mesh vertices information to preCICE
         precice_.setMeshVertices(meshID_, numDataLocations_, vertices, vertexIDs_);
     }
-else
+    else
     {
         FatalErrorInFunction
             << "ERROR: interface points location type "
--- a/Interface.C
+++ b/Interface.C
@@ -272,7 +272,7 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
         for (uint j = 0; j < patchIDs_.size(); j++)
         {
             numDataLocations_ +=
-                    mesh.boundaryMesh()[patchIDs_.at(j)].faceCentres().size();
+                mesh.boundaryMesh()[patchIDs_.at(j)].faceCentres().size();
         }
         DEBUG(adapterInfo("Number of coupling volumes: " + std::to_string(numDataLocations_)));
 
--- a/Interface.C
+++ b/Interface.C
@@ -317,7 +317,7 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
         // Pass the mesh vertices information to preCICE
         precice_.setMeshVertices(meshID_, numDataLocations_, vertices, vertexIDs_);
     }
-    if (!(locationsType_ == "faceNodes" || locationsType_ == "faceCenters" || locationsType_ == "faceCentres" || locationsType_ == "volume"))
+else
     {
         FatalErrorInFunction
             << "ERROR: interface points location type "
--- a/Interface.C
+++ b/Interface.C
@@ -314,7 +314,10 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
             {
                 vertices[verticesIndex++] = faceCenters[i].x();
                 vertices[verticesIndex++] = faceCenters[i].y();
-                vertices[verticesIndex++] = faceCenters[i].z();
+                if (dim_ == 3)
+                {
+                    vertices[verticesIndex++] = faceCenters[i].z();
+                }
             }
         }
 
--- a/Interface.C
+++ b/Interface.C
@@ -274,7 +274,7 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
             numDataLocations_ +=
                     mesh.boundaryMesh()[patchIDs_.at(j)].faceCentres().size();
         }
-        DEBUG(adapterInfo("Number of face centres: " + std::to_string(numDataLocations_)));
+        DEBUG(adapterInfo("Number of coupling volumes: " + std::to_string(numDataLocations_)));
 
         // Array of the mesh vertices.
         // One mesh is used for all the patches and each vertex has 3D coordinates.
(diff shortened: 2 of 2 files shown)


182.md
```diff
@@ -0,0 +1,40 @@
...
+    writeData
+    (
+        Volume_field_data_to_write
+    );
+  };
+};
+
+```
+
+Ofcourse the volume dataFields `Volume_field_data_to_read` and `Volume_field_data_to_write` needs to be added to a module in order to use them.
```
DavidSCN, line 40:
```suggestion
Of course the volume dataFields `Volume_field_data_to_read` and `Volume_field_data_to_write` needs to be added to a module in order to use them.
```

Interface.C
```diff
@@ -255,7 +255,69 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
     }
-    if (!(locationsType_ == "faceNodes" || locationsType_ == "faceCenters" || locationsType_ == "faceCentres"))
+    else if (locationsType_ == "volume")
+    {
+        // module for volume coupling is based on the module for coupling faceCenters
+        // as the module was tested and developed on a previous development where faceNodes
+        // did not work properly
+        //
+        // the module for volume coupling considers the mesh points in the volume and
+        // on the patches in order to take the boundary conditions into account
```
DavidSCN, line 265:
```suggestion
        // on the boundary patches in order to take the boundary conditions into account
```

Interface.C
```diff
@@ -255,7 +255,69 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+        numDataLocations_ = mesh.C().size();
+
+        // Count the data locations for all the patches
+        // and add the those to the previously determined number of mesh points in the volume
+        for (uint j = 0; j < patchIDs_.size(); j++)
+        {
+            numDataLocations_ +=
+                    mesh.boundaryMesh()[patchIDs_.at(j)].faceCentres().size();
+        }
+        DEBUG(adapterInfo("Number of face centres: " + std::to_string(numDataLocations_)));
```
DavidSCN, line 277:
```suggestion
        DEBUG(adapterInfo("Number of coupling volumes: " + std::to_string(numDataLocations_)));
```
We should distinguish here between the boundary coupling and the volume coupling.

Interface.C
```diff
@@ -255,7 +255,69 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+
+        // Array of the mesh vertices.
+        // One mesh is used for all the patches and each vertex has 3D coordinates.
+        double vertices[3 * numDataLocations_];
+
+        // Array of the indices of the mesh vertices.
+        // Each vertex has one index, but three coordinates.
+        vertexIDs_ = new int[numDataLocations_];
```
DavidSCN, line 285: We should better use `std::vector` instead of c-style arrays, but I just saw that we did it elsewhere the same way. So , we should probably do it in a follow up.

Interface.C
```diff
@@ -255,7 +255,69 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+
+        // Get the locations of the volume centered mesh vertices
+        const vectorField & CellCenters = mesh.C();
+
+        for (int i = 0; i < CellCenters.size(); i++)
+        {
+            vertices[verticesIndex++] = CellCenters[i].x();
+            vertices[verticesIndex++] = CellCenters[i].y();
+            vertices[verticesIndex++] = CellCenters[i].z();
+        }
```
DavidSCN, line 298: This wouldn't be compatible with preCICE in 2D. You could simply add an inner loop going through the dimension, as it is done above.

Interface.C
```diff
@@ -255,7 +255,69 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+        for (uint j = 0; j < patchIDs_.size(); j++)
+        {
+            numDataLocations_ +=
+                    mesh.boundaryMesh()[patchIDs_.at(j)].faceCentres().size();
+        }
+        DEBUG(adapterInfo("Number of face centres: " + std::to_string(numDataLocations_)));
+
+        // Array of the mesh vertices.
+        // One mesh is used for all the patches and each vertex has 3D coordinates.
+        double vertices[3 * numDataLocations_];
```
DavidSCN, line 281: The dimension is here hard-coded. Please use `dim_`,which is related to the configured dimension.

Interface.C
```diff
@@ -255,7 +255,69 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+            const vectorField & faceCenters =
+                    mesh.boundaryMesh()[patchIDs_.at(j)].faceCentres();
+
+            // Assign the (x,y,z) locations to the vertices
+            for (int i = 0; i < faceCenters.size(); i++)
+            {
+                vertices[verticesIndex++] = faceCenters[i].x();
+                vertices[verticesIndex++] = faceCenters[i].y();
+                vertices[verticesIndex++] = faceCenters[i].z();
+            }
```
DavidSCN, line 314: Same as above.

Interface.C
```diff
@@ -255,7 +255,69 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+        }
+
+        // Pass the mesh vertices information to preCICE
+        precice_.setMeshVertices(meshID_, numDataLocations_, vertices, vertexIDs_);
+    }
+    if (!(locationsType_ == "faceNodes" || locationsType_ == "faceCenters" || locationsType_ == "faceCentres" || locationsType_ == "volume"))
```
DavidSCN, line 320:
```suggestion
else
```

DavidSCN (commented):
Some smaller comments. Very nice, thanks for your efforts. Have you already tested your implementation? It would be nice to have a smaller test case at hand in order to run your PRs.

Interface.C
Alphaoo1: Yes, good point. 
Will make the required changes.

Interface.C
Alphaoo1: Will do.

Interface.C
Alphaoo1: Yes, we use 3D, so I guess it was hard coded, but yes for a general user, will be better to use `dim_`.

Interface.C
Alphaoo1: Good one, this is easy to miss when implementing a new `locationsType`.
If others are ok with it, then we can keep this change.
I was under the impression that this was used as a safeguard on spelling mistakes.

Alphaoo1:
> Some smaller comments. Very nice, thanks for your efforts. Have you already tested your implementation? It would be nice to have a smaller test case at hand in order to run your PRs.

All the volume coupling done with XDEM is done with this implementation.
Could I get some suggestions, because we use this for OpenFOAM + XDEM coupling?
The first participant is of course OF.
So will need a suggestion for 2nd participant who receives the volume data.

DavidSCN:
> > Some smaller comments. Very nice, thanks for your efforts. Have you already tested your implementation? It would be nice to have a smaller test case at hand in order to run your PRs.
> 
> All the volume coupling done with XDEM is done with this implementation.
> Could I get some suggestions, because we use this for OpenFOAM + XDEM coupling?
> The first participant is of course OF.
> So will need a suggestion for 2nd participant who receives the volume data.

I would propose to do an OpenFOAM-OpenFOAM coupling example, which is as simple as it can be, maybe even uni-directional. One could choose solver according to your supported coupling data fields.

Interface.C
Alphaoo1: I had a discussion with my colleague Xavier BESSERON, he pointed out that since we are working with the volume we would by default need 3 dimensions, and to enforce this the dimensions were hard-coded as `3`.

But maybe some users or test-cases might need this option?

DavidSCN: > But maybe some users or test-cases might need this option?

I don't see any reason to implement this only in 3D. Simple test cases will probably also be in 2D.

Interface.C
```diff
@@ -294,7 +294,10 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
         {
             vertices[verticesIndex++] = CellCenters[i].x();
             vertices[verticesIndex++] = CellCenters[i].y();
-            vertices[verticesIndex++] = CellCenters[i].z();
+            if (dim_ == 2)
```
DavidSCN, line 297: Typo, you need to compare against `dim_==3`

Alphaoo1:
I have added the `dim_` to the volume coupling.

Interface.C
```diff
@@ -255,7 +255,73 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+            // Get the face centers of the current patch
+            const vectorField & faceCenters =
+                    mesh.boundaryMesh()[patchIDs_.at(j)].faceCentres();
+
+            // Assign the (x,y,z) locations to the vertices
+            for (int i = 0; i < faceCenters.size(); i++)
+            {
+                vertices[verticesIndex++] = faceCenters[i].x();
+                vertices[verticesIndex++] = faceCenters[i].y();
+                vertices[verticesIndex++] = faceCenters[i].z();
```
MakisH, line 317: Why not distinguish the 3D case also here?

MakisH:
Thanks a lot for contributing, @Alphaoo1!

> To keep this PR simple I did not add the usage of the volume field.
> Another PR can be opened for that, specifically Generic volume scalar field and Generic volume vector field.

Since I can see this developing into a larger project, I created a branch `volume-coupling` and I am updating the target of this PR. Let's target all related pull requests there. Thank you for starting small, it makes it easier to review!

Alphaoo1:
@MakisH I have made the changes.
Thank you for making a new branch for `volume-coupling`.

I had a question regarding opening new PR based on this branch.
With a PR in mind for `Volume_Coupling ` module, and then another PR for adding data fields to it. 
I can open the PRs for this without the current changes merged.

Shouldn't this PR be merged first to ultimately use the volume coupling for the proposed PRs?

MakisH (approved):
This looks good for now. I am merging to the `volume-coupling`, let's go to the next step.

MakisH:
@Alphaoo1 the automatic workflows still report a few small formatting issues. Could you please fix them?

https://github.com/precice/openfoam-adapter/pull/182/checks?check_run_id=2761909893
https://github.com/precice/openfoam-adapter/pull/182/checks?check_run_id=2761909898

DavidSCN (approved):

MakisH:
@Alphaoo1 thank you for trying to fix the formatting, I think this is a useful experience! I approved & ran the pipeline again, there are still a few issues: https://github.com/precice/openfoam-adapter/pull/182/checks?check_run_id=2781012427

Did you use the script `tools/format-code.sh` here? https://github.com/precice/openfoam-adapter/blob/develop/CONTRIBUTING.md

MakisH:
@Alphaoo1 I fixed the formatting on my system, let's keep an eye on this if it happens again in the next PRs.

It would also be good to always work on a different branch on your fork, for multiple reasons, mainly:

1. You can work on different things at the same time, without waiting for previous PRs to be merged.
2. When I checkout your branch, it will not be `master` (a bit confusing with the main `master`)

Alphaoo1:
> @Alphaoo1 I fixed the formatting on my system, let's keep an eye on this if it happens again in the next PRs.
> 
> It would also be good to always work on a different branch on your fork, for multiple reasons, mainly:
> 
> 1. You can work on different things at the same time, without waiting for previous PRs to be merged.
> 2. When I checkout your branch, it will not be `master` (a bit confusing with the main `master`)

Thank you for the code formatting. I had tried the script, but I was getting errors. So I tried to fix the file manually :sweat_smile: , hence the message on gitter for clang rules. But yes I will keep an eye out. 

I did not understand your 2nd comment. Is it that the default branch is not `master` or the changes I have made are not in `master`?

MakisH:
The main reason I wrote this was actually a technical issue: I used the GitHub command line interface to checkout this PR, which somehow replaced my `master` with your `master`. But in general, the best and most common practice is to open pull requests from a branch which you can then throw away.

Alphaoo1:
@MakisH Yes, I will be creating new branches based on `volume_coupling` for further development.

PR #185, opened 2021-06-29
Bump default OpenFOAM version to v2106

MakisH:
OpenFOAM v2106 was released yesterday. The adapter builds and the tutorials run without any changes. This updates the documentation and GitHub Actions. I will then merge to `master`, without releasing a new version.

@DavidSCN please give me a quick sanity check on the changes and branches.

TODO list:


DavidSCN (approved):
Looks good. Do we have a backward check four our 'oldest' supported version? 
No need to check right now, but it might be valuable to trigger something, maybe `on_push` to develop (not on every PR).

MakisH:
> Looks good. Do we have a backward check four our 'oldest' supported version?
> No need to check right now, but it might be valuable to trigger something, maybe `on_push` to develop (not on every PR).

Nothing changed in the code, so I did not check other versions. We currently only check automatically for the latest ESI OpenFOAM. This is a separate issue.

Issue #187, opened 2021-07-13
Wrong result files of temporarily registered objects


DavidSCN:
In implicit coupling schemes, we currently accept that result files are generated already during the first iteration and force OpenFOAM later within the adapter to regenerate the result files:

https://github.com/precice/openfoam-adapter/blob/fca5e8770cba964cd9b51ce97ab8d608f5a75ca8/Adapter.C#L414-L431

However, some solver (I guess only a few, potentially also in-house solver) postprocess results and register new fields during the generation of output files, e.g., the `laplacianFoam` solver computes the gradients from the solution and registers them temporarily before calling `write()`

https://develop.openfoam.com/Development/openfoam/-/blob/master/applications/solvers/basic/laplacianFoam/write.H

These temporarily registered fields will not be updated when we trigger the result generation in the adapter again since the fields are not registered at this point. As a consequence, we obtain results belonging to the first coupling iteration,e.g., a `gradTx` field (stored as usual, thus, declared as a final result) belonging to the first implicit coupling iteration and a corresponding (and actually correct) `T` field belonging to the converged solution of the coupled problem.

A clean solution here would probably be to prevent the result generation during the first coupling iteration by resetting the `runTime` object when we reload the checkpoint so that the `runTime` object generates the result itself when we converged during the last iteration.

PR #188, opened 2021-07-13
Fixed misplaced data reading in advance function

DavidSCN:
In the final implicit coupling time-step, we first read new coupling data and afterwards write the result files to the system. This leads of course to inconsistent result files, as we store now the converged solution of the past time  step + the coupling data of the first iteration during the new time-step, which might look like this:

 
Note also the inconsistency between the gradients/flux and the coupling variable `T`.  Not completely sure about the reason but the misbehavior is more obvious in case OpenFOAM is the `second` participant (as opposed to most of our tutorials).

We also stored the new coupling data within the new checkpoint, which is (@uekerman corrects me otherwise) wrong (inconsistent). The error is hidden though since we override the coupling data again after reloading the checkpoint.

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -395,9 +395,6 @@ void preciceAdapter::Adapter::execute()
         fulfilledReadCheckpoint();
     }
 
-    // Read the received coupling data from the buffer
-    readCouplingData();
-
     // Adjust the timestep, if it is fixed
     if (!adjustableTimestep_)
     {
@@ -429,6 +426,9 @@ void preciceAdapter::Adapter::execute()
             const_cast<Time&>(runTime_).writeNow();
         }
     }
+
+    // Read the received coupling data from the buffer
+    readCouplingData();
 
     // If the coupling is not going to continue, tear down everything
     // and stop the simulation.
--- a/changelog-entries/188.md
+++ b/changelog-entries/188.md
@@ -1,0 +1,1 @@
+- Fixed the misplaced data reading in the adapter 'advance' function [#188](https://github.com/precice/openfoam-adapter/pull/188)


uekerman:
Could you please add more information on the plot, over which line are you plotting, where is the coupling interface?

> Not completely sure about the reason but the misbehavior is more obvious in case OpenFOAM is the second participant (as opposed to most of our tutorials).

This could make sense. The second participant computes the acceleration and maybe also extrapolation.

I don't understand your change thought, there is no OpenFOAM output between your deleted lines and your new lines? 

> We also stored the new coupling data within the new checkpoint, which is (@uekerman corrects me otherwise) wrong (inconsistent). The error is hidden though since we override the coupling data again after reloading the checkpoint.

I would not call this inconsistent. There is no fixed order between `readData` and `writeCheckpoint` as the data is overwritten anyway as you correctly point out.

DavidSCN:
> Could you please add more information on the plot, over which line are you plotting, where is the coupling interface?

Sure. I'm plotting over the x-axis of the partitioned-heat test case and the interface is located on the right (x=1).

> I don't understand your change though, there is no OpenFOAM output between your deleted lines and your new lines?

There is OpenFOAM output for lines over the newly inserted lines (you may need to expand the code in the file view).

> I would not call this inconsistent. There is no fixed order between readData and writeCheckpoint as the data is overwritten anyway as you correctly point out.

Then how would you call it? If someone would operate on the reloaded data field before we ask preCICE for new coupling data, the data set would still consist of a converged solution + coupling data of the first iteration.

uekerman:
> There is OpenFOAM output for lines over the newly inserted lines (you may need to expand the code in the file view).

I did. There is only the 

```c++
adapterInfo(
                "The coupling timestep completed. "
                "Writing the updated results.",
                "info");
```

?

> Then how would you call it? If someone would operate on the reloaded data field before we ask preCICE for new coupling data, the data set would still consist of a converged solution + coupling data of the first iteration.

Yes, but I guess you could also screw it up the other way around. You need to have a consistent solution when you read and write what in your adapter and when you output data, but there is not the one solution for all adapters. I guess we treat this differently in the different adapters already now.

DavidSCN:
> I did. There is only the

And in the next line, following your adapterInfo, there should be a `writeNow()`

https://github.com/DavidSCN/openfoam-adapter/blob/5502394c4f8d42bc0ce0f5c7354d958bf9960cde/Adapter.C#L426

uekerman:
Ah ok, thanks for pointing out!
I thought this only prints the current time to the terminal.

DavidSCN:
Further concerns before merging here @MakisH ?

MakisH (approved):
The argumentation sounds reasonable to me, especially if you also observe reasonable changes in the results.

As context, this adapter is a bit special because it has to do all the preCICE steps at a single code injection point. This has led to a few "hacks", such as the re-writing of results. Unfortunately, I don't have more decision documentation than what is already in the comments.

I had considered several factors, such as subcycling with adjustable timestep size during implicit coupling, but I had not looked into intermediate results of implicit coupling (since the converged results are the ones that matter).

DavidSCN:
We have anyway #187 as a reminder and we can try out things in the future. Let's merge here since the most obvious drawback was fixed with this simple change.

PR #192, opened 2021-07-28
Fix potentially failing pkg-config commands in Allwmake

### LINKED ISSUE #190: Missing exit message in `Allwmake`
Due to the immediate exit
https://github.com/precice/openfoam-adapter/blob/a5c01a0993c34787c146c70368dd01cec37673e0/Allwmake#L2
 the `Allwmake` script 'exits' as follows:

```
Building the OpenFOAM-preCICE adapter...
```
in case the pkg-config is not discoverable. There is actually a message missing, which tells the user what went wrong.

DavidSCN:
Fixes #190

### DIFF
--- a/Allwmake
+++ b/Allwmake
@@ -32,9 +32,16 @@ log "Building the OpenFOAM-preCICE adapter..."
 # Export the environment variables
 export ADAPTER_PREP_FLAGS
 export ADAPTER_TARGET_DIR
-ADAPTER_PKG_CONFIG_CFLAGS="$(pkg-config --silence-errors --cflags libprecice)"
+
+if pkg-config libprecice; then
+  ADAPTER_PKG_CONFIG_CFLAGS="$(pkg-config --silence-errors --cflags libprecice)"
+  ADAPTER_PKG_CONFIG_LIBS="$(pkg-config --silence-errors --libs libprecice)"
+else
+  log "No configuration file 'libprecice.pc' in the PKG_CONFIG_PATH was found."
+  ADAPTER_PKG_CONFIG_CFLAGS=""
+  ADAPTER_PKG_CONFIG_LIBS=""
+fi
 export ADAPTER_PKG_CONFIG_CFLAGS
-ADAPTER_PKG_CONFIG_LIBS="$(pkg-config --silence-errors --libs libprecice)"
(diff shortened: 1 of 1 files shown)


MakisH (approved):
Good addition, it covers yet another case in which things can go wrong.

I improved some minor aspects, please squash before merging (or revert).

Allwmake
```diff
@@ -32,11 +32,24 @@ log "Building the OpenFOAM-preCICE adapter..."
...
 export ADAPTER_PREP_FLAGS
 export ADAPTER_TARGET_DIR
-ADAPTER_PKG_CONFIG_CFLAGS="$(pkg-config --silence-errors --cflags libprecice)"
+
+if pkg-config libprecice; then
+  ADAPTER_PKG_CONFIG_CFLAGS="$(pkg-config --silence-errors --cflags libprecice)"
+  ADAPTER_PKG_CONFIG_LIBS="$(pkg-config --silence-errors --libs libprecice)"
+else
+  log "WARNING: No configuration file 'libprecice.pc' was found in the PKG_CONFIG_PATH."
+  log "Check the preCICE documentation page 'Linking to preCICE' for more details, or proceed if you know what you are doing."
```
DavidSCN, line 41: Do you mean here by `know what you are doing' someone who exports all necessary paths?

MakisH: For example, yes. We want that the "regular" user uses pkg-config, but there are also other such ways one could install preCICE and make it discoverable.

DavidSCN: Alright. Is 'Linking to precice' the right place to look at/ is there the relevant information? or do we actually want to encourage people to export the `PKG_CONFIG_PATH`? Maybe we should add a few sentences in the adapter wiki and refer to it here?!

MakisH: We currently describe how to export the `PKG_CONFIG_PATH` in the "linking to preCICE" page: https://precice.org/installation-linking.html#make-and-scripts-in-general

I am not sure if we should use URLs in error messages, they tend to get outdated and forgotten.

DavidSCN: I wouldn't put the link here either. I'm just afraid that the description there is too developer oriented. I think the right place to look at for these users would be the Troubleshooting section on the same page.

DavidSCN: However, then let's go with this description.

PR #195, opened 2021-08-19
Remove explicit casting of boundary conditions

DavidSCN:
when writing gradients as this rises the opportunity to select other boundary conditions at the interface such as groovyBCs.

<del> - [ ] I updated the documentation in `docs/` <del/>

### DIFF
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -28,8 +28,7 @@ void preciceAdapter::CHT::HeatFlux::write(double* buffer, bool meshConnectivity,
         int patchID = patchIDs_.at(j);
 
         const scalarField gradientPatch(
-            refCast<const fixedValueFvPatchScalarField>(
-                T_->boundaryField()[patchID])
+            (T_->boundaryField()[patchID])
                 .snGrad());
 
         // Extract the effective conductivity on the patch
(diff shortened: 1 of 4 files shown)


MakisH:
What are possible side-effects of this?

I like explicit casting as it makes the target type clear enough, and I was trying to find the most general type that can support the operations needed. However, if this makes our lives easier, we should definitely integrate it.

We could go one step further and apply it to every type of boundary condition we currently have. This would particularly important for the `inletOutlet` boundary conditions in fluid-fluid coupling.

DavidSCN:
> What are possible side-effects of this?

I don't have any in mind. Of course people can now use the field with any type of boundary conditions but it should be fine though as we still do the correct operations.. I think it widens the applicability of the adapter with some potential danger we or I am not aware of.

MakisH (commented):
In general, I approve this idea: Explicitly casting the type does not necessarily offer extra safety, as the actual boundary condition is provided at runtime, so the compiler cannot check if this is correct. Furthermore, it indeed allows more boundary conditions of similar types.

Let's add a changelog and expand it to more coupling data readers. We should check all the tutorials again, then.

DavidSCN:
Ready to go from my side.

DavidSCN:
Tested it

PR #201, opened 2021-11-05
Print DEBUG output on each rank

### LINKED ISSUE #136: Incomplete debug output on number of nodes/faces in parallel
When running parallel cases with the adapter, it just prints the number of interface nodes/faces on rank 0. 
Hence, in larger applications, the output might give the following:
```
---[preciceAdapter] [DEBUG] Creating interfaces...
---[preciceAdapter] [DEBUG] Number of face centres: 0
---[preciceAdapter] [DEBUG] Interface created on mesh Fluid-Mesh-Centers
---[preciceAdapter] [DEBUG] Adding coupling data writers...
---[preciceAdapter] [DEBUG] Added writer: Stress.
---[preciceAdapter] [DEBUG] Adding coupling data readers...
---[preciceAdapter] [DEBUG] Number of face nodes: 0
```
though there are actually a few hundred coupling nodes on other processes. 

To fix it, we just need here a sum over all processes

https://github.com/precice/openfoam-adapter/blob/b1d80efe3c2ec964f157687753a675dd4502dba7/Interface.C#L72-L74

DavidSCN:
...which is very helpful. It also kind of fixes #136. Please check it out and merge.

TODO list:

- [ ] <del>I updated the documentation in `docs/` </del>

### DIFF
--- a/Utilities.C
+++ b/Utilities.C
@@ -58,7 +58,7 @@ void adapterInfo(const std::string message, const std::string level)
     }
     else if (level.compare("debug") == 0)
     {
-        Info << INFO_STR_ADAPTER
+        Pout << INFO_STR_ADAPTER
              << "[DEBUG] "
              << message.c_str()
              << nl;
--- a/changelog-entries/201.md
+++ b/changelog-entries/201.md
@@ -1,0 +1,1 @@
+- Extended the adapter DEBUG output to print the rank in addition to the message [#201](https://github.com/precice/openfoam-adapter/pull/201)


MakisH (approved):
Good idea! :+1:

My first thought was "but this will be too verbose when using multiple ranks", but then I thought "we probably don't want any debug messages in too large cases anyway".

Here are a few [`Pout` usage examples](https://develop.openfoam.com/search?search=Pout+%3C%3C&group_id=23&project_id=47&scope=&search_code=true&snippets=false&repository_ref=master&nav_source=navbar) - there also seems to be a `DebugPout`, but I guess it `Pout` is the right for us, as the debug mode of the adapter is independent of the debug mode of OpenFOAM. I also see that `Pout` also exists in OpenFOAM.org with the same name, so we should remain compatible.

I have not tried to run the code, but it should just work.

PR #203, opened 2021-11-05
Fix 2D check for parallel cases

### LINKED ISSUE #202: OpenFOAM tutorial fails in parallel
This issue was reported on discourse and I can reproduce it locally: https://precice.discourse.group/t/cannot-run-openfoam-in-parallel-when-using-rbf-mapping/785

DavidSCN:
I'm actually a bit confused that this issue pops up that late. The error occurs only if a rank runs preCICE-wise on an empty partition in 2D. 

This might even be worth a bug-fix release. Fixes #202.

TODO list:

### DIFF
--- a/Interface.C
+++ b/Interface.C
@@ -102,16 +102,20 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
             {
                 const pointField faceNodes =
                     mesh.boundaryMesh()[patchIDs_.at(j)].localPoints();
+                const auto faceNodesSize = faceNodes.size();
                 //Allocate memory for z-coordinates
                 std::array<double, 2> z_location({0, 0});
                 constexpr unsigned int z_axis = 2;
 
                 // Find out about the existing planes
                 // Store z-coordinate of the first layer
-                z_location[0] = faceNodes[0][z_axis];
+                if (faceNodesSize > 0)
+                {
+                    z_location[0] = faceNodes[0][z_axis];
+                }
                 // Go through the remaining points until we find the second z-coordinate
                 // and store it (there are only two allowed in case we are in the xy-layer)
-                for (int i = 0; i < faceNodes.size(); i++)
+                for (int i = 0; i < faceNodesSize; i++)
                 {
                     if (z_location[0] == faceNodes[i][z_axis])
                     {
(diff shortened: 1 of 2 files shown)


MakisH (approved):
I understand that the problem appears while checking if a 2D case is properly aligned at the xy-plane, but the specific parallel rank does not have any nodes.

Why did this not appear before? Well, probably because, in our tutorials, we usually partition the OpenFOAM meshes to include the interface.

What I understand you are doing here: Check if there are any face nodes of the interface patches before trying to access the `faceNodes` array, which is a good idea anyway. This should only improve the situation, not cause any side effects (especially since `z_location` is already initialized).

You are storing the size into `faceNodesSize` (for each patch), which you are reusing further down now that you have it.

I have not tried to run this code, but it looks good to me. :+1:

changelog-entries/203.md
```diff
@@ -0,0 +1 @@
+- Fixed the out-of-bounds data access which might cause trouble in 2D parallel cases [#202](https://github.com/precice/openfoam-adapter/issues/202)
```
MakisH, line 1: I am not sure if the user can clearly understand what this means without looking at the code. Maybe, instead:

```suggestion
- Fixed a potential memory access issue in the xy-alignment check for 2D cases [#202](https://github.com/precice/openfoam-adapter/issues/202)
```

Interface.C
```diff
@@ -102,16 +102,20 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
                 //Allocate memory for z-coordinates
                 std::array<double, 2> z_location({0, 0});
                 constexpr unsigned int z_axis = 2;
 
                 // Find out about the existing planes
                 // Store z-coordinate of the first layer
-                z_location[0] = faceNodes[0][z_axis];
+                if (faceNodesSize > 0)
+                {
+                    z_location[0] = faceNodes[0][z_axis];
```
MakisH, line 114: This is fine now, but we could maybe have detected this situation a bit more easily if we had here `faceNodes.at(0)[z_axis]` (it would throw an exception instead of exploding).

DavidSCN:
Question remains: do we want to add a bugfix release?

Interface.C
DavidSCN: True, but won't change it here for style compatibility reasons.

MakisH:
> Question remains: do we want to add a bugfix release?

We should, and we can include a few more bugfixes. I will have a look this week.

DavidSCN:
> We should, and we can include a few more bugfixes. I will have a look this week.

Maybe I will have a look before, we don't have many commits since the last release ;)

PR #206, opened 2021-11-11
Add a check for supported data locationType

### LINKED ISSUE #204: Missing assertion for supported interface `location` type
Our current adapter can be misused with the available `locations` and unrelated data fields. Example: in our perpendicular flap tutorial, one can select the location type `faceNodes` (for forces and displacements) and the simulation will still work, although this doesn't make sense (the force don't live usually on the face centers). We should safeguard against this.

DavidSCN:
I also introduced a small enum for the locationType

TODO list:

<del> I updated the documentation in `docs/` </del>
- [x] add the data name to the error message

Closes #204

### DIFF
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -110,7 +110,7 @@ bool preciceAdapter::CHT::HeatFlux::isLocationTypeSupported(const bool meshConne
     if (meshConnectivity)
         return true;
     else
-        return (this->locationsType_ == LocationType::faceCenters);
+        return (this->locationType_ == LocationType::faceCenters);
 }
 
 std::string preciceAdapter::CHT::HeatFlux::getDataName() const
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -113,6 +113,11 @@ bool preciceAdapter::CHT::HeatFlux::isLocationTypeSupported(const bool meshConne
         return (this->locationsType_ == LocationType::faceCenters);
 }
 
+std::string preciceAdapter::CHT::HeatFlux::getDataName() const
+{
+    return "HeatFlux";
+}
+
 //----- preciceAdapter::CHT::HeatFlux_Compressible ----------------------------
 
 preciceAdapter::CHT::HeatFlux_Compressible::HeatFlux_Compressible(
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -103,6 +103,16 @@ void preciceAdapter::CHT::HeatFlux::read(double* buffer, const unsigned int dim)
     }
 }
 
+bool preciceAdapter::CHT::HeatFlux::isLocationTypeSupported(const bool meshConnectivity) const
+{
+    // We don't distinguish between reading and writing. Thus, we just return true
+    // in case there was connectivity defined
+    if (meshConnectivity)
+        return true;
+    else
+        return (this->locationsType_ == LocationType::faceCenters);
+}
+
 //----- preciceAdapter::CHT::HeatFlux_Compressible ----------------------------
 
 preciceAdapter::CHT::HeatFlux_Compressible::HeatFlux_Compressible(
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -105,12 +105,19 @@ void preciceAdapter::CHT::HeatFlux::read(double* buffer, const unsigned int dim)
 
 bool preciceAdapter::CHT::HeatFlux::isLocationTypeSupported(const bool meshConnectivity) const
 {
-    // We don't distinguish between reading and writing. Thus, we just return true
-    // in case there was connectivity defined
+    // For cases with mesh connectivity, we support:
+    // - face nodes, only for writing
+    // - face centers, only for reading
+    // However, since we do not distinguish between reading and writing in the code, we
+    // always return true and offload the handling to the user.
     if (meshConnectivity)
+    {
         return true;
+    }
     else
+    {
         return (this->locationType_ == LocationType::faceCenters);
+    }
 }
 
 std::string preciceAdapter::CHT::HeatFlux::getDataName() const
--- a/CouplingDataUser.C
+++ b/CouplingDataUser.C
@@ -31,9 +31,27 @@ void preciceAdapter::CouplingDataUser::setPatchIDs(std::vector<int> patchIDs)
     patchIDs_ = patchIDs;
 }
 
-void preciceAdapter::CouplingDataUser::setLocationsType(std::string locationsType)
+void preciceAdapter::CouplingDataUser::setLocationsType(LocationType locationsType)
 {
     locationsType_ = locationsType;
+}
+
+void preciceAdapter::CouplingDataUser::checkDataLocation(const bool meshConnectivity)
+{
+    if (this->isLocationTypeSupported(meshConnectivity) == false)
+    {
+        std::string location("none");
+        if (locationsType_ == LocationType::faceCenters)
+            location = "faceCenters";
+        else if (locationsType_ == LocationType::faceNodes)
+            location = "faceNodes";
+
+        FatalErrorInFunction
+            << "ERROR: the configured location type "
+            << location
+            << " is not supported for the data "
+            << exit(Foam::FatalError);
+    }
 }
 
 // Dummy implementation which can be overwritten in derived classes if required
--- a/CouplingDataUser.C
+++ b/CouplingDataUser.C
@@ -33,24 +33,24 @@ void preciceAdapter::CouplingDataUser::setPatchIDs(std::vector<int> patchIDs)
 
 void preciceAdapter::CouplingDataUser::setLocationsType(LocationType locationsType)
 {
-    locationsType_ = locationsType;
+    locationType_ = locationsType;
 }
 
-void preciceAdapter::CouplingDataUser::checkDataLocation(const bool meshConnectivity)
+void preciceAdapter::CouplingDataUser::checkDataLocation(const bool meshConnectivity) const
 {
-    if (this->isLocationTypeSupported(meshConnectivity) == false)
+    if (isLocationTypeSupported(meshConnectivity) == false)
     {
         std::string location("none");
-        if (locationsType_ == LocationType::faceCenters)
+        if (locationType_ == LocationType::faceCenters)
             location = "faceCenters";
-        else if (locationsType_ == LocationType::faceNodes)
+        else if (locationType_ == LocationType::faceNodes)
             location = "faceNodes";
 
         FatalErrorInFunction
             << "ERROR: the configured location type "
             << location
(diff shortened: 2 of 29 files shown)


CouplingDataUser.H
```diff
@@ -49,7 +58,10 @@ public:
     void setPatchIDs(std::vector<int> patchIDs);
 
     //- Set the locations type of the interface
-    void setLocationsType(std::string locationsType);
+    void setLocationsType(LocationType locationsType);
+
+    // Check, if the data set supports the interface node location
```
MakisH, line 63: No comma before "if" in English:
```suggestion
    // Check if the dataset supports this interface nodes location
```

CouplingDataUser.H
```diff
@@ -49,7 +58,10 @@ public:
...
+    void checkDataLocation(const bool meshConnectivity);
```
MakisH, line 64: I guess we can also mark this method as `const`.

CouplingDataUser.H
```diff
@@ -60,6 +72,13 @@ public:
     //- Read the coupling data from the buffer
     virtual void read(double* dataBuffer, const unsigned int dim) = 0;
 
+    //- Given the meshConnectivity, return if the underlying loactionType of the
+    //- interface nodes is supported by the data set
+    virtual bool isLocationTypeSupported(const bool meshConnectivity) const = 0;
+
+    //- Get the name of the current data field
+    virtual std::string getDataName() const = 0;
```
MakisH, line 80: Is this related to this PR, or just a nice side addition?

CouplingDataUser.C
```diff
@@ -31,11 +31,32 @@ void preciceAdapter::CouplingDataUser::setPatchIDs(std::vector<int> patchIDs)
...
 
-void preciceAdapter::CouplingDataUser::setLocationsType(std::string locationsType)
+void preciceAdapter::CouplingDataUser::setLocationsType(LocationType locationsType)
 {
     locationsType_ = locationsType;
 }
 
+void preciceAdapter::CouplingDataUser::checkDataLocation(const bool meshConnectivity)
+{
+    if (this->isLocationTypeSupported(meshConnectivity) == false)
```
MakisH, line 41: Why do we need `this->` here? I guess I am overseeing something, but isn't

```suggestion
    if (!isLocationTypeSupported(meshConnectivity))
```
enough?

CouplingDataUser.C
```diff
@@ -31,11 +31,32 @@ void preciceAdapter::CouplingDataUser::setPatchIDs(std::vector<int> patchIDs)
...
+            location = "faceNodes";
+
+        FatalErrorInFunction
+            << "ERROR: the configured location type "
+            << location
+            << " is not supported for the data "
+            << this->getDataName()
+            << ". Please select a different location type or "
+            << "a different data set."
+            << exit(Foam::FatalError);
```
MakisH, line 56: Why not using our own error handling utility?

CHT/HeatFlux.C
```diff
@@ -103,6 +103,21 @@ void preciceAdapter::CHT::HeatFlux::read(double* buffer, const unsigned int dim)
     }
 }
 
+bool preciceAdapter::CHT::HeatFlux::isLocationTypeSupported(const bool meshConnectivity) const
+{
+    // We don't distinguish between reading and writing. Thus, we just return true
+    // in case there was connectivity defined
+    if (meshConnectivity)
+        return true;
```
MakisH, line 111: This sounds a bit complicated to me. Does this mean that we support:

- face centers if mesh connectivity is defined,
- face nodes if mesh connectivity is defined,
- face centers if mesh connectivity is not defined,

both for reading and writing? If not, does a later/previous error catch this?

What did we have before?

Some documentation would be helpful.

FSI/Displacement.C
```diff
@@ -23,7 +23,7 @@ void preciceAdapter::FSI::Displacement::initialize()
 {
     // Initialize appropriate objects for each interface patch, namely the volField and the interpolation object
     // this is only necessary for face based FSI
-    if (this->locationsType_ == "faceCenters" || this->locationsType_ == "faceCentres")
+    if (this->locationsType_ == LocationType::faceCenters)
```
MakisH, line 26: [This page](https://precice.org/adapter-openfoam-config.html#the-adapters-configuration-file) will need an update to remove the `faceCentres` :gb: 

(original motivation: OpenFOAM uses British English for all configuration and this could have been an unexpected issue with some users. But better simplify)

Interface.C
```diff
@@ -31,6 +30,24 @@ preciceAdapter::Interface::Interface(
...
+    {
+        locationsType_ = LocationType::faceNodes;
+    }
+    else
+    {
+        FatalErrorInFunction
+            << "ERROR: interface points location type "
+            << locationsType
+            << " is invalid."
+            << exit(FatalError);
```
MakisH, line 47: Why not using our own error wrapper here?

MakisH (commented):
Nice contribution! :+1:

I feel less and less comfortable with how complex the codebase has developed because of such error handling, but this is definitely a step in the right direction and will help users debug their configs. Hopefully one day we will have nicer architecture that will not require so many files to be updated for such a simple matter.

Here a few questions to consider.

CouplingDataUser.H
```diff
@@ -1,11 +1,20 @@
...
 namespace preciceAdapter
 {
+// A small enum to deal with the different locations of the
+// coupling interface
+enum class LocationType
+{
+    none,
+    faceCenters,
+    faceNodes
+};
```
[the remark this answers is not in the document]
DavidSCN: Yes, this was also my first thought. The problem is the include hierarchy since the `CouplingDataUser` is required in the `Interface` and the `locationType` is eventually stored in the `CouplingDataUser`. If you don't like it here, we probably need an additional header or another file to put it.

Interface.H
```diff
@@ -21,7 +21,7 @@ protected:
     std::string meshName_;
 
     //- Point locations type (faceCenters vs faceNodes)
-    std::string locationsType_;
+    LocationType locationsType_ = LocationType::none;
```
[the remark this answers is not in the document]
DavidSCN: Is there a difference?

CouplingDataUser.H
```diff
@@ -1,11 +1,20 @@
 #ifndef COUPLINGDATAUSER_H
 #define COUPLINGDATAUSER_H
 
+#include "Utilities.H"
```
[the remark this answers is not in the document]
DavidSCN: Required for the error handling. Although I agree, I need to switch to our own error handling.

CouplingDataUser.H
```diff
@@ -27,7 +36,7 @@ protected:
     int dataID_;
 
     //- location type of the interface
-    std::string locationsType_;
+    LocationType locationsType_ = LocationType::none;
```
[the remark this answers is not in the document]
DavidSCN: For the `enum`, the singular makes more sense to me and the variable was a legacy variable. I will change it to singular as well.

CouplingDataUser.H
DavidSCN: No it's related. Otherwise the user won't know the origin of the issue.

CouplingDataUser.C
DavidSCN: I removed the `this` pointer, but I will not remove the 'equal false' since this is much more readable than the `!`

CouplingDataUser.C
DavidSCN: Good hint.

FSI/Displacement.C
DavidSCN: No, it won't, both spellings still work. I just summarized them into a single enum option.

CHT/HeatFlux.C
DavidSCN: We support for writing

- face nodes if mesh connectivity is defined
- face centers if no connectivity is defined

for reading
- face centers

However, we do not distinguish between reading and writing. Thus the function is an `||` of these options.

CouplingDataUser.C
DavidSCN: I copied this actually from another function in the adapter.

CouplingDataUser.H
MakisH: Right, I did not see it in the error messages I looked at that moment, but it is in the error message of the `CouplingDataUser`.

Interface.C
```diff
@@ -31,6 +30,23 @@ preciceAdapter::Interface::Interface(
...
+    else if (locationsType == "faceNodes")
+    {
+        locationType_ = LocationType::faceNodes;
+    }
+    else
+    {
+        adapterInfo("Interface points location type \""
+                    "locations = "
+                        + locationsType + "\" is invalid.",
+                    "error");
```
MakisH, line 46: Did you try if this error is triggered now? Maybe we need the `error-deferred`, because this step happens during the configuration.

CHT/HeatFlux.C
DavidSCN: > (Does this also apply to all other modules?)

Unfortunately not,

Interface.C
DavidSCN: I tried both, and both worked IIRC, I will try once more.

Interface.H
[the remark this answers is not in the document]
MakisH: Not really, I just like the explicit initialization more, especially since it can save us from implicit conversions.
But the rest of the code mostly/only uses assignment anyway.

CHT/HeatFlux.C
MakisH: Then let's just write something like

```suggestion
    // For cases with mesh connectivity, we support:
    // - face centers, both for reading and writing
    // - face nodes, only for writing
    // However, since we do not distinguish between reading and writing in the code, we
    // always return true and offload the handling to the user.
    if (meshConnectivity)
        return true;
```

I still think that it is not clear what we are trying to say here.

I also have the feeling that we have a larger issue here, which we are not handling properly. Overall, we have a cross product of:

- module (at least CHT and FSI support some operations on some data on either centers or nodes)
- specific coupling data user (heat flux, displacement, force, ...)
- face centers, face nodes
- reading, writing

I understand that we are still catching some of the combinations which are not meant to work, but not all of them. This can create a false impression to the user that they can rely on error handling to choose a working combination.

This PR is definitely something we need, I am just a bit worried seeing the whole design shaking as we cannot clearly and completely define the checks in a simple way.

(by the way, we should have braces around these one-line branches, for maintainability)

Interface.H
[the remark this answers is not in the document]
DavidSCN: Done

CHT/HeatFlux.C
DavidSCN: Ok, I will add a correct version of your comment in all data modules where we support mesh connectivity.

MakisH (approved):
With the footnote of https://github.com/precice/openfoam-adapter/issues/208, this PR looks good already!

Issue #55, opened 2018-12-20
Redesign checkpointing for any type


MakisH:
Due to the way the project evolved, checkpointing takes a very large portion of the code, which could be replaced by templates. We should try to redesign this, even if we have probably added every possible type to be checkpointed.

MakisH:
I believe that we can consider this closed with the macro introduced in #209.

PR #209, opened 2021-12-11
Various adjustments

olesenm:
Hi @MakisH - just gave a test compilation with the current OpenFOAM develop branch (mostly wanted to see if we made any changes that broke compilation for you). Needed a few changes for clang (ambiguity with tmp -> Field etc).

While I was looking at the code, noticed a few places with copy instead of reference (copying the pressure field is the nastiest one).

Cleaned up some of the handling of adding in checkpoint fields to remove large chunks of near-duplication code. The bit with `getObjectPtr` may or may not be what you want (depending on which versions of OpenFOAM you are targeting). Can replace with the equivalent:
```
mesh_.db().getObjectPtr<volScalarField>(...) ->
&(const_cast<volScalarField>(mesh_.db().lookupObject<volScalarField(...))
```
A bit more typing, but it's wrapped in a macro anyhow.
Either way, for this type of code I think it is better to be passing through the pointer and let the called method determine its validity. This makes it easier to wrap up a find/add logic together.


preciceAdapterFunctionObject.H
```diff
@@ -84,11 +84,11 @@ class preciceAdapterFunctionObject
 
     // Private Member Functions
 
-    //- Disallow default bitwise copy construct
-    preciceAdapterFunctionObject(const preciceAdapterFunctionObject&);
+    //- No copy copy construct
```
DavidSCN, line 87:
```suggestion
    //- No copy construct
```

Adapter.C
```diff
@@ -831,411 +831,167 @@ void preciceAdapter::Adapter::setupMeshVolCheckpointing()
...
-        {
-            adapterInfo("Could not checkpoint " + objectNames_[i], "warning");
-        }
+#undef doLocalCode
+#define doLocalCode(GeomField)                                           \
+    /* Checkpoint registered GeomField objects */                        \
+    for (const word& obj : mesh_.sortedNames<GeomField>())               \
+    {                                                                    \
+        addCheckpointField(mesh_.thisDb().getObjectPtr<GeomField>(obj)); \
+        DEBUG(adapterInfo("Checkpoint " + obj + " : " #GeomField));      \
```
DavidSCN, line 843: Without a request to change this: Is there a specific reason that you prefer a macro over a template?

DavidSCN (commented):
Awesome, thanks for this great contribution! My personal understanding of `tmp` in the OpeFOAM context was never good enough in order to apply it appropriately and I thought `auto` might allow me to get away with it, which seems to be wrong.

We had a few bug fixes within the last weeks and are now waiting for the upcoming OpenFOAM release in order to release the adapter alongside.

Now that you already touch the checkpointing allow me to ask a question: we have oftentimes the case that we reload the checkpoint after a single time step was computed. Does OpenFOAM store fields of the previous time step automatically? If it does, we could access these 'old' fields in order to reload the checkpointing instead of storing them explicitly and reloading them later on (would need to reload the old fields, though). This could save us a considerable amount of memory (namely all global fields at least once). I never really figured this out.

Adapter.C
olesenm: This indeed could be rewritten as a (static) templated private method taking three arguments (mesh db, vector of pointers, vector of pointers for the copy).
Could also unroll the methods and online them within the macro.
The main reason for the macro approach is to minimise the changes to the code and also leave the individual methods as-is in case there is a need to do something more than the push_back that it currently does. Or if surface fields need different treatment etc.

olesenm:
> (namely all global fields at least once). I never really figured this out.

Can't answer immediately, need a closer look at how/where it is being used. I suspect will also need a closer look at the registry flags to see what the adapter has done.

MakisH (commented):
Thanks a lot for the contributions and for the amazing care of checking for breaking changes! I have again a very good feeling about this OpenFOAM release.

Nothing of these changes should not work for past versions of OpenFOAM, right? Everything looks familiar.

By the way, a large part of this PR closes #55.

Adapter.C
```diff
@@ -831,411 +831,167 @@ void preciceAdapter::Adapter::setupMeshVolCheckpointing()
 
 void preciceAdapter::Adapter::setupCheckpointing()
 {
-    // Add fields in the checkpointing list
-    DEBUG(adapterInfo("Creating a list of checkpointed fields..."));
+    // Add fields in the checkpointing list - sorted for parallel consistency
```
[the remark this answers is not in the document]
olesenm: Just want to ensure that your listed fields ("abc", "def") always appear in exactly the same order on all processors, independent of the order in which they were originally inserted into the objectRegistry. In general, there should not be any differences between any of the processors, if there are any hash collisions they will have been resolved with the same chaining. However, for the small cost of a sort we can ensure that we don't not have to rely on hope and hashing functions (paranoid is best).

Adapter.C
olesenm: upper or lowercase doesn't really matter. In recent times I've abandoned proper names and just gone with 'doLocalCode', which describes exactly what it is. The advantage of this one is that we have have the undef/define/undef all in local scope and don't have to ever worry about an accidental undef on a meaningful macro name.

Templates are also fine, but they would add a file-scope location function.  In this case however, you may still want to wrap with a macro.  There are probably other places to fix first.

Adapter.C
```diff
@@ -831,411 +831,167 @@ void preciceAdapter::Adapter::setupMeshVolCheckpointing()
...
 {
-    surfaceScalarField* copy = new surfaceScalarField(field);
-    meshSurfaceScalarFields_.push_back(&field);
-    meshSurfaceScalarFieldCopies_.push_back(copy);
-    return;
+    {
+        meshSurfaceScalarFields_.push_back(&field);
+        meshSurfaceScalarFieldCopies_.push_back(new surfaceScalarField(field));
+    }
 }
```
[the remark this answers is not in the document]
olesenm: Didn't check everything, but should probably work ok with at least 1912 I think.

Adapter.C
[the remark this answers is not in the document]
MakisH: I see. The order indeed matters as it needs to be the same when we write and read multiple fields.

Adapter.C
DavidSCN: I would keep it as suggested here, considering that the same coding style is used throughout the OpenFOAM code base in order to keep things consistent.

DavidSCN:
I also resolved the merge conflict..

Adapter.C
MakisH: Then I also fully agree. The name was just not very clear to me on the first sight.

MakisH (approved):

DavidSCN:
> Can't answer immediately, need a closer look at how/where it is being used. I suspect will also need a closer look at the registry flags to see what the adapter has done.

Merging here, but want still want to mention that I'm interest in this.

olesenm:
> Merging here, but want still want to mention that I'm interest in this.

Sure, but deserves a separate issue.

DavidSCN:
Opened in #210

MakisH:
An update on the compatibility with other versions:
- All recent .com version (v1912-v2112) built and ran successfully in our new GitHub Actions workflow to build with any version.
- All .org versions (4-8) complained at the exact point that @olesenm mentioned above

In the OpenFOAM4 branch (on which every other .org branch is based upon), I applied the following patch to make it work:

```diff
-    for (const word& obj : mesh_.sortedNames<GeomField>())               \
-    {                                                                    \
-        addCheckpointField(mesh_.thisDb().getObjectPtr<GeomField>(obj)); \
+    for (const word& obj : mesh_.lookupClass<GeomField>().toc())                                    \
+    {                                                                                               \
+        addCheckpointField(&(const_cast<GeomField&>(mesh_.thisDb().lookupObject<GeomField>(obj)))); \

```

Note that I reverted the `sortedNames` back to the `toc`, as I could not find a templated version of `sortedNames` in OpenFOAM4.

I hope I did not overlook/misuse anything here. I sometimes get confused in these long chains of lookups and type conversions (loving `auto` and C++17 automatic type deduction in templates).

olesenm:
Should at least use sortedToc() instead of toc() for overall consistency

MakisH:
> @MakisH Should at least use sortedToc() instead of toc() for overall consistency

Good point, thank you. I applied this patch and rebased again.

PR #214, opened 2022-01-28
Add a custom build workflow

MakisH:
This adds a manually-triggered GitHub Actions workflow that can build:
- any branch of the adapter (but not from external pull requests)
- with any of the OpenFOAM versions available on Ubuntu 20.04 or 18.04 (default)
- with any released version of preCICE (that provides Debian packages)
- with or without running the tutorials flow-over-heated-plate and/or quickstart

The workflow uploads the adapter build logs (`Allwmake.log`, `wmake.log`, and `ldd.log`), the tutorial logs, and the complete case directories to the build artifacts.

The idea is that don't have each version-specific branch run a separate workflow (which would require fixing some conflicts automatically), but we manually prepare any update branches and test them whenever needed. I would mainly like to have this to ensure backwards compatibility with older .com versions.

I have [tested it on my fork](https://github.com/MakisH/openfoam-adapter/actions/workflows/build-custom.yml).

I don't use caches to reduce complexity. In any case, we should not run this frequently enough for caches to become relevant.

Unfortunately, we cannot change the name of the workflow based on the inputs (this is a [commonly requested feature](https://github.community/t/github-actions-dynamic-name-of-the-workflow-with-workflow-dispatch/150327)). To account for that, I included a "report" in the beginning of the run.

Example screenshots:

Screenshot from 2022-01-28 10-56-26

Screenshot from 2022-01-28 10-58-06

Screenshot from 2022-01-28 10-58-31

@DavidSCN @uekerman Do you think we should have any other options or report anything else additionally? Do you notice anything strange in the diff?

TODO list:

- [x] I updated the documentation in `docs/` -> should go to the contributor guidelines
- [x] I added a changelog entry in `changelog-entries/` (create directory if missing) -> added directly in the prepared file from develop


DavidSCN (commented):
I think this is a good effort in order to have a quick check for compatibility. I'm not sure if it is enough for releasing individual branches/versions, though. Maybe this is then solved by the systemtests in the future at some point.

MakisH:
> I think this is a good effort in order to have a quick check for compatibility. I'm not sure if it is enough for releasing individual branches/versions, though. Maybe this is then solved by the systemtests in the future at some point.

@DavidSCN thank you for the appreciation and the feedback. It looks like running some tutorials was not so complicated, since I had already a few things figured out. I added more features, including running the tutorials and archiving the logs. I also reduced duplication.

I will rebase on develop, update directly the changelog (since we have already prepared it), and add some documentation.

Anything else?

DavidSCN:
> Anything else?

No, looks great, thank you. I'm not sure about the compatibility with upcoming systemtests, but you probably know it, so please go ahead.

MakisH (commented):
I went through the diff once again and it looks fine to me. It has also rested for a few days.

I just now switched the default to Ubuntu 18.04 and added options for OpenFOAM 5 and 6. I have tested a few potential edge cases.

The most important is to get this workflow in the default branch, so that it appears in the Actions menu. After that, we can select an updated workflow from any branch.

PR #215, opened 2022-02-01
Disable automatic links checking (make manual)

MakisH:
We currently have an automatic check for unreachable links. This can turn a bit annoying, as it most often fails in irrelevant PRs.

This converts it to a manual workflow. I am not making it a cron job as it will again associate its status with a specific commit, which would be misleading.

TODO list:

- [x] I updated the documentation in `docs/` -> Added in `CONTRIBUTING.md`
- [x] I added a changelog entry in `changelog-entries/` (create directory if missing) -> Added in the already prepared changelog.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -1,0 +1,14 @@
+name: Check links (manual)
+on: workflow_dispatch
+jobs:
+  check_links:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


MakisH (commented):
This is a very short and rather non-critical change. The diff looks as intended. Branches are as intended. All TODOs implemented.

PR #216, opened 2022-02-03
Add pull request template, document versioning

### LINKED ISSUE #178: Need a release workflow
We closed #170 with the argument that we can just have a reminder in our release workflow to update the branches (with simple merging, to not get too many trivial pull requests).

In this release workflow we would need notes to:

- [ ] Bump the version message n in `Adapter.C`
- [ ] Merge the changelog entries
- [ ] Prepare release notes
- [ ] Check the documentation for any missing updates
- [ ] Merge PR of `develop` to `master` (simple merge, not squash & merge)
- [ ] Merge `master` back to `develop` and change the version message.
- [ ] Update the version-specific OpenFOAM branches

What else would we need?

We could make this a PR template, but in our experience with precice/precice, GitHub does not let you select the template when you open a pull request. Instead, we could add this as a Markdown file in `tools/`.

MakisH:
This PR:

- adds a release pull request template in `tools/` (meant to just be copied in the PR description manually). Closes #178.
- explains the version numbers in the "Get the OpenFOAM adapter" page
- clarifies a few minor things in the OpenFOAM support page
- changes the version from the (originally planned) `1.0.1` to the now agreed `1.1.0`.

TODO list:

@DavidSCN have a look on the template and versioning strategy. Do we still need to clarify anything else?

### DIFF
--- a/tools/release_pull_request_template.md
+++ b/tools/release_pull_request_template.md
@@ -1,0 +1,72 @@
+# Release checklist
+
+Copy this template to the release pull request description.
+
+## Update workflows
+
+- [ ] `.github/workflows/build.yml` and `.github/workflows/install-dependencies.sh`: Update the OpenFOAM and preCICE versions, if needed.
+- [ ] `build-custom.yml`: Add any new OpenFOAM versions and update the default preCICE version, if needed.
+
+## Prepare branches
+
+- [ ] Merge `master` to `develop`, if needed
+- [ ] Open a pull request from `develop` to `master`.
+  - We don't use release branches, unless there are changes we don't want to release, potentially when preparing a major release.
+- [ ] Run and fix all workflows
+
+## Test
+
+- [ ] Run and check flow-over-heated-plate OpenFOAM-OpenFOAM
+- [ ] Run and check quickstart
+- [ ] Run and check perpendicular-flap OpenFOAM-deal.II
+- [ ] Run and check elastic-tube-3d OpenFOAM-CalculiX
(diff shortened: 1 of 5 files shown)


tools/release_pull_request_template.md
```diff
@@ -0,0 +1,79 @@
...
+  - We don't use release branches, unless there are changes we don't want to release, potentially when preparing a major release.
+- [ ] Run and fix all workflows
+
+## Test
+
+- [ ] Run and check flow-over-heated-plate OpenFOAM-OpenFOAM
+- [ ] Run and check quickstart
+- [ ] Run and check perpendicular-flap OpenFOAM-deal.II
+- [ ] Run and check elastic-tube-3d OpenFOAM-CalculiX
+- [ ] Run and check partitioned-pipe sonicLiquidFoam-sonicLiquidFoam
```
DavidSCN, line 23: How do I check cases without reference results?

DavidSCN (approved):

tools/release_pull_request_template.md
MakisH: These are meant to be regressions checks. In the future, we will have the system tests comparing the interface VTK files. For now, most problems would appear just when running the simulation. If the simulations does not start, crashes, or the results look qualitatively different, then there is definitely a problem. If this is ok, I would try to compare some basic results to the previous release (depending on the case), especially if there are significant changes in the adapter.

I also think it is a good idea to always add a detailed comment with what I checked and what I observed, as well as mention what I did not check.

uekerman (approved):
Checklist looks very clean :+1:

PR #218, opened 2022-02-06
Add a hint to cite the adapter

MakisH:
This adds the following hint at the end of each simulation:

Screenshot from 2022-02-06 19-50-36

The motivation is that sometimes, users cite the preCICE library, but not the adapters they are using. This is a suggestion that we could port also to other adapters. @uekerman 

A few notes:
- I want this to appear in the end, because information in the beginning is often missing.
- I added this in the `!isCouplingOngoing()` branch and not in the destructor, so that it only appears at the end of a (hopfully) successful simulation, not next to wild adapter or OpenFOAM errors.
- The green color highlights the message, but also serves as an implicit "success" sign at the end. I hope it is not too aggressive. We do use color codes in other places as well, this is the only case where we use green.
- I point to the website, as we want to be able to update this in the future. The version is already anyway reported in the beginning.

@DavidSCN do you agree? Any further suggestions? Let's integrate this in this release, as users take a while to upgrade.

TODO list:

- [x] I updated the documentation in `docs/` -> nothing needed
- [x] I added a changelog entry in `changelog-entries/` (create directory if missing) -> directly in the prepared changelog

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -468,11 +468,9 @@ void preciceAdapter::Adapter::execute()
             "The simulation was ended by preCICE. "
             "Calling the end() methods of any functionObject explicitly.",
             "info");
-        adapterInfo("\033[32m" // green color 
-                    "Great that you are using the OpenFOAM-preCICE adapter! "
+        adapterInfo("Great that you are using the OpenFOAM-preCICE adapter! "
                     "Next to the preCICE library and any other components, please also cite this adapter. "
-                    "Find how on https://precice.org/adapter-openfoam-overview.html."
-                    "\033[0m", // restore color
+                    "Find how on https://precice.org/adapter-openfoam-overview.html.",
                     "info");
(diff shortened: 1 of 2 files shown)


DavidSCN (commented):
I have no idea whether it is good practice to include something in this direction. In case I start developing something (as a user) and run a particular case on and on I could imagine that the green message is rather annoying.

MakisH:
I removed the color. I guess this is also good enough.

uekerman (approved):
It might be annoying to see this over and over again. But let's give it a shot and ask for feedback eventually. Good that you removed the green color.

MakisH:
> It might be annoying to see this over and over again. But let's give it a shot and ask for feedback eventually. Good that you removed the green color.

Well... There is so much potentially redundant output that we see over and over again, I am not sure if this is too much to ask for free and academic software. Now without the color, it should be easily ignorable after the first encounter.

PR #220, opened 2022-02-17
Extend Allwmake suggestions for possible problems

MakisH:
This extend the `Allwmake` output to propose more ideas when something is going wrong during building. Inspired by recent topics on Discourse.

@DavidSCN do you think we should add anything else here?

TODO list:

- [x] I updated the documentation in `docs/` -> N/A
- [x] I added a changelog entry in `changelog-entries/` (create directory if missing) ~~-> Not sure if we really need one.~~ added

### DIFF
--- a/Allwmake
+++ b/Allwmake
@@ -54,7 +54,7 @@ if command -v pkg-config >/dev/null 2>&1; then
         Check the preCICE documentation page 'Linking to preCICE' for more details, or proceed if you know what you are doing."
     fi
 else
-    warning "pkg-config is not installed and is used to detect the preCICE installation. Only proceed if you know what you are doing."
+    warning "pkg-config is not installed and cannot be used to detect the preCICE installation. Only proceed if you know what you are doing."
 fi
 export ADAPTER_PKG_CONFIG_CFLAGS
 export ADAPTER_PKG_CONFIG_LIBS
--- a/Allwmake
+++ b/Allwmake
@@ -54,7 +54,7 @@ if command -v pkg-config >/dev/null 2>&1; then
         Check the preCICE documentation page 'Linking to preCICE' for more details, or proceed if you know what you are doing."
     fi
 else
-    warning "pkg-config is not installed and cannot be used to detect the preCICE installation. Only proceed if you know what you are doing."
+    warning "pkg-config is not installed and cannot be used to detect the preCICE installation. If building succeeds, ignore this warning."
 fi
 export ADAPTER_PKG_CONFIG_CFLAGS
 export ADAPTER_PKG_CONFIG_LIBS
--- a/Allwmake
+++ b/Allwmake
@@ -24,6 +24,11 @@ DOC_COMPATIBILITY="https://www.precice.org/adapter-openfoam-support.html"
 log() {
     echo "$@" | tee -a "Allwmake.log"
 }
+# Function to print warnings
+warning() {
+    log "WARNING: $*"
+    export ADAPTER_WARNINGS="true"
+}
 ################################################################################
 
 # Information header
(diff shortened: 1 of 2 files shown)


Allwmake
```diff
@@ -97,8 +98,11 @@ else
     if [ "${ADAPTER_WMAKE_UNDEFINED_SYMBOLS}" -gt 0 ] || [ "${ADAPTER_LD_UNDEFINED_SYMBOLS}" -gt 0 ]; then
         log "=== ERROR: Building completed with linking problems: there were undefined symbols. ==="
         log "Possible causes:"
-        log "- Is preCICE discoverable at runtime? Check the content of ADAPTER_PKG_CONFIG_LIBS above."
+        log "- Is preCICE discoverable at runtime? Check the content of pkg-config output above."
```
DavidSCN, line 101: I think users recently also faced the issue of pkg-config not being installed. Since the script relies on pkg-config, we may want to add a safety check for this as well. Also, what is the pkg-config output supposed to be? How can I fix it in case it doesn't show the right thing?

DavidSCN:
A bit sag that it will take quite a while until these usability changes reach the user.

Allwmake
MakisH: Good point about not having pkg-config installed. I will try to catch this (rare) case in the same PR.

This is supposed to be the pkg-config output, should I refer to it in some other way?

```
log "If not already known by the system, preCICE may be located using:"
log "  pkg-config --cflags libprecice  = ${ADAPTER_PKG_CONFIG_CFLAGS}"
log "  pkg-config --libs libprecice    = ${ADAPTER_PKG_CONFIG_LIBS}"
```

MakisH: no, this is set automatically and is not in the head of the script. Here I just copied the log part of the script, normally you would get paths and options there (at least a `-lprecice`).

MakisH: I added a check if pkg-config is installed and I throw a warning, if not. In principle, `pkg-config` is not an always required dependency, as there are also other ways to make preCICE discoverable.

I checked this by moving `/usr/bin/pkg-config`. This is the log I get:

```
Building the OpenFOAM-preCICE adapter...
WARNING: pkg-config configuration file 'libprecice.pc' was not found in the PKG_CONFIG_PATH.
        Check the preCICE documentation page 'Linking to preCICE' for more details, or proceed if you know what you are doing.

Current OpenFOAM environment:
  WM_PROJECT = OpenFOAM
  WM_PROJECT_VERSION = v2112

The adapter will be built into /home/makish/OpenFOAM/makish-v2112/platforms/linux64GccDPInt32Opt/lib
Additional preprocessor/compiler options: 

Building with WMake (see the wmake.log log file)...

Compiling enabled on 4 cores
wmake libso (openfoam-adapter)
Everything looks fine in wmake.log.
Everything looks fine in ldd.log.
=== OK: Building completed successfully! ===
There have been warnings. Check the log above.
```

I also implemented a new function for warnings, just to reduce duplication. I could also do this for errors, but that's already getting into a different direction.

Allwmake
```diff
@@ -33,23 +38,27 @@ log "Building the OpenFOAM-preCICE adapter..."
...
+    if pkg-config libprecice; then
+        ADAPTER_PKG_CONFIG_CFLAGS="$(pkg-config --silence-errors --cflags libprecice)"
+        ADAPTER_PKG_CONFIG_LIBS="$(pkg-config --silence-errors --libs libprecice)"
+        log ""
+        log "If not already known by the system, preCICE may be located using:"
+        log "  pkg-config --cflags libprecice  = ${ADAPTER_PKG_CONFIG_CFLAGS}"
+        log "  pkg-config --libs libprecice    = ${ADAPTER_PKG_CONFIG_LIBS}"
+    else
+        warning "pkg-config configuration file 'libprecice.pc' was not found in the PKG_CONFIG_PATH.
+        Check the preCICE documentation page 'Linking to preCICE' for more details, or proceed if you know what you are doing."
```
DavidSCN, line 54: I like the new output. Only the last part 'or proceed if you know what you are doing' fits not so well in my opinion as there are only two possible scenarios: building succeeds -> should I proceed now? building  fails -> I can't really proceed.

Allwmake
```diff
@@ -33,23 +38,27 @@ log "Building the OpenFOAM-preCICE adapter..."
...
+    fi
 else
-  log "WARNING: No configuration file 'libprecice.pc' was found in the PKG_CONFIG_PATH."
-  log "Check the preCICE documentation page 'Linking to preCICE' for more details, or proceed if you know what you are doing."
-  ADAPTER_PKG_CONFIG_CFLAGS=""
-  ADAPTER_PKG_CONFIG_LIBS=""
+    warning "pkg-config is not installed and is used to detect the preCICE installation. Only proceed if you know what you are doing."
```
DavidSCN, line 57:
```suggestion
    warning "pkg-config is not installed and cannot be used to detect the preCICE installation. Only proceed if you know what you are doing."
```

and the same as above regarding 'proceeding'.

Allwmake
MakisH: Good point. I had in mind that the system may know about preCICE even without pkg-config, but this is not really helpful for the user. I replaced it with "If building succeeds, ignore this warning.".

Allwmake
MakisH: @davidscn I have addressed all your comments in this conversation:
- This works both for users with and without pkg-config, as well as with discoverable and non-discoverable `libprecice.pc`.
- If any of these is not found, the user gets a warning that essentially says "it's fine if building works".

Please check and approve, if ok.

davidscn: Looks good now.

PR #224, opened 2022-03-11
Implement restart from reference configuration

### LINKED ISSUE #223: Coupling data jump in restarted FSI simulation
Please provide the following information:
- Adapter version: v1.1.0
- OpenFOAM version: OpenFOAM 2012
- preCICE version: v2.3.0

I'm currently experimenting with restart functionalities by simulating two seconds of our perpendicular flap tutorial and restarting it again: first I simulate the complete two seconds as a reference and exporting the vtks, afterwards, I simulate the first second and restart it later on using the replay mode of aste on the solid side. Looking at the `Stress` data results in the following 
   

There is an unreasonable jump at the watchpoint I don't understand. Restarting the same simulation using a real (the solid) participant results in a different simulation.

davidscn:
Closes #223 

With this, we can restart the 'initial' simulation by resetting the deformation accumulated throughout the simulation. It would be important to check how other (structural) codes handle the restart: If they define also the 'deformed' interface mesh when restarting, the reset here would be undesired. Do we have solver which behave this way? I only checked deal.II for now, and the functionality used there is entirely based on the initial configuration.

TODO list:

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -110,6 +110,10 @@ bool preciceAdapter::Adapter::configFileRead()
                     }
                     DEBUG(adapterInfo("    connectivity : " + std::to_string(interfaceConfig.meshConnectivity)));
 
+                    // Reset the displacement when defining the interface
+                    interfaceConfig.resetDisplacement = interfaceDict.lookupOrDefault<bool>("resetDisplacement", true);
+                    DEBUG(adapterInfo("    reset displacement : " + std::to_string(interfaceConfig.resetDisplacement)));
+
                     DEBUG(adapterInfo("    patches      : "));
                     auto patches = interfaceDict.get<wordList>("patches");
                     for (auto patch : patches)
@@ -249,7 +253,7 @@ void preciceAdapter::Adapter::configure()
         DEBUG(adapterInfo("Creating interfaces..."));
         for (uint i = 0; i < interfacesConfig_.size(); i++)
         {
-            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity);
+            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, interfacesConfig_.at(i).resetDisplacement);
             interfaces_.push_back(interface);
             DEBUG(adapterInfo("Interface created on mesh " + interfacesConfig_.at(i).meshName));
 
--- a/Adapter.C
+++ b/Adapter.C
@@ -114,8 +114,8 @@ bool preciceAdapter::Adapter::configFileRead()
                     // We do this by resetting the displacement when defining the interface.
                     // Since this is a feature that may not work as expected, depending on the implementation of the
                     // structure solver used, we make this feature opt-in.
-                    interfaceConfig.restoreUndeformedInterface = interfaceDict.lookupOrDefault<bool>("restoreUndeformedInterface", false);
-                    DEBUG(adapterInfo("    restore undeformed interface : " + std::to_string(interfaceConfig.restoreUndeformedInterface)));
+                    interfaceConfig.restartFromDeformed = interfaceDict.lookupOrDefault<bool>("restartFromDeformed", true);
+                    DEBUG(adapterInfo("    restart from deformed : " + std::to_string(interfaceConfig.restartFromDeformed)));
 
                     DEBUG(adapterInfo("    patches      : "));
                     auto patches = interfaceDict.get<wordList>("patches");
@@ -259,7 +259,7 @@ void preciceAdapter::Adapter::configure()
             std::string namePointDisplacement = FSIenabled_ ? FSI_->getPointDisplacementFieldName() : "default";
             std::string nameCellDisplacement = FSIenabled_ ? FSI_->getCellDisplacementFieldName() : "default";
 
-            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, interfacesConfig_.at(i).restoreUndeformedInterface, namePointDisplacement, nameCellDisplacement);
+            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, interfacesConfig_.at(i).restartFromDeformed, namePointDisplacement, nameCellDisplacement);
             interfaces_.push_back(interface);
             DEBUG(adapterInfo("Interface created on mesh " + interfacesConfig_.at(i).meshName));
 
--- a/Adapter.C
+++ b/Adapter.C
@@ -113,7 +113,7 @@ bool preciceAdapter::Adapter::configFileRead()
                     // When restarting FSI simulations, we may need to account for previous displacement.
                     // We do this by resetting the displacement when defining the interface.
                     // Since this is a feature that may not work as expected, depending on the implementation of the
-                    // structure solver used, we allow the user to enable this feature.
+                    // structure solver used, we make this feature opt-in.
                     interfaceConfig.resetDisplacement = interfaceDict.lookupOrDefault<bool>("resetDisplacement", false);
                     DEBUG(adapterInfo("    reset displacement : " + std::to_string(interfaceConfig.resetDisplacement)));
 
--- a/Adapter.C
+++ b/Adapter.C
@@ -251,7 +251,7 @@ void preciceAdapter::Adapter::configure()
         {
             std::string namePointDisplacement = FSIenabled_ ? FSI_->getPointDisplacementFieldName() : "default";
             std::string nameCellDisplacement = FSIenabled_ ? FSI_->getCellDisplacementFieldName() : "default";
-            bool restartFromDeformed = FSIenabled_ ? FSI_->restartFromDeformed() : false;
+            bool restartFromDeformed = FSIenabled_ ? FSI_->isRestartingFromDeformed() : false;
 
             Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, restartFromDeformed, namePointDisplacement, nameCellDisplacement);
             interfaces_.push_back(interface);
--- a/Adapter.C
+++ b/Adapter.C
@@ -114,8 +114,8 @@ bool preciceAdapter::Adapter::configFileRead()
                     // We do this by resetting the displacement when defining the interface.
                     // Since this is a feature that may not work as expected, depending on the implementation of the
                     // structure solver used, we make this feature opt-in.
-                    interfaceConfig.resetDisplacement = interfaceDict.lookupOrDefault<bool>("resetDisplacement", false);
-                    DEBUG(adapterInfo("    reset displacement : " + std::to_string(interfaceConfig.resetDisplacement)));
+                    interfaceConfig.restoreUndeformedInterface = interfaceDict.lookupOrDefault<bool>("restoreUndeformedInterface", false);
+                    DEBUG(adapterInfo("    restore undeformed interface : " + std::to_string(interfaceConfig.restoreUndeformedInterface)));
 
                     DEBUG(adapterInfo("    patches      : "));
                     auto patches = interfaceDict.get<wordList>("patches");
@@ -259,7 +259,7 @@ void preciceAdapter::Adapter::configure()
             std::string namePointDisplacement = FSIenabled_ ? FSI_->getPointDisplacementFieldName() : "default";
             std::string nameCellDisplacement = FSIenabled_ ? FSI_->getCellDisplacementFieldName() : "default";
 
-            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, interfacesConfig_.at(i).resetDisplacement, namePointDisplacement, nameCellDisplacement);
+            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, interfacesConfig_.at(i).restoreUndeformedInterface, namePointDisplacement, nameCellDisplacement);
             interfaces_.push_back(interface);
             DEBUG(adapterInfo("Interface created on mesh " + interfacesConfig_.at(i).meshName));
 
--- a/Adapter.C
+++ b/Adapter.C
@@ -111,7 +111,7 @@ bool preciceAdapter::Adapter::configFileRead()
                     DEBUG(adapterInfo("    connectivity : " + std::to_string(interfaceConfig.meshConnectivity)));
 
                     // Reset the displacement when defining the interface
-                    interfaceConfig.resetDisplacement = interfaceDict.lookupOrDefault<bool>("resetDisplacement", true);
+                    interfaceConfig.resetDisplacement = interfaceDict.lookupOrDefault<bool>("resetDisplacement", false);
                     DEBUG(adapterInfo("    reset displacement : " + std::to_string(interfaceConfig.resetDisplacement)));
 
                     DEBUG(adapterInfo("    patches      : "));
--- a/Adapter.C
+++ b/Adapter.C
@@ -253,7 +253,14 @@ void preciceAdapter::Adapter::configure()
         DEBUG(adapterInfo("Creating interfaces..."));
         for (uint i = 0; i < interfacesConfig_.size(); i++)
         {
-            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, interfacesConfig_.at(i).resetDisplacement);
+            std::string namePointDisplacement = "pointDisplacement";
+            std::string nameCellDisplacement = "cellDisplacement";
+            if (FSIenabled_)
+            {
+                namePointDisplacement = FSI_->namePointDisplacement;
+                nameCellDisplacement = FSI_->nameCellDisplacement;
+            }
+            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, interfacesConfig_.at(i).resetDisplacement, namePointDisplacement, nameCellDisplacement);
             interfaces_.push_back(interface);
             DEBUG(adapterInfo("Interface created on mesh " + interfacesConfig_.at(i).meshName));
 
--- a/Adapter.C
+++ b/Adapter.C
@@ -110,13 +110,6 @@ bool preciceAdapter::Adapter::configFileRead()
                     }
                     DEBUG(adapterInfo("    connectivity : " + std::to_string(interfaceConfig.meshConnectivity)));
 
-                    // When restarting FSI simulations, we may need to account for previous displacement.
-                    // We do this by resetting the displacement when defining the interface.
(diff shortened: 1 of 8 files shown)


uekerman:
> I only checked deal.II for now, and the functionality used there is entirely based on the initial configuration.

I would indeed expect this behavior from the restart of a CSM solver. Same mesh, "only" different displacement values.

AndresPedemonteFIUBA:
Hi,

> It would be important to check how other (structural) codes handle the restart: If they define also the 'deformed' interface mesh when restarting, the reset here would be undesired. Do we have solver which behave this way?

I ran a test using FOAM and CalculiX. I exported and visualized the initial meshes (init) after the restart; both were 'deformed' meshes. I realized it because my initial configuration (i.e. the initial run, not the restart) had a `y-dead=true` for the mappings which caused issues after the restart unless removed. This might therefore cause a conflict with CalculiX's restart feature.

DavidSCN:
@AndresPedemonteFIUBA, thanks for pointing this out, I know (cf also with the [source code comment in ccx](https://github.com/precice/calculix-adapter/blob/4496e8f160ec121481653baf44ec032813409ac6/adapter/CCXHelpers.c#L78) as pointed out by @boris-martin) and I'm not sure how to handle or change the behavior properly. Creating transparency and enabling the reference restart optional in a first step is probably a good idea.

Adapter.C
```diff
@@ -254,7 +254,14 @@ void preciceAdapter::Adapter::configure()
         DEBUG(adapterInfo("Creating interfaces..."));
         for (uint i = 0; i < interfacesConfig_.size(); i++)
         {
-            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, interfacesConfig_.at(i).resetDisplacement);
+            std::string namePointDisplacement = "pointDisplacement";
+            std::string nameCellDisplacement = "cellDisplacement";
```
MakisH, line 258: This is something we should either add in the adapter config or drop / hard-code. It does not really help to have the names here.

Adapter.C
```diff
@@ -110,6 +110,10 @@ bool preciceAdapter::Adapter::configFileRead()
                     }
                     DEBUG(adapterInfo("    connectivity : " + std::to_string(interfaceConfig.meshConnectivity)));
 
+                    // Reset the displacement when defining the interface
+                    interfaceConfig.resetDisplacement = interfaceDict.lookupOrDefault<bool>("resetDisplacement", true);
```
MakisH, line 114: This also needs to be documented in https://github.com/precice/openfoam-adapter/blob/develop/docs/config.md#advanced-configuration

Interface.C
```diff
@@ -88,6 +92,11 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
         }
         DEBUG(adapterInfo("Number of face centres: " + std::to_string(numDataLocations_)));
 
+        Foam::volVectorField const* cellDisplacement = nullptr;
+        if (mesh.foundObject<volVectorField>(nameCellDisplacement))
+            cellDisplacement =
+                &mesh.lookupObject<volVectorField>(nameCellDisplacement);
```
MakisH, line 98: Comment for the intention missing.

Adapter.C
```diff
@@ -110,6 +110,10 @@ bool preciceAdapter::Adapter::configFileRead()
                     }
                     DEBUG(adapterInfo("    connectivity : " + std::to_string(interfaceConfig.meshConnectivity)));
 
+                    // Reset the displacement when defining the interface
```
MakisH, line 113: The comment here is a bit confusing, as the resetting does not happen here. Maybe:

```suggestion
                    // When restarting FSI simulations, we need to account for previous accumulated displacement.
                    // We do this by resetting the displacement when defining the interface.
                    // Since this is a feature that may not work as expected, depending on the implementation of the
                    // structure solver used, we allow the user to disable this feature.
```

Interface.C
```diff
@@ -104,9 +113,12 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
         for (uint j = 0; j < patchIDs_.size(); j++)
         {
             // Get the face centers of the current patch
-            const vectorField faceCenters =
+            vectorField faceCenters =
                 mesh.boundaryMesh()[patchIDs_.at(j)].faceCentres();
 
+            if (cellDisplacement != nullptr && resetDisplacement_)
+                faceCenters -= cellDisplacement->boundaryField()[patchIDs_.at(j)];
```
MakisH, line 120: A comment for the intention is missing. Not only for relating this to restarting, but also to explain the mechanism.

I would expect something like (if I understand correctly):

```suggestion
            // Move the interface according to the current values of the cellDisplacement field,
            // to account for any displacements accumulated before restarting the simulation.
            // This is information that OpenFOAM reads from its result/restart files.
            // If the simulation is not restarted, the displacement should be zero and this line should have no effect.
            if (cellDisplacement != nullptr && resetDisplacement_)
                faceCenters -= cellDisplacement->boundaryField()[patchIDs_.at(j)];
```

MakisH (commented):
I had another look, and left a few more thoughts, mostly on improving the documentation/comments (the intention and mechanism is a bit unclear at the moment).

I am also wondering is `reset` is the right term here. I always think of "reset" as "going back to the default/zero state", while here we want to "continue the game from the saved point".

Maybe `restart`, `recover`, or something else would fit better?

MakisH:
Discussing with @davidscn: This is important, but changing the default could be surprising for users that already have adapted to the current behavior. Let's make it an optional feature (disabled by default) and reconsider in a future, breaking release.

Adapter.C
davidscn: I removed the hard-coded names.

Adapter.C
davidscn: added

docs/config.md
```diff
@@ -395,6 +395,22 @@ FSI
 
 Use the option `namePointDisplacement unused;` for solvers that do not create a pointDisplacement field, such as the RBFMeshMotionSolver.
 
+#### Restarting FSI simulations
+
+Restarting a coupled simulation using the OpenFOAM adapter works in principle in the same way as restarting a standalone OpenFOAM solver. However, the adapter and preCICE always redefine the coupling interface and the interface node locations. In case of FSI simulations, the interface deforms over time, which leads to the definition of a deformed interface during the restart. The boolean variable `resetDisplacement` (`false` by default) allows to account for the previously accumulated interface deformation such that the initial interface configuration (`t = 0`) is completely recovered. The setting here needs to coincide with the behavior of the selected solid solver.
```
MakisH, line 400: > However, the adapter and preCICE always redefine the coupling interface and the interface node locations.

What does "always" and "redefine" mean in this context? I am confused by this sentence.

> The setting here needs to coincide with the behavior of the selected solid solver.

I would use "agree" here instead of "coincide" (simpler word).

Adapter.C
```diff
@@ -110,6 +110,13 @@ bool preciceAdapter::Adapter::configFileRead()
                     }
                     DEBUG(adapterInfo("    connectivity : " + std::to_string(interfaceConfig.meshConnectivity)));
 
+                    // When restarting FSI simulations, we may need to account for previous displacement.
+                    // We do this by resetting the displacement when defining the interface.
+                    // Since this is a feature that may not work as expected, depending on the implementation of the
+                    // structure solver used, we allow the user to enable this feature.
```
MakisH, line 116: I think this better reflect what you want to say:

```suggestion
                    // structure solver used, we make this feature opt-in.
```

Adapter.C
```diff
@@ -249,7 +256,10 @@ void preciceAdapter::Adapter::configure()
         DEBUG(adapterInfo("Creating interfaces..."));
         for (uint i = 0; i < interfacesConfig_.size(); i++)
         {
-            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity);
+            std::string namePointDisplacement = FSIenabled_ ? FSI_->namePointDisplacement : "default";
+            std::string nameCellDisplacement = FSIenabled_ ? FSI_->nameCellDisplacement : "default";
+
+            Interface* interface = new Interface(*precice_, mesh_, interfacesConfig_.at(i).meshName, interfacesConfig_.at(i).locationsType, interfacesConfig_.at(i).patchNames, interfacesConfig_.at(i).meshConnectivity, interfacesConfig_.at(i).resetDisplacement, namePointDisplacement, nameCellDisplacement);
```
MakisH, line 262: Do we really need to pass these as arguments? We read them again in `FSI.C`, but I see that we now need them inside `Interface.C`. This starts getting too complicated (many of the changes in this PR are just for this), but I am not sure of any good solution at the moment.

I guess for now, we should either keep it as it is (but open an issue to fix the situation later), or keep the hard-coded names, documenting that restarting ignores these options. I think I would favor the hard-coded names for now (sorry for reverting your efforts).

What do you think?

Adapter.C
```diff
@@ -110,6 +110,13 @@ bool preciceAdapter::Adapter::configFileRead()
                     }
                     DEBUG(adapterInfo("    connectivity : " + std::to_string(interfaceConfig.meshConnectivity)));
 
+                    // When restarting FSI simulations, we may need to account for previous displacement.
+                    // We do this by resetting the displacement when defining the interface.
+                    // Since this is a feature that may not work as expected, depending on the implementation of the
+                    // structure solver used, we allow the user to enable this feature.
+                    interfaceConfig.resetDisplacement = interfaceDict.lookupOrDefault<bool>("resetDisplacement", false);
```
MakisH, line 117: A comment that I had previously made outside a threaded conversation:

I am also wondering if `reset` is the right term here. I always think of "reset" as "going back to the default/zero state", while here we want to "continue the game from the saved point".

Maybe `restart`, `recover`, or something else would fit better?

Adapter.C
davidscn: There are no hard-coded names here. If the user specifies names in the FSI part of the dict, they will be picked-up here. Am I missing something? The 'default' here is just to don't throw uninitialized memory around when the FSI module is disabled (and the option is irrelevant).

FSI/FSI.H
```diff
@@ -63,6 +57,12 @@ public:
 
     //- Add coupling data readers
     bool addReaders(std::string dataName, Interface* interface);
+
+    //- Name of the pointDisplacement field
+    std::string namePointDisplacement = "pointDisplacement";
+
+    //- Name of the pointDisplacement field
+    std::string nameCellDisplacement = "cellDisplacement";
```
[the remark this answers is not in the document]
davidscn: See above, alternative would be to add getter. I don't have a strong opinion on either option.

Adapter.C
davidscn: > going back to the default/zero state

Thinking in terms of the interface deformation, that's exactly what's happening here. I would maybe try to find a name that considers the 'restart' association.

Interface.C
```diff
@@ -196,9 +219,21 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+            // Similar to the cell displacement above:
+            // Move the interface according to the current values of the cellDisplacement field,
+            // to account for any displacements accumulated before restarting the simulation.
+            // This is information that OpenFOAM reads from its result/restart files.
+            // If the simulation is not restarted, the displacement should be zero and this line should have no effect.
+            if (pointDisplacement != nullptr && resetDisplacement_)
+            {
+                const vectorField& resetField = refCast<const vectorField>(
+                    pointDisplacement->boundaryField()[patchIDs_.at(j)]);
+                faceNodes -= resetField;
```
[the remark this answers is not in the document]
davidscn: One is for connectivity, one for the definition. Having the declaration twice narrows the definition scope considerably.

Adapter.C
MakisH: Indeed, this part is not reading anything from the config (I somehow made this assumption while reviewing everything together), it is only accessing the now public variable, which has already been defined at the configuration of the FSI module.

Looking at it again, I think the main aspect I don't like is that we are passing even more parameters via constructor arguments. In the context of the current state (see also mesh connectivity), I guess this is fine.

Interface.C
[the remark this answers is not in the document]
MakisH: I meant, why not:

```suggestion
                faceNodes -= pointDisplacement->boundaryField()[patchIDs_.at(j)];
```

FSI/FSI.H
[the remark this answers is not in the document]
MakisH: Adding a getter would, in my view, keep this consistent with the rest of the modules (and not need to change the names).

Adapter.C
davidscn: Yes, I was also briefly thinking about the constructor length while changing it here, but couldn't find a silver bullet solution right now.

Interface.C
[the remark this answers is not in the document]
davidscn: The compiler complains about your version, as there are apparently too many overloads for the operator (or similar). 

```cpp
faceNodes -= refCast<const vectorField>(pointDisplacement->boundaryField()[patchIDs_.at(j)]);
```
would compile, but I think the two-liner version above is more readable (at least for me).

docs/config.md
```diff
@@ -395,6 +395,22 @@ FSI
...
+```c++
+FSI
+{
+    // Account for previous displacement during a restart
+    resetDisplacement true;
+}
+```
+
+{% note %}
+The CalculiX adapter restarts simulations from the deformed interface mesh.
```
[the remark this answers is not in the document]
davidscn: added something

Adapter.C
davidscn: Renamed it to 'restoreUndeformedInterface', please comment again if you want it to be changed once more,

docs/config.md
davidscn: Updated.

docs/config.md
```diff
@@ -438,6 +438,22 @@ FF
...
+```c++
+FSI
+{
+    // Account for previous displacement during a restart
+    restoreUndeformedInterface true;
+}
+```
+
+{% note %}
+The CalculiX adapter restarts simulations from the deformed interface mesh, so that the default setting `restoreUndeformedInterface false` should be used.
```
MakisH, line 454: This sounds to me clearer and closer to the intended style: https://precice.org/docs-meta-overview.html#language--style

```suggestion
The CalculiX adapter restarts simulations from the deformed interface mesh. This means that you need to set `restoreUndeformedInterface false`.
```

Adapter.C
MakisH: I think that `restoreUndeformedInterface` is similarly confusing as `resetDisplacement`.

Since we want to communicate that this is (only) relevant for restarts, how about something like `restartFromDeformed`?

From the perspective of preCICE, it makes sense that we tell preCICE "these are the locations of the nodes that the simulation originally started from".

From the perspective of the user, however, the important information is the restarting part. To restart from a deformed case, we need to restore the original frame of reference.

Could we even make this a keyword-based option? Something like:
```c++
FSI{
    restartFrom zeroDisplacement;
}
```
with alternative:
```c++
FSI{
    restartFrom lastDisplacement;
}
```

We can then internally map it to the current `resetDisplacement` or `restoreUndeformedInterface`.

MakisH:
Alright, so, how do we test this?

davidscn:
You can just run a simulation and restart from a particular time stamp and export the precice interface. You don't need to execute the actual coupling to test this.

docs/config.md
```diff
@@ -438,6 +438,22 @@ FF
...
+#### Restarting FSI simulations
+
+Restarting a coupled simulation using the OpenFOAM adapter works in principle in the same way as restarting a standalone OpenFOAM solver. However, the adapter and preCICE define the coupling interface and the interface node locations based on the mesh at the particular time. In case of FSI simulations, the interface deforms over time, which leads to the definition of a deformed interface during the restart. The boolean variable `restartFromDeformed` (`true` by default) allows to account for the previously accumulated interface deformation such that the initial interface configuration (`t = 0`) is completely recovered. The setting here needs to agree with the behavior of the selected solid solver.
+
+```c++
+FSI
+{
+    // Account for previous displacement during a restart
+    restartFromDeformed true;
+}
```
MakisH, line 450: The current documentation confused me for a while:

```suggestion
Interface1
{
    mesh                Fluid-Mesh;
    patches             (flap);
    
    // Account for previous displacement during a restart
    restartFromDeformed true;
}
```

MakisH (commented):
> You can just run a simulation and restart from a particular time stamp and export the precice interface. You don't need to execute the actual coupling to test this.

If I just run a (single-physics) simulation, it needs to be a simulation that deforms the interface, right?

I tried running and restarting the perpendicular flap with OpenFOAM both for the fluid and the solid. I did that by:
- setting `max-time` to 0.5 in the `precice-config.xml`
- running the simulation
- changing the `system/controlDict` in both solvers to `startFrom latestTime`
- running again
- refreshing the results in ParaView
   - I loaded the fluid and solid results, applied a warp by vector (`D`) on the solid, grouped together, applied a slice, and visualized the mesh. It seems to be going smoothly, without returning to initial state or having any gaps.

I observed no differences, independent of the value of `restartFromDeformed` (even if I set a different value to each). This confuses me. I have not thought this through, but I would expect that the option has some effect.

I then tried coupling the same Fluid case (starting from `latestTime`) with CalculiX and restarting.
   - With `restartFromDeformed true;`, the interface returned to the zero state after 0.5.
   - With `restartFromDeformed false;`, I observed the same.
   
I then realized I should also add `initialize="true"` in the exchange tag of displacements. However, I got the same effect with both values of `restartFromDeformed`.

I noticed that the documentation was wrong and that I had to set the option at a different scope. However, even after fixing that (and checking that it was toggling in the debug output), I was still not able to observe a difference between these two options with neither OpenFOAM-CalculiX, nor OpenFOAM-OpenFOAM.

Could you please describe your exact testing procedure? I guess I am missing some step.

docs/config.md
davidscn: I moved the parameter now into the FSI dict (although it would fir conceptually better to the interface). I guess for the user, the FSI dict is the way to go.

davidscn:
Yes, sorry I had the option in the interface config instead of the FSI config, I changed it now to match the docs.

Also the changed default switched the logic, I also fixed that in the code now.

How to test:

1. run athe perpendicular flap completely
2. select `startTime = 0.7` (as it allows to see it quite clear)
3. run a few time-steps
4. open ParaView and select `Point Gaussian` visualization

Doing this with both options (enabled and disabled) should deliver different results.

docs/config.md
```diff
@@ -450,9 +450,9 @@ FSI
 }
 ```
 
-{% note %}
-The CalculiX adapter restarts simulations from the deformed interface mesh. This means that you need to set `restartFromDeformed true`.
-{% endnote %}
+{% important %}
+The option here defines the way the interface mesh is initialized when restarting an FSI simulation in OpenFOAM. In order to restart a coupled simulation, your solid solver needs to be capable of restarting mechanism as well. You can find more information about restarting coupled simulations on [Dsicourse](https://precice.discourse.group/t/how-can-i-restart-a-coupled-simulation/675).
```
MakisH, line 454:
```suggestion
The option here defines the way the interface mesh is initialized when restarting an FSI simulation in OpenFOAM. In order to restart a coupled simulation, your solid solver needs to be capable of restarting as well. Furthermore, the two participants need to follow the same assumption for the initialization, which for OpenFOAM you can configure with this option. You can find more information about restarting coupled simulations on [Dsicourse](https://precice.discourse.group/t/how-can-i-restart-a-coupled-simulation/675).
```

davidscn:
I added a note on solid solvers. If the solid solver operates only in reference coordinates, the option is a bit misleading.

FSI/FSI.C
```diff
@@ -210,3 +223,18 @@ bool preciceAdapter::FSI::FluidStructureInteraction::addReaders(std::string data
...
+{
+    return nameCellDisplacement_;
+}
+
+std::string preciceAdapter::FSI::FluidStructureInteraction::getPointDisplacementFieldName()
+{
+    return namePointDisplacement_;
+}
+
+bool preciceAdapter::FSI::FluidStructureInteraction::restartFromDeformed()
```
MakisH, line 237: Better name this `getRestartFromDeformed`, otherwise it sounds like an action.

MakisH (approved):
Thank you very much for updating the documentation and rethinking the names and defaults! ☺️ I am happy with the current state, feel free to merge when you feel ready.

As we discussed:
- It looks like our CalculiX case is not configured to restart (I don't know how to do that at the moment).
- Similarly, the deal.II example codes we have don't support a similar feature yet.

I watched @davidscn testing this at the preCICE level, and it seems to work as expected: the interface is set either at the zero or at the last known location, depending on the option. But we could not really test it with an actual example yet.

My tests with the OpenFOAM-OpenFOAM perpendicular flap case also looked like they were not affected by the option. I assume this is because OpenFOAM anyway does not move the points. Giving a deformed interface to preCICE still works, but probably only because the nearest neighbors for the mapping remain the same.

Since this does not change the default behavior, I don't have any further doubts or suggestions.

davidscn:
To document this: I disabled the option for solid solvers and made it only available for fluid solvers. All solid solvers I tested (solids4foam and solid-openfoam) use only the reference configuration. If one wants to restart a simulation using either of these solid solvers, one has to select `restartFromDeformed false` on the Fluid side and nothing on the solid side.

PR #226, opened 2022-03-14
Add website update

fsimonis:
This PR adds a workflow which triggers the website repo to update its submodules in case the documentation changes on develop.

Can we test this PR without merging it directly?


MakisH (approved):
This looks to me like a good addition and the code looks reasonable. I am not sure how to test it, but I also don't think we need to test this before merging: it should only trigger a workflow that we are always fine if it is triggered.

The `docs/**` pattern is enough.

PR #244, opened 2022-09-19
Remove -j 4 flag, document WM_NCOMPPROCS

MakisH:
This replaces/closes #219.

By default, the adapter does not take very long to compile (maybe a minute or two), and the typical user would only do that once. Instead of hard-coding the number of processes, or trying to run using all processes (both could lead to issues on some systems), this PR:

- Reports the current value of this variable:
   ```bash
   Current OpenFOAM environment:
     WM_PROJECT = OpenFOAM
     WM_PROJECT_VERSION = v2206
     WM_NCOMPPROCS = 4 (number of threads for parallel builds)
   ```
   If the variable is not set, it is just shown as empty.
- Documents the option as a side-hint:
   ```text
   If you are building the adapter often, you may want to build it in parallel. You can set the environment variable `WM_NCOMP_PROCS` to the number of parallel threads you want WMake to use.
   ```

TODO list:

- [ ] Squash before merging

### DIFF
--- a/changelog-entries/244.md
+++ b/changelog-entries/244.md
@@ -1,0 +1,1 @@
+- Removed the default `-j 4` option from the wmake command. Instead, documented the `WM_NCOMPPROCS` option of OpenFOAM. [#244](Remove -j 4 flag, document WM_NCOMPPROCS)
--- a/changelog-entries/244.md
+++ b/changelog-entries/244.md
@@ -1,1 +1,1 @@
-- Removed the default `-j 4` option from the wmake command. Instead, documented the `WM_NCOMPPROCS` option of OpenFOAM. [#244](Remove -j 4 flag, document WM_NCOMPPROCS)
+- Removed the default `-j 4` option from the wmake command. Instead, documented the `WM_NCOMPPROCS` option of OpenFOAM. [#244](https://github.com/precice/openfoam-adapter/pull/244)
--- a/docs/get.md
+++ b/docs/get.md
@@ -18,6 +18,8 @@ The adapter also requires [pkg-config](https://linux.die.net/man/1/pkg-config) t
 
 Adding `-DADAPTER_DEBUG_MODE` flag to the `ADAPTER_PREP_FLAGS` activates additional debug messages. You may also change the target directory or specify the number of threads to use for the compilation. See the comments in `Allwmake` for more.
 
+If you are building the adapter often, you may want to build it in parallel. You can set the environment variable `WM_NCOMP_PROCS` to the number of parallel threads you want WMake to use.
+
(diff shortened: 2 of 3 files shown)


changelog-entries/244.md
```diff
@@ -0,0 +1 @@
+- Removed the default `-j 4` option from the wmake command. Instead, documented the `WM_NCOMPPROCS` option of OpenFOAM. [#244](Remove -j 4 flag, document WM_NCOMPPROCS)
```
davidscn, line 1:
```suggestion
- Removed the default `-j 4` option from the wmake command. Instead, documented the `WM_NCOMPPROCS` option of OpenFOAM. [#244](https://github.com/precice/openfoam-adapter/pull/244)
```

MakisH: Thanks for the fix! Please check again and approve, if ok.

docs/get.md
```diff
@@ -13,12 +13,13 @@ To build the adapter, you need to install a few dependencies and then execute th
 4. Execute the build script: `./Allwmake`.
     * See and adjust the configuration in the beginning of the script first, if needed.
     * Check for any error messages and suggestions at the end.
-    * Modify the `adapter_build_command` to e.g. build using more threads, e.g. `wmake -j 4 libso`.
 
 The adapter also requires [pkg-config](https://linux.die.net/man/1/pkg-config) to [link to preCICE](https://precice.org/installation-linking.html). This is a very common dependency on Linux and is usually already installed.
 
 Adding `-DADAPTER_DEBUG_MODE` flag to the `ADAPTER_PREP_FLAGS` activates additional debug messages. You may also change the target directory or specify the number of threads to use for the compilation. See the comments in `Allwmake` for more.
 
+If you are building the adapter often, you may want to build it in parallel. You can set the environment variable `WM_NCOMP_PROCS` to the number of parallel threads you want WMake to use.
```
davidscn, line 21: That's not the same spelling as above.

davidscn (commented):
Looks like a neat and OpenFOAM native way. Only thing from my side:

When the variable is undefined, I get the following log

```
  WM_NCOMPPROCS =  (number of threads for parallel builds)
```
I would define the variable to be `1` in case it is not defined, just to make it well-defined and user-visible.

MakisH:
> Looks like a neat and OpenFOAM native way. Only thing from my side:
> 
> When the variable is undefined, I get the following log
> 
> ```
>   WM_NCOMPPROCS =  (number of threads for parallel builds)
> ```
> 
> I would define the variable to be `1` in case it is not defined, just to make it well-defined and user-visible.

Thanks for the feedback, I added a default value if unset. It should now be:

```
  WM_NCOMPPROCS = 1 (number of threads for parallel builds)
```

Issue #237, opened 2022-08-03
Clang compiler gives several related warnings about overridden member functions


solids4foam:
Building the OpenFOAM adapter with clang gives several warnings related to overridden member functions. Details are given below.

- Compiler: Apple clang version 12.0.5 (clang-1205.0.22.9)
- Adapter version/commit number/last download:
- OpenFOAM version: OpenFOAM-v2012
- preCICE version: openfoam-adapter_v1.1.0_OpenFOAMv1812-v2112-newer
- Attach the adapter building files `allwmake.log`, `ldd.log`, as well as any logs that could be helpful: wmake.log, allwmake.log and ldd.log (note that ldd is unavailable on macOS but I can get something equivalent: see below) attached.

A typical warning looks as follows (you can see all the warnings in the attached wmake.log): 

```cpp
In file included from CHT/Temperature.C:1:
CHT/Temperature.H:29:10: warning: 'write' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    void write(double* buffer, bool meshConnectivity, const unsigned int dim);
         ^
./CouplingDataUser.H:70:18: note: overridden virtual function is here
    virtual void write(double* dataBuffer, bool meshConnectivity, const unsigned int dim) = 0;
                 ^
In file included from CHT/Temperature.C:1:
CHT/Temperature.H:32:10: warning: 'read' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    void read(double* buffer, const unsigned int dim);
         ^
./CouplingDataUser.H:73:18: note: overridden virtual function is here
    virtual void read(double* dataBuffer, const unsigned int dim) = 0;
                 ^
In file included from CHT/HeatFlux.C:1:
CHT/HeatFlux.H:36:18: warning: 'write' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    virtual void write(double* buffer, bool meshConnectivity, const unsigned int dim);
                 ^
./CouplingDataUser.H:70:18: note: overridden virtual function is here
    virtual void write(double* dataBuffer, bool meshConnectivity, const unsigned int dim) = 0;
                 ^
In file included from CHT/HeatFlux.C:1:
CHT/HeatFlux.H:40:18: warning: 'read' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    virtual void read(double* buffer, const unsigned int dim);
                 ^
./CouplingDataUser.H:73:18: note: overridden virtual function is here
[...]

PR #245, opened 2022-09-19
Cleanup virtual/override/final

MakisH:
This is implementing the [C++ Core Guideline 128](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final) and is mainly a reaction to https://github.com/precice/openfoam-adapter/issues/237.

I tried reproducing with GCC and `-Wsuggest-override`, but this led to an unreadably long log, since OpenFOAM itself does not seem to follow the rule. I also tried making WMake to use Clang on my system, which in the end I could not figure out. @solids4foam / @philipcardiff could you please try if this PR compiles fine for you?

@davidscn please also have a look. Possible side-effect: if people extend the adapter by deriving from some of our classes, the `final` may restrict them from doing so. If anyone runs in this case, please report here, and just change the `final` to `override`. But for now, I think stricter measures are saner.

TODO list:

### DIFF
--- a/CHT/HeatFlux.H
+++ b/CHT/HeatFlux.H
@@ -33,16 +33,16 @@ public:
 
     //- Compute heat flux values from the temperature field
     //  and write them into the buffer
-    virtual void write(double* buffer, bool meshConnectivity, const unsigned int dim);
+    void write(double* buffer, bool meshConnectivity, const unsigned int dim) final;
 
     //- Read heat flux values from the buffer and assign them to
     //  the gradient of the temperature field
-    virtual void read(double* buffer, const unsigned int dim);
+    void read(double* buffer, const unsigned int dim) final;
 
-    bool isLocationTypeSupported(const bool meshConnectivity) const override;
+    bool isLocationTypeSupported(const bool meshConnectivity) const final;
 
     //- Get the name of the current data field
-    std::string getDataName() const override;
+    std::string getDataName() const final;
 
     //- Destructor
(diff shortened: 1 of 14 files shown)


davidscn (approved):

philipcardiff:
> This is implementing the [C++ Core Guideline 128](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final) and is mainly a reaction to #237.
> 
> I tried reproducing with GCC and `-Wsuggest-override`, but this led to an unreadably long log, since OpenFOAM itself does not seem to follow the rule. I also tried making WMake to use Clang on my system, which in the end I could not figure out. @solids4foam / @philipcardiff could you please try if this PR compiles fine for you?
> 
> @davidscn please also have a look. Possible side-effect: if people extend the adapter by deriving from some of our classes, the `final` may restrict them from doing so. If anyone runs in this case, please report here, and just change the `final` to `override`. But for now, I think stricter measures are saner.
> 
> TODO list:
> 
> * I updated the documentation in `docs/` -> N/A
> * [x]  I added a changelog entry in `changelog-entries/` (create directory if missing)

I can confirm that the warnings are now gone when I compile with Clang.

I used `Apple clang version 12.0.5 (clang-1205.0.22.9)`.

PR #246, opened 2022-09-19
Change xy-plane check error to warning

### LINKED ISSUE #228: Enable 2d axisymmetric cases
For 2d cases with a wedge for axisymmetric simulation this causes an error. I'd suggest to change that to a warning.

https://github.com/precice/openfoam-adapter/blob/a95da52062c2f503ce58d9509c3e47f5efdbf149/Interface.C#L156

MakisH:
This adds support for axisymmetric cases, closes #228.

@arvedes FYI (I also directly used the `adapterInfo` function). Thanks for contributing the suggestion!

@davidscn please do a sanity check. In #228, you suggested a more elaborate fix, but according to @arvedes, this is already enough for his use case.

TODO list:

- [ ] Squash before merging

### DIFF
--- a/Interface.C
+++ b/Interface.C
@@ -153,11 +153,11 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
                     }
                     else
                     {
-                        FatalErrorInFunction
-                            << "It seems like you are using preCICE in 2D and your geometry is not located int the xy-plane. "
-                               "The OpenFOAM adapter implementation supports preCICE 2D cases only with the z-axis as out-of-plane direction."
-                               "Please rotate your geometry so that the geometry is located in the xy-plane."
-                            << exit(FatalError);
+                        adapterInfo("It seems like you are using preCICE in 2D and your geometry is not located int the xy-plane. "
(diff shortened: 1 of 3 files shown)


arvedes:
Great, thanks a lot!

davidscn (commented):
> you suggested a more elaborate fix, but according to @arvedes, this is already enough for his use case.

Well, of course it is enough, as it removes now the barrier completely, whereas my idea was to remove the barrier only conditional. There is not much to review here, if you think that it is safe enough, feel free to merge,

MakisH:
> > you suggested a more elaborate fix, but according to @arvedes, this is already enough for his use case.
> 
> Well, of course it is enough, as it removes now the barrier completely, whereas my idea was to remove the barrier only conditional. There is not much to review here, if you think that it is safe enough, feel free to merge,

A malformed case should anyway crash later or for other reasons, right? So, a warning should be enough here to understand what went wrong (assuming a user would look at the beginning of the log).

davidscn:
This concern is only about usability, yes.

Issue #235, opened 2022-07-27
Compatibility with preCICE v3


MakisH:
At some point in the next year, preCICE v3 is coming. We need to update the adapter accordingly.

Changes to consider:

- [x] Drop issue https://github.com/precice/openfoam-adapter/issues/3
- [x] Merged `initializeData` into `intialize`: https://github.com/precice/precice/pull/1350 -> Solved by #266.
- [x] Updated connectivity API (adjust `setMeshEdges` and `setMeshTriangles` to use vertex IDs): https://github.com/precice/precice/pull/1322, https://github.com/precice/precice/issues/1321 -> Solved by #297.
- [x] Updated actions/checkpointing API: https://github.com/precice/precice/issues/719 -> Solved by #285
- [ ] Unsigned sizes: https://github.com/precice/precice/issues/405
- [ ] (additional) Exiting the coupling earlier, if needed: https://github.com/precice/precice/issues/1118
- [x] Apply any waveform iteration related changes needed. -> Solved by #285.

MakisH:
The adapter supports preCICE v3 since v1.3.0.

Issue #58, opened 2018-12-20
FSI: Subcycling with implicit coupling


MakisH:
Additional fields should be checkpointed, for stability. Meanwhile, these fields should not be checkpointed when we don't subcycle.

See the method `storeMeshPoints()` in `Adapter.C` for more. @derekrisseeuw could you maybe provide a few more details?

Relevant to #7.

derekrisseeuw:
In the openfoam simulation 3 parts need updating. A distinction is being made between fluid fields and mesh fields, which both need checkpointing. The Time is the third element which needs updating. Fluid fields are all the <type><rank>Fields, such as the velocity which is a volVectorField. These can include old fields.

For the mesh only the points and oldpoints (location of the mesh) and the mesh flux area read. This is 

The fields are stored and read when required by preCICE by evaluating the bool `isWriteCheckpointRequired` and `isReadCheckpointRequired`. If so, the respective functions `writeCheckpoint` or `readCheckpoint` are called which follow similar procedures but one is for writing and the other for reading. 

In writeCheckpoint, first the time is checkpointed, next the mesh is written. This includes writing the checkpoint, and the meshPhi. This is valid for both implicit simulations with the same timestep as simulations with subcycling. The catch is found in the checkpointing of the mesh (old)volumes: V, V0 (and V00 dependent on the time scheme). Without subcycling these are not to be touched because openFOAM computes them correctly for the single timestep. However, for subcycling they should be set back to the original checkpointed value, see the comments in the `Adapter.C` for this.

MakisH:
In other words (correct me if I am wrong):

* When there is no coupling at all, the solver updates the old values at every time step, doing something like `V00 = V0`, `V0 = V`, `V = compute()`.

* Without subcycling, the solver only executes one step before having to return to the previous coupling time and reload the checkpoint, so it has executed the same update as above (once). When reloading the mesh, we should also return `V` to its value before `compute()`. **The important question is:** when does OpenFOAM update the `V0` and `V00`? If in the meantime they have not been updated yet at the time we reload the checkpoint, then we should leave them as is. But why not store them and reload them eitherways? (apart from any performance reasons)

* When subcycling, the solver executes multiple small time steps until it has to reload a checkpoint, so it has executed multiple updates. If we do not restore the `V0` and `V00` as well, there is a problem.

* In explicit coupling, it doesn't matter (there is no checkpointing).

By the way, this is related to #55.

MakisH:
[...]

PR #247, opened 2022-09-19
Remove deprecated isReadDataAvailable

MakisH:
As discussed in https://github.com/precice/precice/issues/1223, preCICE v2.5.0 deprecated `isReadDataAvailable()`, which is planned for removal in preCICE v3. Related: #235.

This PR already removes this call, to avoid the deprecation warning when compiling the adapter.

It should currently not matter for results, as it re-reads the same information on the boundary, but it should make time interpolation easier in the future. I quickly ran the flow-over-heated-plate case with subcycling (2 solver time steps for 1 coupling time window) and implicit coupling (hint: we currently have issues with FSI and subcycling #58), and the result files are identical. In this small case, the runtime seems to be approximately the same (without any rigorous performance test, just one sample for each case).

Still, the adapter performs now each assignment on the boundaries every time.

@uekerman / @davidscn Anything else to consider here?

The diff may look a bit complicated, but it is just removing one `if` check and indenting-out.

TODO list:

### DIFF
--- a/Interface.C
+++ b/Interface.C
@@ -371,38 +371,34 @@ void preciceAdapter::Interface::createBuffer()
 
 void preciceAdapter::Interface::readCouplingData()
 {
-    // Are new data available or is the participant subcycling?
-    if (precice_.isReadDataAvailable())
+    // Make every coupling data reader read
+    for (uint i = 0; i < couplingDataReaders_.size(); i++)
     {
-        // Make every coupling data reader read
-        for (uint i = 0; i < couplingDataReaders_.size(); i++)
+        // Pointer to the current reader
+        preciceAdapter::CouplingDataUser*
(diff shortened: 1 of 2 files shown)


davidscn (approved):
Looks good

thesamriel:
Using serial coupling, participant 1 reads all quantities from a zero buffer before the first timestep. This sets the boundary values at the interface to 0 for all read variables, correct?

This becomes problematic when using compressible flows in fluid-fluid coupling, since a 0 pressure leads to a division by 0 error.

Ran into this problem for my *flow over heated wall segment* test case for fluid-fluid coupling after updating the adapter.

MakisH:
This sounds like a data initialization issue to me, which should be documented independently.

The purpose of the `isReadDataAvailable` was mostly to not read again and again when subcycling.

PR #248, opened 2022-09-19
Reword warning for nearest-projection mapping

### LINKED ISSUE #135: Wrong Assertion for nearest projection mappings
At the time we implemented the nearest projection mapping, we asserted, that it is not used in FSI simulations, since there, the `Solid` participant needs to provide connectivity.
However, after introducing #125 , this is not valid any more, since we can use nearest-projection mappings for stress data.

https://github.com/precice/openfoam-adapter/blob/b1d80efe3c2ec964f157687753a675dd4502dba7/Adapter.C#L150-L159

Either we should remove this Assertion completely or just raise a warning, that this is only relevant for FSI in combination with stress data (which is currently only supported by our deal.II adapter).

MakisH:
Closes #135.

I followed the easier approach for now: convert the error to a warning (when FSI + connectivity) and add a note in the documentation. I think it does not need to be more complicated, while configuring FSI with mesh connectivity is already a specific enough case for a warning.

The adapter now refers to the documentation via URL, which I think we should start doing more often. Even if the targets change eventually, a URL is better than looking manually through the documentation.

@davidscn please check. I have not yet checked what happens with solid OpenFOAM solvers that can write displacements.

TODO list:

- [x] I added a changelog entry in `changelog-entries/` (create directory if missing) -> not needed with the changes after review
- [ ] Squash before merging

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -279,7 +279,7 @@ Data is obtained at the face centers, then interpolated to face nodes. Here, we 
 It is important to notice that the target data location is again the face center mesh of the coupling partner. In the standard CHT case, where both data sets are exchanged by a nearest-projection mapping, this leads to two interface meshes (centers and nodes) per participant. Having both the centers and nodes defined, we can skip one interpolation step and read data directly to the centers (cf. picture solver B).
 
 {% note %}
-As already mentioned, the `Fluid` participant does not need to provide the mesh connectivity in case of a standard FSI. Therefore, the `Solid` participant needs to provide it and nothing special needs to be considered compared to other mapping methods. This implementation supports all CHT-related fields, which are mapped with a `consistent` constraint.
+As already mentioned, the `Fluid` participant does not need to provide the mesh connectivity in case of a standard FSI writing forces. It can, however, provide the mesh connectivity when writing stresses. Alternatively, the `Solid` participant can to provide mesh connectivity on the displacements mesh, and nothing special needs to be considered in the adapter implementation compared to other mapping methods. This implementation supports all CHT-related fields, which are mapped with a `consistent` constraint.
 {% endnote %}
 
 ### Additional properties for some solvers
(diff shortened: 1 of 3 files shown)


docs/config.md
```diff
@@ -279,7 +279,7 @@ Data is obtained at the face centers, then interpolated to face nodes. Here, we
...
 
 {% note %}
-As already mentioned, the `Fluid` participant does not need to provide the mesh connectivity in case of a standard FSI. Therefore, the `Solid` participant needs to provide it and nothing special needs to be considered compared to other mapping methods. This implementation supports all CHT-related fields, which are mapped with a `consistent` constraint.
+As already mentioned, the `Fluid` participant does not need to provide the mesh connectivity in case of a standard FSI writing forces. It can, however, provide the mesh connectivity when writing stresses. Alternatively, the `Solid` participant can provide mesh connectivity on the displacements mesh, in which case nothing special needs to be considered in the adapter implementation compared to other mapping methods. This implementation supports all CHT-related fields, which are mapped with a `consistent` constraint.
```
davidscn, line 282: No unfortunately not. We can provide the connectivity, but writing the data (e.g. `Stress`) won't be handled correctly. That's one of (my) bad design decisions, which requires to address this in each data field again. Right now, I can only think of scenarios where this behavior would either fallback to NN (such as for `Forces` etc.) and be ignored or would just be erroneous. Enabling it doesn't make sense in my opinion, unless we change the implementation.

davidscn (changes requested):

docs/config.md
MakisH: How would you phrase this part of the documentation, then?

davidscn: Correct would be: Mapping with connectivity information is not implemented for FSI and therefore, not supported. It's only supported for CHT related fields.

MakisH: I updated the message and made it trigger an error again.

MakisH:
I updated the documentation according to the review (forgotten that yesterday) and deleted the changelog entry, as the change is now tiny. @davidscn back to you.

davidscn (approved):

Issue #229, opened 2022-06-26
Missing support for volume coupling


MakisH:
By "volume coupling" we mean coupling overlapping regions. While preCICE was mainly designed for surface coupling, most of the features can be applied for volume coupling as well, since we only need to give a cloud of points to preCICE. Read more in the preCICE forum/FAQ: https://precice.discourse.group/t/can-precice-be-used-for-volume-coupling/27

Some tutorials with volume coupling in preCICE include:
- [Volume-coupled diffusion](https://precice.org/tutorials-volume-coupled-diffusion.html) (coupling two FEniCS instances)
- [Channel transport](https://precice.org/tutorials-channel-transport.html) (coupling two Nutils instances)

For the OpenFOAM adapter, in particular, the community has already built prototypes:
- First was https://github.com/precice/openfoam-adapter/pull/88, who extended the adapter to couple OpenFOAM with an LBM solver, overlapping over a region of the domain.
- Then there was https://github.com/precice/openfoam-adapter/pull/97, based on https://upcommons.upc.edu/handle/2117/189920, coupling OpenFOAM with LS-DYNA for CHT.
   - https://github.com/precice/openfoam-adapter/pull/183 builds upon that, trying to couple any state variable with OpenFOAM.

Together with @tirgendetwas, we are trying to unify all these contributions into one general approach, integrate them into the main development branch of the adapter, and create a few tutorials. We plan to support at least CHT and FF, as well as the latest available mapping options for volume-coupled scenarios.

This issue serves as a main status board, in which we will track the status and refer to related issues and PRs. We will consider it "closed" when volume coupling makes it into a release (expected around the end of 2022).

MakisH:
Added with https://github.com/precice/openfoam-adapter/pull/255

PR #183, opened 2021-06-17
Volume coupling module

Alphaoo1:
In this PR I add a class for the Volume coupling module.
This will form the basis for further development on the Volume Coupling module, it's data fields and test cases.

TODO list:

- [x] I updated the documentation in `docs/`
- [x] I added a changelog entry in `changelog-entries/` (create directory if missing)

PR #255, opened 2022-10-04
Volume coupling

tirgendetwas:
This is a solution to issue #229 .

Volume coupling implementation for the OpenFOAM adapter. Our main goals are:

**1. The coupling is integrated with existing features** (currently planned for FF and CHT modules):
We started building upon PR #183 but have since decided against a separate Volume Coupling module. The volume coupling is being implemented for each state variable (each variable which would make sense to couple over a volume).

**2. Configurable coupling region:**
Implemented in [PR#270](https://github.com/precice/openfoam-adapter/pull/270)

Relevant tutorials: [channel transport](https://github.com/precice/tutorials/pull/315) and [volume-coupled flow](https://github.com/precice/tutorials/pull/350)

TODO list:


LouieVoit:
Hi,

Don't we want to be able to use the volume coupling also for CHT, FF or FSI ? I'm not sure I understand why it's a separate module. 
Also, I've extended the adapter to use volume coupling in my own fork of the adapter. Within the Interface.C file, I've added a zone option in the preciceDict dictionary to define OpenFOAM zones to use for the coupling. That way one does not have to use all the mesh for the coupling. Should it be added to this PR ?

Thanks :-)
Louis

MakisH:
Hi @LouieVoit,

we started from a past contribution that already implemented volume coupling as a separate module (#183). We have already discussed and concluded that this is not the right approach. We also wanted to add zones, similarly to #84.

This should eventually become clear in the description of this PR (todo @tirgendetwas).

Interface.C
```diff
@@ -259,22 +264,91 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+        // did not work properly
+        //
+        // the module for volume coupling considers the mesh points in the volume and
+        // on the boundary patches in order to take the boundary conditions into account
+
+        // get the number (volume centered) mesh points in the volume
+        numDataLocations_ = mesh.C().size();
+
+        // Count the data locations for all the patches
+        // and add the those to the previously determined number of mesh points in the volume
```
MakisH, line 297:
```suggestion
        // and add those to the previously determined number of mesh points in the volume
```

Interface.C
```diff
@@ -259,22 +264,91 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+    else if (locationType_ == LocationType::volume)
+    {
+        // module for volume coupling is based on the module for coupling faceCenters
+        // as the module was tested and developed on a previous development where faceNodes
+        // did not work properly
+        //
+        // the module for volume coupling considers the mesh points in the volume and
+        // on the boundary patches in order to take the boundary conditions into account
+
+        // get the number (volume centered) mesh points in the volume
```
MakisH, line 293:
```suggestion
        // Get the number of (volume centered) mesh points in the volume
```

Interface.C
```diff
@@ -259,22 +264,91 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+        // Each vertex has one index, but three coordinates.
+        vertexIDs_ = new int[numDataLocations_];
+
+        // Initialize the index of the vertices array
+        int verticesIndex = 0;
+
+        // Get the locations of the volume centered mesh vertices
+        const vectorField& CellCenters = mesh.C();
+
+        for (int i = 0; i < CellCenters.size(); i++)
```
MakisH, line 319:
```suggestion
        for (uint i = 0; i < CellCenters.size(); i++)
```

FF/Velocity.C
```diff
@@ -76,7 +112,7 @@ void preciceAdapter::FF::Velocity::read(double* buffer, const unsigned int dim)
 
 bool preciceAdapter::FF::Velocity::isLocationTypeSupported(const bool meshConnectivity) const
 {
-    return (this->locationType_ == LocationType::faceCenters);
+    return (this->locationType_ == LocationType::faceCenters || this->locationType_ == LocationType::volume);
```
MakisH, line 115: This is about supporting mesh connectivity. Do we support mesh connectivity for volume meshes?

CouplingDataUser.H
```diff
@@ -13,7 +13,8 @@ enum class LocationType
 {
     none,
     faceCenters,
-    faceNodes
+    faceNodes,
+    volume
```
MakisH, line 17: Would it make more sense to call this `cellCenters`, as I think you have it elsewhere? I am not sure.

CHT/Temperature.C
```diff
@@ -88,7 +104,7 @@ bool preciceAdapter::CHT::Temperature::isLocationTypeSupported(const bool meshCo
     }
     else
     {
-        return (this->locationType_ == LocationType::faceCenters);
+        return (this->locationType_ == LocationType::faceCenters || this->locationType_ == LocationType::volume);
```
MakisH, line 107: Same here: specific to connectivity.

MakisH (commented):
Some first comments (not tested yet).

CHT/HeatFlux.C
```diff
@@ -37,20 +37,20 @@ void preciceAdapter::CHT::HeatFlux::write(double* buffer, bool meshConnectivity,
         // If we use the mesh connectivity, we interpolate from the centres to the nodes
         if (meshConnectivity)
         {
-            //Setup Interpolation object
+            // Setup Interpolation object
```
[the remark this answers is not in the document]
tirgendetwas: `.clang-format` actually isn't an empty file, but for some reason the git changelog lists it that way. It seems to be the same as the adapter's original `.clang-format`, I'm not sure what happened in [this commit](https://github.com/precice/openfoam-adapter/commit/579c5e4c511437009a61950baaafbc5b91ce5bae) 

Also, installing `clang-format 11.0.1` and applying the script again made no difference.

Something is not completely right...

CouplingDataUser.H
tirgendetwas: Maybe something like `volumeCenters`? I think it would make sense to keep "volume" in the name, maybe combine the two ideas :smile:

CHT/HeatFlux.C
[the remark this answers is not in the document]
MakisH: Strange... I will have a look before merging, let's keep this unresolved.

CouplingDataUser.H
MakisH: I like the name! Go ahead.

FF/Velocity.C
MakisH: What I meant in this comment was not to implement mesh connectivity as part of this PR (even though it would be nice, but let's document it in an issue first). I just meant to remove this addition, because we don't support it yet.

MakisH: Your original suggestion was in the right direction. I got confused, as the connectivity feature was added on top of everything much later, but the architecture did not really allow us for a nice implementation.

If we add the `|| this->locationType_ == LocationType::volume`, this means that volume centers are supported regardless of mesh connectivity or not. However, we need to distinguish here between mesh connectivity (supporting only `faceCenters`) and no mesh connectivity (supporting both).

As you already pointed out in a direct message, we can do this similarly to the Temperature class.

FF/Pressure.C
```diff
@@ -16,6 +16,15 @@ void preciceAdapter::FF::Pressure::write(double* buffer, bool meshConnectivity,
 {
     int bufferIndex = 0;
 
+    if (this->locationType_ == LocationType::volumeCenters)
+    {
+        // set in Interface::addCouplingDataR/W
```
MakisH, line 21: What does this comment refer to?
Would a comment consistent with the "For every boundary patch of the interface" (i.e., "For every internal cell of the region") be more helpful to understand the intention?

CHT/Temperature.C
```diff
@@ -84,11 +100,11 @@ bool preciceAdapter::CHT::Temperature::isLocationTypeSupported(const bool meshCo
     // always return true and offload the handling to the user.
     if (meshConnectivity)
     {
-        return true;
+        return (this->locationType_ != LocationType::volumeCenters); // we do not support meshConnectivity for volumeCenters (yet)
```
MakisH, line 103: Would a positive formulation ("if faceCenters or faceNodes") be clearer, safer, and consistent, e.g., with the `Pressure.C` and `Velocity.C`?
The comment can remain.

Interface.C
```diff
@@ -275,6 +280,75 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
             }
         }
     }
+
```
MakisH, line 283:
```suggestion
```

Interface.C
```diff
@@ -275,6 +280,75 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+    else if (locationType_ == LocationType::volumeCenters)
+    {
+        // The volume coupling implementation is based on the coupling with faceCenters
+        // as the volume coupling was tested and developed on a previous development where faceNodes
+        // did not work properly
```
MakisH, line 288: I am not sure I understand this comment. Do you mean that the branch that we started from did not yet have `faceNodes` implemented or complete?

In any case, we are assuming cell-based FVM, which is the most common approach (there is also nodes-based FVM, but this is way out of scope of the complete adapter at the moment, as only some extensions of OpenFOAM implement it). As such, all solvers store all fields (with some exceptions) in the centers of the cells.

Independent on what the development was based upon, why does it matter?

Interface.C
```diff
@@ -118,6 +122,7 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
             {
                 const pointField faceNodes =
                     mesh.boundaryMesh()[patchIDs_.at(j)].localPoints();
+
```
MakisH, line 125:
```suggestion
```

Interface.C
```diff
@@ -275,6 +280,75 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
     }
+
+    else if (locationType_ == LocationType::volumeCenters)
+    {
+        // The volume coupling implementation is based on the coupling with faceCenters
+        // as the volume coupling was tested and developed on a previous development where faceNodes
+        // did not work properly
+        //
+        // The volume coupling implementation considers the mesh points in the volume and
+        // on the boundary patches in order to take the boundary conditions into account
```
MakisH, line 291: So, does this identify all points in the volume region + all points on the face centers of the region? Why do we need that?

I would expect that, when coupling a volume, we don't couple the boundaries. Or at least not with the same values. I would expect that the boundaries of the volume are specified by the solver, based on the rest of the domain. This makes mostly sense when coupling only one region of the whole mesh.

MakisH (commented):
Some comments, as I am trying to get an overview of the current state, and keep the rock rolling. No rush to implement.

The main aspect that you may already want to check, from a debugging perspective, is how you specify the locations to couple (volume + patches, while I think it should only be volume).

Interface.C
```diff
@@ -275,6 +280,75 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+        // Each vertex has one index, but three coordinates.
+        vertexIDs_ = new int[numDataLocations_];
+
+        // Initialize the index of the vertices array
+        int verticesIndex = 0;
+
+        // Get the locations of the volume centered mesh vertices
+        const vectorField& CellCenters = mesh.C();
+
+        for (uint i = 0; i < CellCenters.size(); i++)
```
MakisH, line 319: I am getting a compiler warning with OpenFOAM v2212 and GCC 11.3:

```
g++ -std=c++14 -m64 -pthread -DOPENFOAM=2212 -DWM_DP -DWM_LABEL_SIZE=32 -Wall -Wextra -Wold-style-cast -Wnon-virtual-dtor -Wno-unused-parameter -Wno-invalid-offsetof -Wno-attributes -Wno-unknown-pragmas -O3  -DNoRepository -ftemplate-depth-100 -fdiagnostics-color=always -I/usr/lib/openfoam/openfoam2212/src/finiteVolume/lnInclude -I/usr/lib/openfoam/openfoam2212/src/meshTools/lnInclude -I/usr/lib/openfoam/openfoam2212/src/transportModels/ -I/usr/lib/openfoam/openfoam2212/src/transportModels/incompressible/lnInclude -I/usr/lib/openfoam/openfoam2212/src/transportModels/compressible/lnInclude -I/usr/lib/openfoam/openfoam2212/src/transportModels/twoPhaseMixture/lnInclude -I/usr/lib/openfoam/openfoam2212/src/transportModels/interfaceProperties/lnInclude -I/usr/lib/openfoam/openfoam2212/src/transportModels/immiscibleIncompressibleTwoPhaseMixture/lnInclude -I/usr/lib/openfoam/openfoam2212/src/thermophysicalModels/basic/lnInclude -I/usr/lib/openfoam/openfoam2212/src/TurbulenceModels/turbulenceModels/lnInclude -I/usr/lib/openfoam/openfoam2212/src/TurbulenceModels/compressible/lnInclude -I/usr/lib/openfoam/openfoam2212/src/TurbulenceModels/incompressible/lnInclude -I/usr/lib/openfoam/openfoam2212/src/triSurface/lnInclude -I/home/gc/repos/precice/precice/src -I../  -iquote. -IlnInclude -I/usr/lib/openfoam/openfoam2212/src/OpenFOAM/lnInclude -I/usr/lib/openfoam/openfoam2212/src/OSspecific/POSIX/lnInclude   -fPIC -c FSI/FSI.C -o Make/linux64GccDPInt32Opt/FSI/FSI.o
Interface.C: In member function ‘void preciceAdapter::Interface::configureMesh(const Foam::fvMesh&, const string&, const string&)’:
Interface.C:362:28: warning: comparison of integer expressions of different signedness: ‘uint’ {aka ‘unsigned int’} and ‘Foam::label’ {aka ‘int’} [-Wsign-compare]
  362 |         for (uint i = 0; i < CellCenters.size(); i++)
      |            
```

(not sure why I get different line numbers)

tirgendetwas: It seems that `CellCenters.size()` is of type `Foam::label` which is an `int64_t`/`int32_t` (signed) under the hood. Solved the problem by casting for now...

FF/Pressure.C
tirgendetwas: I added this comment for myself at an early stage and forgot to remove it later on. Fixed!

Interface.C
MakisH: Why not just `auto`? We can use C++11 features.

Interface.C
tirgendetwas: This is a comment from Prasad's original fork. It might have been relevant at the time, but if it isn't anymore I can remove it (or only leave the first line)

Interface.C
tirgendetwas: The coupling includes the cell centers on the whole domain (at this implementation stage) and the face centers for all specified patches - basically adding the cell centers of the volume to the `face nodes`-surface coupling case. 

So in a tutorial case which uses `locations    volumeCenters`, all of the internal fields will be coupled (cell centers), and additionally specified patches can be coupled as well (on face centers).

Interface.C
tirgendetwas: Thanks, that worked!

Interface.C
MakisH: I think we can remove this.

Interface.C
MakisH: So, if I provide `locations volumeCenters` and don't specify any patches, only the volume region will be modified, without the patches. If I also specify patches, these will also be modified. This sounds reasonable behavior to me, we just need to document it.

Do we need to use different coupling meshes for volumes and patches? I am trying to think of mapping issues, especially now that we have volume mapping as well. @uekerman

uekerman: mmh, not too sure about this (if I understood it correctly). Mixing surface and volume coupling here could confuse more than it helps. There could also be cases where we want to read values in the volume from one solver and additionally read values on the surface from another solver. I would always treat volume and surface as two different meshes in preCICE.
Or is there a specific use case where we would need this feature?

tirgendetwas: So how should we resolve this? Previous applications in that area have used both patches and volume also ([PR#88](https://github.com/precice/openfoam-adapter/pull/88/files) and [PR#97](https://github.com/precice/openfoam-adapter/pull/97/files)), so I suspected it might have been necessary and kept it like that. The channel transport [tutorial](https://github.com/precice/tutorials/pull/315) case I have set up looks very similar whether I couple patches + volume or only volume.

mosayebshams:
Hi @MakisH and @tirgendetwas,

I was wondering if there was a toy-case for the volume-coupling to guide me through adjusting `precice-config.xml` and `preciceDict` files to read pressure data from specific volume locations in a fluid-openfoam simulation.

docs/README.md
```diff
@@ -27,6 +27,14 @@ This adapter can read/write the following fields:
...
 - Velocity gradient (read + write)
 
+In addition, the adapter supports the following fields in a volume coupling setup:
+
+- Temperature (read* + write)
+- Pressure (read* + write)
+- Velocity (read + write)
+
+(*) Technical foundation available, the user needs to provide additional, application-specific code for enforcing the source terms.
+
```
MakisH, line 37: I polished a bit the documentation and added volume coupling in the overview page of the adapter documentation.

@tirgendetwas is this formulation correct?

MakisH (approved):
The code looks quite clean, and this is very ready to merge! :rocket: :smile: Thanks, a lot!

I am running the tutorials and will merge once I have checked everything.

(The CI is just broken because the OpenFOAM packages are failing the integrity check. It works on my machine.)

docs/README.md
tirgendetwas: I'm not sure, what do you mean by "the user needs to provide additional, application-specific code"? We haven't managed to read Pressure so far. Temperature should probably work if it's done through the enthalpy, but we haven't tried that

MakisH: For temperature, I understand that one just needs to provided a codedSource that corresponds to their thermophysical model, no?

I can remove the reading part from temperature and pressure.

PR #256, opened 2022-11-09
Add timers

MakisH:
## Motivation

We so far know the time spent in OpenFOAM (total), as well as the time spent in preCICE. However, we have no easy way to find out the time spent only in the adapter.

This is a first attempt to add timers to the adapter.

## Implementation

This uses the [`clockValue`](https://www.openfoam.com/documentation/guides/v2206/api/classFoam_1_1clockValue.html) type already defined by OpenFOAM, which uses a [`std::chrono::high_resolution_clock`](https://en.cppreference.com/w/cpp/chrono/high_resolution_clock) under the hood.

Not to be confused with [`clockTime`](https://www.openfoam.com/documentation/guides/latest/api/classFoam_1_1clockTime.html). Not sure if that one would be better, but the [OpenFOAM clock test](https://develop.openfoam.com/Development/openfoam/-/blob/master/applications/test/clock/Test-clock.C) seems to be using `clockValue`.

I am declaring two types in the header file of the function object:
- `clockValue clockGlobal_` serves as a clock that is reset and sampled many times.
- `clockValue timeInAll_` serves as a cumulative timer.

Before each timed region, the clock is sampled with `clockGlobal_.update()`.
After each timed region, the clock is sampled and the duration from the last sample is computed with `clockGlobal_.elapsed()`.

## How to test this

Run a preCICE tutorial. For the flow-over-heated-plate, I get the following for the Fluid participant:
```text
ExecutionTime = 21.63 s  ClockTime = 25 s
(...)
Total time spent in the adapter and preCICE: "00:00.929" (format: day-hh:mm:ss.ms)
```

and for the Solid participant (second in the implicit coupling scheme, i.e., computing the acceleration):
```text
ExecutionTime = 0.3 s  ClockTime = 24 s
(...)
Total time spent in the adapter and preCICE: "00:22.025" (format: day-hh:mm:ss.ms)
```

Here are the event files:
- [precice-Fluid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9975086/precice-Fluid-events-summary.log)
- [precice-Solid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9975092/precice-Solid-events-summary.log)

Conclusions:
- Edit: It looks like the timer previously counted the initialization time twice in the adapter. The actual numbers are smaller than listed here.
- The time spent in the adapter + preCICE is trivial for the Fluid (which is also the slow solver in the simulation).
- The time spent in the adapter + preCICE is quite large for the Solid, but this is only because the case is extremely light.
- On the Fluid side, preCICE reports 232ms in `advance()`, 186ms in `initialize()`, 44ms in `finalize()`. If we extract these from the 929ms spent in the (adapter + preCICE), the time spent in the adapter is 929ms - 232ms - 186ms - 44ms = 467ms. (<2% of ClockTime)
- On the Solid side, preCICE reports 21367ms in `advance()` (most of it in waiting for data from Fluid), 317ms in `initialize()`, 89ms in `finalize()`. If we extract this from the 22025ms of the (adapter + preCICE), the time spent in the adapter is 252ms. (1% of ClockTime)
- General conclusion: The time contribution of the adapter is very small compared to the time needed by preCICE (which, as we know from previous performance tests, is trivial compared to the solver time for large cases).

## Potential extensions

A global clock should be enough to extract the time spent only in the adapter, when compared with the preCICE events summary files. We could explicitly track the calls to preCICE with a separate timer, but I am afraid that would make the code a bit more complicated than needed.

### Working with multiple timers

The first commit demonstrates an example with three timers in the constructor of `Adapter`, but this will squashed away later on. I add it here since I think it helps understand how to use clocks in OpenFOAM:

```c++
    clockGlobal_.update();

    clockAdapter_.update();
    Info << "Timing something that takes 1s..." << nl;
    sleep(1);
    timeInAdapter_ += clockAdapter_.elapsed();

    Info << "timeInAdapter = " << timeInAdapter_.seconds() << "s" << nl;

    // excluded time
    Info << "Executing something in preCICE for 1s, which shoud be timed separately..." << nl;
    clockPrecice_.update();
    sleep(1);
    timeInPrecice_ += clockPrecice_.elapsed();
    Info << "timeInAdapter = " << timeInAdapter_.seconds() << "s" << nl;
    Info << "timeInPrecice = " << timeInPrecice_.seconds() << "s" << nl;

    // included time
    clockAdapter_.update();
    Info << "Timing something that takes 1s..." << nl;
    sleep(1);
    timeInAdapter_ += clockAdapter_.elapsed();
    Info << "elapsed = " << clockAdapter_.elapsed() << nl;
    Info << "timeInAdapter = " << timeInAdapter_.seconds() << "s" << nl;
    Info << "timeInPrecice = " << timeInPrecice_.seconds() << "s" << nl;

    timeInAll_ += clockGlobal_.elapsed();
    Info << "timeInAll = " << timeInAll_.seconds() << "s" << nl;
```

This leads to the following output:

```text
Timing something that takes 1s...
timeInAdapter = 1s
Executing something in preCICE for 1s, which shoud be timed separately...
timeInAdapter = 1s
timeInPrecice = 1s
Timing something that takes 1s...
elapsed = 1.00025
timeInAdapter = 2s
timeInPrecice = 1s
timeInAll = 3s
```

TODO list:

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -12,34 +12,6 @@ preciceAdapter::Adapter::Adapter(const Time& runTime, const fvMesh& mesh)
 {
     adapterInfo("Loaded the OpenFOAM-preCICE adapter - v1.1.0 + unreleased changes.", "info");
 
-    /* Example of using different timers - remove before merging,
-       together with the timers defined in Adapter.H.
-       
-    clockAdapter_.update();
-    Info << "Timing something that takes 1s..." << nl;
-    sleep(1);
-    timeInAdapter_ += clockAdapter_.elapsed();
-
-    Info << "timeInAdapter = " << timeInAdapter_.seconds() << "s" << nl;
-
-    // excluded time
-    Info << "Executing something in preCICE for 1s, which shoud be timed separately..." << nl;
-    clockPrecice_.update();
-    sleep(1);
-    timeInPrecice_ += clockPrecice_.elapsed();
-    Info << "timeInAdapter = " << timeInAdapter_.seconds() << "s" << nl;
-    Info << "timeInPrecice = " << timeInPrecice_.seconds() << "s" << nl;
-
-    // included time
-    clockAdapter_.update();
-    Info << "Timing something that takes 1s..." << nl;
-    sleep(1);
-    timeInAdapter_ += clockAdapter_.elapsed();
-    Info << "elapsed = " << clockAdapter_.elapsed() << nl;
-    Info << "timeInAdapter = " << timeInAdapter_.seconds() << "s" << nl;
-    Info << "timeInPrecice = " << timeInPrecice_.seconds() << "s" << nl;
-    */
-
     return;
 }
 
--- a/Adapter.C
+++ b/Adapter.C
@@ -12,6 +12,35 @@ preciceAdapter::Adapter::Adapter(const Time& runTime, const fvMesh& mesh)
 {
     adapterInfo("Loaded the OpenFOAM-preCICE adapter - v1.1.0 + unreleased changes.", "info");
 
+    clockGlobal_.update();
+
+    clockAdapter_.update();
+    Info << "Timing something that takes 1s..." << nl;
+    sleep(1);
+    timeInAdapter_ += clockAdapter_.elapsed();
+
+    Info << "timeInAdapter = " << timeInAdapter_.seconds() << "s" << nl;
+
+    // excluded time
+    Info << "Executing something in preCICE for 1s, which shoud be timed separately..." << nl;
+    clockPrecice_.update();
+    sleep(1);
+    timeInPrecice_ += clockPrecice_.elapsed();
+    Info << "timeInAdapter = " << timeInAdapter_.seconds() << "s" << nl;
+    Info << "timeInPrecice = " << timeInPrecice_.seconds() << "s" << nl;
+
+    // included time
+    clockAdapter_.update();
+    Info << "Timing something that takes 1s..." << nl;
+    sleep(1);
+    timeInAdapter_ += clockAdapter_.elapsed();
+    Info << "elapsed = " << clockAdapter_.elapsed() << nl;
+    Info << "timeInAdapter = " << timeInAdapter_.seconds() << "s" << nl;
+    Info << "timeInPrecice = " << timeInPrecice_.seconds() << "s" << nl;
+
+    timeInAll_ += clockGlobal_.elapsed();
+    Info << "timeInAll = " << timeInAll_.seconds() << "s" << nl;
+
     return;
 }
 
--- a/Adapter.C
+++ b/Adapter.C
@@ -246,7 +246,8 @@ void preciceAdapter::Adapter::configure()
         DEBUG(adapterInfo("  preCICE solver interface was created."));
 
         // Create interfaces
-        clockAdapter_.update();
+        clockValue clock;
+        clock.update();
         DEBUG(adapterInfo("Creating interfaces..."));
         for (uint i = 0; i < interfacesConfig_.size(); i++)
         {
@@ -334,7 +335,7 @@ void preciceAdapter::Adapter::configure()
             // Create the interface's data buffer
             interface->createBuffer();
         }
-        timeInMeshSetup_ += clockAdapter_.elapsed();
+        timeInMeshSetup_ += clock.elapsed();
 
         // Initialize preCICE and exchange the first coupling data
         initialize();
@@ -492,7 +493,8 @@ void preciceAdapter::Adapter::adjustTimeStep()
 
 void preciceAdapter::Adapter::readCouplingData()
 {
-    clockAdapter_.update();
+    clockValue clock;
+    clock.update();
     DEBUG(adapterInfo("Reading coupling data..."));
 
     for (uint i = 0; i < interfaces_.size(); i++)
@@ -500,13 +502,14 @@ void preciceAdapter::Adapter::readCouplingData()
         interfaces_.at(i)->readCouplingData();
     }
 
-    timeInRead_ += clockAdapter_.elapsed();
+    timeInRead_ += clock.elapsed();
     return;
 }
 
 void preciceAdapter::Adapter::writeCouplingData()
 {
-    clockAdapter_.update();
+    clockValue clock;
+    clock.update();
     DEBUG(adapterInfo("Writing coupling data..."));
 
     for (uint i = 0; i < interfaces_.size(); i++)
@@ -514,16 +517,17 @@ void preciceAdapter::Adapter::writeCouplingData()
         interfaces_.at(i)->writeCouplingData();
     }
 
-    timeInWrite_ += clockAdapter_.elapsed();
+    timeInWrite_ += clock.elapsed();
     return;
 }
 
 void preciceAdapter::Adapter::initialize()
 {
     DEBUG(adapterInfo("Initializing the preCICE solver interface..."));
-    clockAdapter_.update();
+    clockValue clock;
+    clock.update();
     timestepPrecice_ = precice_->initialize();
-    timeInInitialize_ = clockAdapter_.elapsed();
+    timeInInitialize_ = clock.elapsed();
 
     preciceInitialized_ = true;
 
@@ -534,9 +538,9 @@ void preciceAdapter::Adapter::initialize()
     }
 
     DEBUG(adapterInfo("Initializing preCICE data..."));
-    clockAdapter_.update();
+    clock.update();
     precice_->initializeData();
-    timeInInitializeData_ = clockAdapter_.elapsed();
+    timeInInitializeData_ = clock.elapsed();
 
     adapterInfo("preCICE was configured and initialized", "info");
 
@@ -550,9 +554,10 @@ void preciceAdapter::Adapter::finalize()
         DEBUG(adapterInfo("Finalizing the preCICE solver interface..."));
 
         // Finalize the preCICE solver interface
-        clockAdapter_.update();
+        clockValue clock;
+        clock.update();
         precice_->finalize();
-        timeInFinalize_ = clockAdapter_.elapsed();
+        timeInFinalize_ = clock.elapsed();
 
         preciceInitialized_ = false;
 
@@ -571,9 +576,10 @@ void preciceAdapter::Adapter::advance()
 {
     DEBUG(adapterInfo("Advancing preCICE..."));
 
-    clockAdapter_.update();
+    clockValue clock;
+    clock.update();
     timestepPrecice_ = precice_->advance(timestepSolver_);
-    timeInAdvance_ += clockAdapter_.elapsed();
+    timeInAdvance_ += clock.elapsed();
 
     return;
 }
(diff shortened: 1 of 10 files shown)


MakisH:
## More data: Perpendicular flap (FSI)

### OpenFOAM-OpenFOAM

Fluid:
- `ExecutionTime = 17.55 s  ClockTime = 20 s`
- `Total time spent in the adapter and preCICE: "00:02.304" (format: day-hh:mm:ss.ms)`
- Time spent in `advance()`: 429ms
- Time spent in the adapter itself: 2304ms-429ms = 1875ms (9% of ClockTime)
   - Potential reason: force calculation 
- [precice-Fluid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9975260/precice-Fluid-events-summary.log)

Solid:
- `ExecutionTime = 0.61 s  ClockTime = 21 s`
- `Total time spent in the adapter and preCICE: "00:17.385" (format: day-hh:mm:ss.ms)`
- Time spent in `advance()`: 17116ms (mostly waiting for data from Fluid)
- Time spent in the adapter itself: 17385ms - 17116ms = 269ms (1% of ClockTime)
- [precice-Solid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9975281/precice-Solid-events-summary.log)

### OpenFOAM-deal.II

Fluid:
- `ExecutionTime = 133.82 s  ClockTime = 141 s`
- `Total time spent in the adapter and preCICE: "01:21.787" (format: day-hh:mm:ss.ms)`, i.e., 81787ms
- Time spent in `advance()`: 74794ms
- Time spent in the adapter itself: 81787ms - 74794ms = 6993ms (5% of ClockTime)
- [precice-Fluid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9975317/precice-Fluid-events-summary.log)

Solid:

```text
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |   141,907s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Advance adapter                 |     1.007 |    43,509s |     30,66% |
| Assemble rhs                    |     1.007 |    22,699s |     16,00% |
| Output results                  |        51 |     7,837s |      5,52% |
| Solve system                    |     1.007 |    65,468s |     46,13% |
+---------------------------------+-----------+------------+------------+
```

- [precice-Solid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9975328/precice-Solid-events-summary.log)

### OpenFOAM-CalculiX

Fluid:
- `ExecutionTime = 263.55 s  ClockTime = 264 s`
- `Total time spent in the adapter and preCICE: "03:34.402" (format: day-hh:mm:ss.ms)`, i.e., 214402ms
- Time spent in `advance()`: 209312ms
- Time spent in the adapter itself: 214402ms - 209312ms = 5090ms (2% of ClockTime)
- [precice-Fluid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9975401/precice-Fluid-events-summary.log)

Solid:
- `Total CalculiX Time: 267.859203`
- [precice-Solid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9975408/precice-Solid-events-summary.log)

davidscn (commented):
I think it would make sense to have a separate timer for the initialization and the online phase. Also it would be nice to exclude the `precice->advance` call. All other calls are anyway somewhat part of the adapter.

MakisH:
Do you prefer splitting this to multiple timers before merging, or merge this overall timer and reconsidering more timers later?

I can try sketching something already.

MakisH:
I added more timers and I think this now covers quite a wide and interesting set of regions. I also realized that I was double-counting the time for setup/initialization in the previous implementation, as I was timing both the constructor and the called `read()` method (now I only time the `read()`).

Running the flow over a heated plate now gives:

## Fluid

```text
-------------------- preCICE adapter timers (primary rank) --------------------
Total time spent in the adapter and preCICE: "00:00.814" (format: day-hh:mm:ss.ms)
  Time spent in the adapter and preCICE only for setting up: "00:00.256" (read() function)
  Time spent in the adapter and preCICE only for iterations: "00:00.558" (execute() and adjustTimeStep() functions)
Time spent exclusively in preCICE: "00:00.575"
  Time spent exclusively in preCICE for initialize(): "00:00.245" (part of setup time)
  Time spent exclusively in preCICE for initializeData(): "00:00" (part of setup time)
  Time spent exclusively in preCICE for advance(): "00:00.280" (part of iterations time)
  Time spent exclusively in preCICE for finalize(): "00:00.048" (part of iterations time)
Time in initialization, advance, and iterations includes time waiting for other participants.
Time in preCICE setMeshVertives, read/write, and state queries is considered as time spent in the adapter.
See the precice-<participant>-events-summary.log for details in time spent in preCICE itself.
-------------------------------------------------------------------------------
```

A bit above, the time spent inside the Interface class is also printed:

```text
---[preciceAdapter] Time spent inside the Interface class for creating: 00:00, for reading: 00:00.008, and for writing: 00:00.001 (mesh Fluid-Mesh).
```

OpenFOAM reports `ExecutionTime = 19.77 s  ClockTime = 23 s`

Looking at the data here, what worries me is that from the 558ms spent in iterating (`execute()` and `adjustTimeStep()`), 280ms is spent in preCICE and almost no time seems to be spent in read/write. I assume that the rest of 278ms comes from checkpointing. Maybe I should also add a timer for that.

## Solid

```text
-------------------- preCICE adapter timers (primary rank) --------------------
Total time spent in the adapter and preCICE: "00:19.911" (format: day-hh:mm:ss.ms)
  Time spent in the adapter and preCICE only for setting up: "00:00.319" (read() function)
  Time spent in the adapter and preCICE only for iterations: "00:19.592" (execute() and adjustTimeStep() functions)
Time spent exclusively in preCICE: "00:19.987"
  Time spent exclusively in preCICE for initialize(): "00:00.312" (part of setup time)
  Time spent exclusively in preCICE for initializeData(): "00:00" (part of setup time)
  Time spent exclusively in preCICE for advance(): "00:19.584" (part of iterations time)
  Time spent exclusively in preCICE for finalize(): "00:00.090" (part of iterations time)
Time in initialization, advance, and iterations includes time waiting for other participants.
Time in preCICE setMeshVertives, read/write, and state queries is considered as time spent in the adapter.
See the precice-<participant>-events-summary.log for details in time spent in preCICE itself.
-------------------------------------------------------------------------------
```

and

```text
---[preciceAdapter] Time spent inside the Interface class for creating: 00:00, for reading: 00:00, and for writing: 00:00 (mesh Solid-Mesh).
```

## preCICE

- [precice-Fluid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9997957/precice-Fluid-events-summary.log)
- [precice-Solid-events-summary.log](https://github.com/precice/openfoam-adapter/files/9997959/precice-Solid-events-summary.log)

MakisH:
I added some more timers for checkpointing and I moved the read/write times to the Adapter class itself. I also replaced the "adapter clock" with local clocks, to avoid any unintended clock resets by inner regions, and polished the formatting. It now looks like this:

```text
-------------------- preCICE adapter timers (primary rank) --------------------------
Total time in adapter + preCICE: "00:00.830" (format: day-hh:mm:ss.ms)
  For setting up (S):            "00:00.186" (read() function)
  For iterations (I):            "00:00.643" (execute() and adjustTimeStep() functions)

Time exclusively in the adapter: "00:00.108"
  (S) reading preciceDict:       "00:00"
  (S) constructing preCICE:      "00:00.007"
  (S) setting up the interfaces: "00:00"
  (S) setting up checkpointing:  "00:00"
  (I) writing data:              "00:00.001"
  (I) reading data:              "00:00.008"
  (I) writing checkpoints:       "00:00.039"
  (I) reading checkpoints:       "00:00.057"
  (I) writing OpenFOAM results:  "00:00.246" (at the end of converged time windows)

Time exclusively in preCICE:     "00:00.467"
  (S) initialize():              "00:00.178"
  (S) initializeData():          "00:00"
  (I) advance():                 "00:00.238"
  (I) finalize():                "00:00.050"
  These times include time waiting for other participants.
  See also precice-<participant>-events-summary.log.
-------------------------------------------------------------------------------------
```

Now all numbers seem to add up pretty closely: we are missing 1ms out of the 186ms for the setup time and 4ms out of the 643ms for iterations time. This is probably to a large extent due to rounding/cut-off errors.

Through this process, I discovered that a significant part of the time measured here is due to this workaround:

https://github.com/precice/openfoam-adapter/blob/552c882deb4767d15bcc85da004444b213aa32bb/Adapter.C#L435-L452

I think this is now ready and worth considering for v1.2.0.

MakisH:
Let's make this configurable at compile time and switch it off by default.

MakisH:
I made the timers configurable at compile time and documented the option.
I tested both with the timing mode enabled and disabled, and I also checked with timing+debug modes enabled.

I could have probably made the implementation more beautiful with some macro, but since we also time regions of multiple lines of code, I did not want to look for any potential side-effects (macros can be nasty).

preciceAdapterFunctionObject.C
```diff
@@ -65,31 +66,70 @@ Foam::functionObjects::preciceAdapterFunctionObject::preciceAdapterFunctionObjec
 
 Foam::functionObjects::preciceAdapterFunctionObject::~preciceAdapterFunctionObject()
 {
+    #ifdef ADAPTER_TIMING_MODE
+    Info << "-------------------- preCICE adapter timers (primary rank) --------------------------" << nl;
+    Info << "Total time in adapter + preCICE: " << timeInAll_.str() << " (format: day-hh:mm:ss.ms)" << nl;
+    Info << "  For setting up (S):            " << timeInSetup_.str() << " (read() function)" << nl;
+    Info << "  For iterations (I):            " << timeInExecute_.str() << " (execute() and adjustTimeStep() functions)" << nl << nl;
```
davidscn, line 73: Is it a cumulative time or a per iteration time?

Adapter.C
```diff
@@ -23,6 +23,10 @@ bool preciceAdapter::Adapter::configFileRead()
     // See also comment in preciceAdapter::Adapter::configure().
     try
     {
+#ifdef ADAPTER_TIMING_MODE
```
davidscn, line 26: The `ifdef`s all over the place are actually a bit distracting. I'm not sure whether it's worth the saving in timing(?). Maybe we can find a less nosiy solution (e.g. using macros) or just add the `ifdef`s for the final cout.

preciceAdapterFunctionObject.C
davidscn:
```suggestion
    Info << "  For all iterations (I):            " << timeInExecute_.str() << " (execute() and adjustTimeStep() functions)" << nl << nl;
```
or maybe remove `For iterations` and `For setting up` completely as the categories make as cumulative quantities no real difference. Or maybe we select a different word than `iterations` such as `time loop`, which is in itself already cumulative.

Allwmake
```diff
@@ -3,7 +3,9 @@ set -e -u
 
 ########## CONFIGURATION - You may want to change these ########################
 
-# Optional: Preprocessor flags ("-DADAPTER_DEBUG_MODE" enables debug messages)
+# Optional: Preprocessor flags
+# "-DADAPTER_DEBUG_MODE" enables debug messages
+# "-DADAPTER_TIMING_MODE" enables time measurements
```
davidscn, line 8: From the user perspective, this sounds a bit like the adapter is doing something differently ('mode'), but it is actually doing the same, just measuring. Maybe better use
```suggestion
# "-DADAPTER_ENABLE_TIMINGS" enables time measurements
```

davidscn (commented):
Not really a strong opinion on this. Printing information on the primary rank only seems somewhat arbitrary in parallel runs, but maybe that's fine for now.

Allwmake
MakisH: Good point. Renamed everywhere.

davidscn (approved):
Only one unresolved conversation, feel free to merge.

PR #262, opened 2022-11-21
Merge `initializeData` into `initialize`

davidscn:
Related to https://github.com/precice/precice/pull/1350 and part of https://github.com/precice/openfoam-adapter/issues/235


MakisH (approved):
The code makes sense, thanks! Have you tested that it still initializes data correctly?

docs/changelog-entries/262.md
```diff
@@ -0,0 +1 @@
+- Merged `initializeData` into `initialize` [#262](https://github.com/precice/openfoam-adapter/pull/262).
```
MakisH, line 1: This should be in `changelog-entries/` (root directory), which is ignored by the linter.

Issue #263, opened 2022-11-23
Adapter requires C++14


davidscn:
```bash
FSI/Force.C:22:29: error: 'make_unique' is not a member of 'std'
   22 |         ForceOwning_ = std::make_unique<volVectorField>(volVectorField(
      |                             ^~~~~~~~~~~
FSI/Force.C:22:29: note: 'std::make_unique' is only available from C++14 onwards
```

I just realized that the c++ standard depends on the OpenFOAM build. On the cluster (e.g. SuperMUC) using v2006, they used C++11. Although I really prefer the `make_unique` style, I will just add a C++11 compatible version using 'new'

Related tp #260

davidscn:
If your OpenFOAM build uses cpp11, then we are bound to the OpenFOAM Cpp standard or we have to rebuild OpenFOAM on the cluster, which seems tedious.

MakisH:
I think that this message is only due to the default options of OpenFOAM, which WMake uses. We should still be able to link a C++11 OpenFOAM to a C++14 adapter.

What I meant was specifying a different standard, e.g., by setting `-std=c++14`) as an additional compiler flag. But this is a discussion for another issue.

davidscn:
This was closed as part of #264

PR #264, opened 2022-11-23
Switch to 'new' instead of make_shared

davidscn:
In order to stick to C++11. Related to #263.

### DIFF
--- a/FSI/Force.C
+++ b/FSI/Force.C
@@ -19,7 +19,7 @@ preciceAdapter::FSI::Force::Force(
     }
     else
     {
-        ForceOwning_ = std::make_unique<volVectorField>(volVectorField(
+        ForceOwning_.reset(new volVectorField(
             IOobject(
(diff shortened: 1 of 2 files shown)


MakisH:
For OpenFOAM.org, I anyway had to use a `Foam::autoPtr`. Switching to that would be more sustainable: https://github.com/precice/openfoam-adapter/commit/896baaaaaf530b24aa04c897038bc55680f95c3c

MakisH (approved):
I tried the `autoPtr` solution also for v2206, but the interface has changed compared to previous versions. I like this solution better, as it does not depend on OpenFOAM.

Please add a changelog entry. We should then make a bugfix release (and port this change to the version-specific branches).

edit: I fixed the markdown linting issue on develop.

MakisH:
Small fix in the changelog: `make_unique`, not `make_shared.

PR #269, opened 2023-01-03
Release v1.2.2

MakisH:
Makes the adapter compatible with v2212, fixing a bug (unneeded `refCast` in the `SinkTemperature`), as discussed in https://develop.openfoam.com/Development/openfoam/-/issues/2671#note_59586

TODO list:

The CI seems to be failing due to outdated cache, since [a custom build works](https://github.com/precice/openfoam-adapter/actions/runs/3830907520/jobs/6519369726). I will try to fix this.

Edit: There seem to be more similar issues. I need to look deeper into this.


MakisH:
I checked the heat-exchanger tutorial, in which we are using the changed code. I ran the tutorial for ten coupling time windows, converted both OpenFOAM and CalculiX results to VTK, and compared the files with `md5sum`. They are identical.

MakisH:
Looking at more places, we could potentially revisit the type selections in some regions, mainly for reasons of clarity. Since `refCast` does a checked [`dynamic_cast`](https://en.cppreference.com/w/cpp/language/dynamic_cast) under the hood, I don't expect any changes in the results.

One example:

https://github.com/precice/openfoam-adapter/blob/eef7431ebb41a5ba58280a5a02e692c75333d0e7/Interface.C#L213-L235

Here, `refCast` converts a `const Boundary &` to a `const vectorField &`. We later subtract this from a `pointField`. This is not necessarily a problem, as the compiler should check all types. But documenting here in case this ever helps understand any related issues.

I think this release is ready. @davidscn what do you think?

davidscn (approved):
Looks good

PR #270, opened 2023-01-16
Add coupling region to volume coupling

tirgendetwas:
Based on [PR#88](https://github.com/precice/openfoam-adapter/pull/88).

Adds the option to configure one or multiple coupling regions defined by OpenFOAM `cellSets`.

Relevant tutorial case: [volume-coupled flow](https://github.com/precice/tutorials/pull/350)

TODO list:

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -119,16 +119,23 @@ bool preciceAdapter::Adapter::configFileRead()
                         DEBUG(adapterInfo("      - " + patch));
                     }
 
-                    // TODO cellSets don't have to be specified?
-                    // should specifying cellSets only work for locationType volume?
-                    // should we allow *not* specifying cellSets for locationType volume (and assume it's the whole domain)
-                    // + not associating them to patches?
+                    // TODO: check if this works correctly
                     DEBUG(adapterInfo("    cellSets      : "));
-                    auto cellSets = interfaceDict.get<wordList>("cellSets");
+                    auto cellSets = interfaceDict.lookupOrDefault<wordList>("cellSets", wordList());
+                    
                     for (auto cellSet : cellSets)
                     {
                         interfaceConfig.cellSetNames.push_back(cellSet);
                         DEBUG(adapterInfo("      - " + cellSet));
+                    }
+
+                    if (!interfaceConfig.cellSetNames.empty() && interfaceConfig.locationsType != "volume")
+                    {
+                        DEBUG(adapterInfo("Cell sets are not supported for locationType != volume. \n"
+                                          "Please configure the desired interface with the locationsType volume. \n"
+                                          "Have a look in the adapter documentation for detailed information.",
+                                          "warning"));
+                        return false;
                     }
                     
                     DEBUG(adapterInfo("    writeData    : "));
--- a/Adapter.C
+++ b/Adapter.C
@@ -101,9 +101,9 @@ bool preciceAdapter::Adapter::configFileRead()
                     // By default, assume that no mesh connectivity is required (i.e. no nearest-projection mapping)
                     interfaceConfig.meshConnectivity = interfaceDict.lookupOrDefault<bool>("connectivity", false);
                     // Mesh connectivity only makes sense in case of faceNodes, check and raise a warning otherwise
-                    if (interfaceConfig.meshConnectivity && interfaceConfig.locationsType == "faceCenters")
+                    if (interfaceConfig.meshConnectivity && (interfaceConfig.locationsType == "faceCenters" || interfaceConfig.locationsType == "volume"))
                     {
-                        DEBUG(adapterInfo("Mesh connectivity is not supported for faceCenters. \n"
+                        DEBUG(adapterInfo("Mesh connectivity is not supported for faceCenters or volume. \n"
                                           "Please configure the desired interface with the locationsType faceNodes. \n"
                                           "Have a look in the adapter documentation for detailed information.",
                                           "warning"));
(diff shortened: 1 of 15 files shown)


CHT/Temperature.C
```diff
@@ -21,10 +21,27 @@ void preciceAdapter::CHT::Temperature::write(double* buffer, bool meshConnectivi
...
+        {
+            for (uint j = 0; j < cellSetNames_.size(); j++) {
+                cellSet overlapRegion(T_->mesh(), cellSetNames_[j]);
+	            const labelList & cells = overlapRegion.toc();
+
+                for( int i=0; i < cells.size(); i++)
+                {
+                    // Copy temperature into the buffer
+                    buffer[bufferIndex++] = T_->ref()[cells[i]];
+                }
```
MakisH, line 41: Could we simplify by using a ranged for loop here?
This question should probably also refer to the `forAll` in the empty cellSet case.

CHT/Temperature.C
```diff
@@ -21,10 +21,27 @@ void preciceAdapter::CHT::Temperature::write(double* buffer, bool meshConnectivi
...
         {
-            buffer[bufferIndex++] = T_->ref()[i];
+            forAll(T_->ref(), i)
+            {
+                buffer[bufferIndex++] = T_->ref()[i];
+            }
+        }
+        else
+        {
+            for (uint j = 0; j < cellSetNames_.size(); j++) {
```
MakisH, line 33: So, this should also work for multiple cell sets, without the one affecting the other, correct? It would be nice to showcase at least two cell sets in the respective tutorial.

Interface.C
```diff
@@ -290,8 +293,24 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh)
...
+        if (!cellSetNames_.empty()){
+            // For every cellSet that participates in the coupling
+            for (uint j = 0; j < cellSetNames_.size(); j++)
+            {
+                // Create a cell set
+                cellSet overlapRegion(mesh, cellSetNames_[j]);
+
+                // Add the cells ID's to the vector and count how many overlap cells does the interface has
+                overlapCells.push_back(overlapRegion.toc());
+                numDataLocations_ += overlapCells[j].size();
```
MakisH, line 308: Just checking/trying to understand: What is the structure of `overlapCells`? What does `overlapRegion.toc()` return?

I understand that this part of the code tries to identify the number of cells of the mesh overlapping with each cellSet. This is needed to identify the number of coupling data locations. Correct?

FF/Pressure.C
```diff
@@ -18,11 +18,27 @@ void preciceAdapter::FF::Pressure::write(double* buffer, bool meshConnectivity,
...
         {
-            buffer[bufferIndex++] = p_->ref()[i];
+            forAll(p_->ref(), i)
+            {
+                buffer[bufferIndex++] = p_->ref()[i];
+            }
         }
+        else
+        {
+            for (auto cellSetName : cellSetNames_) {
```
MakisH, line 30: Here, and in similar places where you iterate over a container to only read values, you could default to `const auto &`, for safety and performance reasons (at least in the parts of the code called in every coupling time window).

```suggestion
            for (const auto & cellSetName : cellSetNames_) {
```

Adapter.C
```diff
@@ -118,6 +118,25 @@ bool preciceAdapter::Adapter::configFileRead()
                         DEBUG(adapterInfo("      - " + patch));
                     }
 
+                    // TODO: check if this works correctly
```
MakisH, line 121: Does it?

MakisH (commented):
Some comments, mainly for me to get an overview of the current state.
No need to rush implementing, but please have a look at previous conversations and mark any completed/outdated as "resolved".

Interface.H
```diff
@@ -65,6 +68,7 @@ public:
         std::string meshName,
         std::string locationsType,
         std::vector<std::string> patchNames,
+        std::vector<std::string> cellSetNames,
```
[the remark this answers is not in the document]
tirgendetwas: What do you mean? Should I add something specific to the FSI module?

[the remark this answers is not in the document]
MakisH: I am not sure anymore what I had in mind at that point, but it looks like you don't need to do anything related to the FSI module.

Interface.C
tirgendetwas: Yes, correct. `toc()` returns the keys of a `HashTable` in a list, in this case the cellIDs of the cells in the given `cellSet` (`cellSet` inherits from `HashTable`). So we add the number of cellIDs in this `cellSet` to the total number of data locations. 

I think `overlapCells` is actually unnecessary and we could directly do `numDataLocations_ += overlapRegion.toc().size()` without going through the vector.

FF/FF.C
```diff
@@ -161,6 +161,13 @@ bool preciceAdapter::FF::FluidFluid::addReaders(std::string dataName, Interface*
             new Velocity(mesh_, nameU_));
         DEBUG(adapterInfo("Added reader: Velocity."));
     }
+    else if (dataName.find("U_vol") == 0)
```
MakisH, line 164: Is it documented somewhere why we need the additional `U_vol`, while `p_rgh` is just as we take it from OpenFOAM?

Interface.C
```diff
@@ -324,11 +327,32 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh, const std::str
     }
     else if (locationType_ == LocationType::volumeCenters)
     {
+
+        // TODO is volume option added everywhere it's supposed to
```
MakisH, line 331: Reminder to check

Interface.C
tirgendetwas: (`overlapCells` is actually used in the code later, so I have left it in)

MakisH:
The CI is again broken because the hashes of the OpenFOAM packages are outdated. The code builds normally on my machine. Merging this to the `volume-coupling` and continuing there.
Thanks a lot! :rocket: :smile:

MakisH (approved):

PR #280, opened 2023-02-18
Update Actions set-output syntax

MakisH:
Based on the deprecation described in https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Triggered a [custom build](https://github.com/precice/openfoam-adapter/actions/runs/4213332611) to check.

TODO list:

- I added a changelog entry in `changelog-entries/` (create directory if missing) -> Not user-facing or of any other significance.

### DIFF
--- a/.github/workflows/build-custom.yml
+++ b/.github/workflows/build-custom.yml
@@ -82,69 +82,69 @@ jobs:
           OpenFOAMv2212)
             wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | sudo bash
             sudo apt-get install openfoam2212-dev
-            echo "::set-output name=openfoam_exec::/usr/bin/openfoam2212";;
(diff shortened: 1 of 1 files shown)


MakisH:
A different CI workflow now fails to find the package `openfoam2212-dev`. This looks completely unrelated, yet we should find out why this happens now.

Having a first look at the [documentation](https://develop.openfoam.com/Development/openfoam/-/wikis/precompiled/debian#ubuntu) and the respective PPA repository, I could not really understand if anything changed recently.

MakisH:
The action I triggered seems to find the `openfoam2212-dev` package, but it seems to fail at what I actually wanted to update: https://github.com/precice/openfoam-adapter/actions/runs/4213332611/jobs/7312975116

I don't understand why at the moment.

davidscn (approved):
Yes, the failing tests seems unrelated.

PR #284, opened 2023-03-20
Add a citation file

davidscn:
.. which would allow to directly extract a bibtex from the main repo.

I followed the instructions given here: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files


MakisH (approved):

CITATION.cff
```diff
@@ -0,0 +1,37 @@
...
+- family-names: "Schneider"
+  given-names: "David"
+  orcid: "https://orcid.org/0000-0002-3487-9688"
+- family-names: "Uekermann"
+  given-names: "Benjamin"
+  orcid: "https://orcid.org/0000-0002-1314-9969"
+title: "OpenFOAM-preCICE: Coupling OpenFOAM with External Solvers for Multi-Physics Simulations"
+version: 1.1.0
+doi: 10.51560/ofj.v3.88
+date-released: 2022-02-08
```
davidscn, line 16: @MakisH do you also understand this as the release date of the software? It was only my best guess.

MakisH: If we are talking specifically about version 1.1.0, then yes.
But I have no idea how citation files are meant to be used. Should we specify:
- first publication on GitHub?
- v1.0.0?
- the version used in some academic paper? (as we do here)
- the latest version, and update it all the time, referring to the repository?

I assume that, since we are anyway referring to a paper, it is appropriate to refer to v1.1.0 and its release date.

Is there any CFF documentation on this?

davidscn: All I read/found is in the above GitHub link.

PR #266, opened 2022-12-09
Replace actions with explicit calls

davidscn:
TODO list:

- [ ] I updated the documentation in `docs/`
- [ ] I added a changelog entry in `changelog-entries/` (create directory if missing)

Issue #258, opened 2022-11-15
Reimplement connectivity definition


davidscn:
https://github.com/precice/openfoam-adapter/blob/2a51e6e24bc36776051a06566413bc66cbe181bd/Interface.C#L309-L316

See https://github.com/precice/precice/issues/1248#issuecomment-1103875457, we need to find a new solution here.

MakisH:
To summarize:

## Current implementation

The affected function is `preciceAdapter::Interface::configureMesh` in [Interface.C](https://github.com/precice/openfoam-adapter/blob/v1.2.0/Interface.C#L249-L320). What this does is:

1. Gets a list of faces and coordinates at the interface:

https://github.com/precice/openfoam-adapter/blob/v1.2.0/Interface.C#L271-L273

2. Modifies the coordinates to account for any deformation
3. Uses `Foam::triPointRef Foam::tetIndices::faceTri(const polyMesh& mesh) const` to triangulate each face and get back a triangle:

https://github.com/precice/openfoam-adapter/blob/v1.2.0/Interface.C#L293

4. Uses the preCICE `getMeshVertexIDsFromPositions` to populate an array of preCICE vertex IDs for each triangle (`triVertIDs`):

https://github.com/precice/openfoam-adapter/blob/v1.2.0/Interface.C#L309

6. Passes each triangle edge to the preCICE `setMeshTriangleWithEdges`:

https://github.com/precice/openfoam-adapter/blob/v1.2.0/Interface.C#L316

## What changes in preCICE v3

The discussion in https://github.com/precice/precice/issues/1248 concluded in removing `getMeshVertexIDsFromPositions`, since the implementation is complicated and unreliable. We need to directly give vertices to preCICE.

There has also been a series of changes, most recently https://github.com/precice/precice/pull/1322:

Respective changelog entry:

> - Removed edge IDs from the API: `setMeshEdge` now returns nothing, all `setMeshXWithEdges` variants were removed from the API, and `setMeshTriangle|Quad` now accepts vertex IDs instead of edge IDs.
> - Added efficient bulk functions `setMeshEdges|Triangles|Quads|Tetrahedra` for setting connectivity based on an array of vertex IDs.
> - Changed API to guarantee the creation of hierarchical connectivity. For example, `setMeshTriangle` now creates necessary edges internally.

This means that we can no longer give locations to `setMeshTriangleWithEdges` (+ renaming to `setMeshTriangle`).

[...]

PR #285, opened 2023-03-20
Port the adapter to preCICE v3

davidscn:
This PR includes most of the changes required to port the adapter to use the upcoming preCICE v3.

Depends on #266 and implements https://github.com/precice/precice/issues/1471

Follow-up regarding connectivity: #297.

Related to #235.

TODO list:


Adapter.C
```diff
@@ -462,7 +459,8 @@ void preciceAdapter::Adapter::execute()
     ACCUMULATE_TIMER(timeInWriteResults_);
 
     // Read the received coupling data from the buffer
-    readCouplingData();
+    // Fits to an implicit Euler
+    readCouplingData(runTime_.deltaT().value());
```
MakisH, line 463: Rubber-ducking, to cross-check that this is what you intended:

1. OpenFOAM computes a new time step size (which may be modified by the adapter) (let's call it **dt1**)
2. OpenFOAM computes a new solution with this time step size
3. The adapter writes data, calls `advance()`, and takes care of checkpointing
4. In case there is a fixed time step size, the adapter resets it to the value from the configuration (let's call it **dt2**)
5. The adapter reads data at a step **dt1** (if dynamic dt) or **dt2** (if fixed dt).

Sounds correct for the dynamic time step size, but we may need to check what happens at a fixed time step size. We may be introducing (or already had) a bug here.

Adapter.H
```diff
@@ -106,7 +105,7 @@ private:
     std::vector<Interface*> interfaces_;
 
     //- preCICE solver interface
-    precice::SolverInterface* precice_ = NULL;
+    precice::Participant* precice_ = NULL;
```
MakisH, line 108: Since we are changing this, let's directly make it nullptr:

```suggestion
    precice::Participant* precice_ = nullptr;
```

docs/changelog-entries/262.md
```diff
@@ -0,0 +1 @@
+- Merged `initializeData` into `initialize` [#262](https://github.com/precice/openfoam-adapter/pull/262).
```
MakisH, line 1: Let's update this to something like "Migrated the adapter to the preCICE v3 API" and also update the PR title and description, to include all related changes.

MakisH (commented):
Very nice, thanks for all the effort!

Can we now make this PR the central/single PR that ports the adapter to v3? This means changing the title and description. It is currently a bit lost like this.

The only thing missing seems to be the connectivity: (https://github.com/precice/openfoam-adapter/issues/258)

```
Interface.C: In member function ‘void preciceAdapter::Interface::configureMesh(const Foam::fvMesh&, const string&, const string&)’:
Interface.C:306:26: error: ‘class precice::Participant’ has no member named ‘getMeshVertexIDsFromPositions’
  306 |                 precice_.getMeshVertexIDsFromPositions(meshID_, faceField.size() * (triaPerQuad * nodesPerTria), triCoords, triVertIDs);
      |                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Interface.C:313:30: error: ‘class precice::Participant’ has no member named ‘setMeshTriangleWithEdges’; did you mean ‘setMeshTriangles’?
  313 |                     precice_.setMeshTriangleWithEdges(meshID_, triVertIDs[facei * nodesPerTria], triVertIDs[facei * nodesPerTria + 1], triVertIDs[facei * nodesPerTria + 2]);
      |                              ^~~~~~~~~~~~~~~~~~~~~~~~
      |                              setMeshTriangles
```

MakisH:
Since other projects depend on porting the OpenFOAM adapter to develop (@ezonta, @tirgendetwas), we can also temporarily work around the issue by removing the calls to the removed methods (which means that the tutorials relying on nearest-projection mapping would give different results). The [`workaround-connectivity-v3`](https://github.com/precice/openfoam-adapter/tree/workaround-connectivity-v3) branch compiles with preCICE develop. The flow-over-heated-plate tutorial (with config updates) completes.

davidscn:
> Since other projects depend on porting the OpenFOAM adapter to develop (@ezonta, @tirgendetwas), we can also temporarily work around the issue by removing the calls to the removed methods (which means that the tutorials relying on nearest-projection mapping would give different results)

Yes. Right now, the nearest-projection configuration in the adapter has a mediocre usability at a debatable feature gain, so I would consider it rather low priority.

Adapter.C
MakisH: @davidscn any comment here?

MakisH (approved):
@davidscn I would suggest that we merge this already, so that we rebase and move on with #297 and #298. Feel free to merge this one if you feel it is ready.

The CI currently fails because it is using preCICE v2.5.0. Let's fix this in a separate PR.

PR #289, opened 2023-05-19
Fix iterator in FSI reader and writer for multiple patches

davidscn:
The current FSI reader (Force/Stress/Displacement/DisplacementDelta) don't behave correctly in case of multiple patches: The previous implementation didn't take the size of the previous patch into account, i.e., was only iterating over the 'beginning' of the read and write buffer.
In most of our cases (actually all I know), we only use one patch per interface mesh, where this is not relevant. However, as soon as multiple patches are combined into one interface mesh, this becomes relevant and produces malicious results. In the CHT module, we do it actually correct.

Might be worth a bugfix release compatible with preCICE v2

FYI @LeonardWilleke

TODO list:

- <del> I updated the documentation in `docs/` </del>

### DIFF
--- a/FSI/Displacement.C
+++ b/FSI/Displacement.C
@@ -48,6 +48,7 @@ void preciceAdapter::FSI::Displacement::write(double* buffer, bool meshConnectiv
 
     if (this->locationType_ == LocationType::faceCenters)
     {
+        int bufferIndex = 0;
         // For every boundary patch of the interface
         for (const label patchID : patchIDs_)
         {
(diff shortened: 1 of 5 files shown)


MakisH (approved):
The changes make sense, and we clearly had a bug before.
Have you tested the fix? If yes, how?

Both the CHT and FF modules are not affected. We should indeed release a bugfix.

davidscn:
>  Have you tested the fix? If yes, how?

Yes, I compared (at least for the forces) the data against the function object results, which produces now correct results. I didn't check the displacement, but it is somewhat obvious that the same applies. Would be necessary to add tests for the coupling physics, though.

PR #291, opened 2023-06-14
Bugfix release v1.2.3

davidscn:
TODO list:

- <del> I added a changelog entry in `changelog-entries/` (create directory if missing) </del>

I cherry-pick this commit to all supported OpenFOAM versions as well. @MakisH please review, I will merge then

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,4 +1,5 @@
 {
     "MD013": false,
-    "MD033": false
+    "MD033": false,
+    "MD034": false
 }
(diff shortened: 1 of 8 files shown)


MakisH:
Please also cherry-pick https://github.com/precice/openfoam-adapter/commit/f91e2047e23e5924a7bf7f5864376750c198fbc7, to fix the CI

MakisH:
See also the release checklist: https://github.com/precice/openfoam-adapter/blob/develop/tools/release_pull_request_template.md

No need to run every manual test, but since at least we need to port this to the Foundation versions, it makes sense.

MakisH:
> > Please also cherry-pick [f91e204](https://github.com/precice/openfoam-adapter/commit/f91e2047e23e5924a7bf7f5864376750c198fbc7), to fix the CI
> 
> Not sure about this, it is not really a bug and not user-affecting, but I can add it.

If not, our repository landing page will look like we are not even looking at our CI, which is misleading.

MakisH (approved):

PR #292, opened 2023-06-14
Adding arguments to the dockerfile, to make it more flexible

valentin-seitz:
This is mainly needed later when we build for the systemtests. 
I chose the variables similar enough to the ones beeing used in the `components.yaml` defining this openfoam-adapter. 

See https://github.com/precice/tutorials/pull/334/files#diff-22b22ed982724ed250125c665add6f391f2aa273571366fbb73c0220ad73cc3e


.github/workflows/Dockerfile
```diff
@@ -1,14 +1,18 @@
...
 
+ARG OPENFOAM_EXECUTABLE=openfoam2112
+ARG OPENFOAM_ADAPTER_REF=develop
 # Install OpenFOAM
-ENV OPENFOAM_VERSION=2112
-RUN wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | sudo bash
-RUN sudo apt-get -qq install openfoam${OPENFOAM_VERSION}-dev
+RUN apt-get update
+RUN wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | bash
+RUN apt-get -qq install $OPENFOAM_EXECUTABLE-dev git
```
MakisH, line 10: I hope this `<name>-dev` package naming is the same also for the Foundation versions (please check).

MakisH, line 10: Shouldn't this be in this order?

```suggestion
RUN wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | bash
RUN apt-get update
RUN apt-get -qq install $OPENFOAM_EXECUTABLE-dev git
```

.github/workflows/Dockerfile
```diff
@@ -1,14 +1,18 @@
...
 
 # Bust the cache: we want that the next steps are always executed
 ARG CACHEBUST
 
 # Build the OpenFOAM adapter
-RUN git clone --depth 1 --branch develop https://github.com/precice/openfoam-adapter.git
+RUN git clone --depth 1 --branch $OPENFOAM_ADAPTER_REF https://github.com/precice/openfoam-adapter.git
 WORKDIR /openfoam-adapter
-RUN /usr/bin/openfoam${OPENFOAM_VERSION} ./Allwmake
+RUN /usr/bin/$OPENFOAM_EXECUTABLE ./Allwmake
```
MakisH, line 18: I would make the whole path part of the `OPENFOAM_EXECUTABLE`. ESI installs in `/usr/bin`, while Foundation installs in `/opt/`. How how we do it in the custom build workflow: https://github.com/precice/openfoam-adapter/blob/develop/.github/workflows/build-custom.yml

.github/workflows/Dockerfile
valentin-seitz: No, but currently we only plan to support that in the CI

valentin-seitz: No i think we need to update the package lists before we can execute the script. The script does not make sure it has the newest package lists and sometimes the apt will fail because package mirrors moved.

.github/workflows/Dockerfile
valentin-seitz: Since we currently only plan to support the ESI version i would leave it as be.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+ARG PRECICE_TAG=latest
+ARG from=precice/precice:${PRECICE_TAG}
+FROM $from
+
+ARG OPENFOAM_EXECUTABLE=openfoam2112
+# Install OpenFOAM
+RUN apt-get update &&\
```
MakisH, line 7: Document why we need the update in a comment

tools/docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    wget -q -O - https://dl.openfoam.com/add-debian-repo.sh | bash &&\
+    apt-get -qq install $OPENFOAM_EXECUTABLE-dev git
```
MakisH, line 9:
```suggestion
    apt-get -qq install ${OPENFOAM_EXECUTABLE}-dev git
```

tools/docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+ARG OPENFOAM_ADAPTER_REF=develop
+# Build the OpenFOAM adapter
+RUN git clone --depth 1 --branch $OPENFOAM_ADAPTER_REF https://github.com/precice/openfoam-adapter.git
```
MakisH, line 13:
```suggestion
RUN git clone --depth 1 --branch ${OPENFOAM_ADAPTER_REF} https://github.com/precice/openfoam-adapter.git
```

tools/docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+WORKDIR /openfoam-adapter
+RUN /usr/bin/$OPENFOAM_EXECUTABLE ./Allwmake &&\
```
MakisH, line 15:
```suggestion
RUN /usr/bin/${OPENFOAM_EXECUTABLE} ./Allwmake &&\
```

MakisH (approved):

PR #299, opened 2023-08-09
Switch CI to build preCICE from develop

MakisH:
Hoping that it will now start succeeding.

TODO list:


MakisH:
@davidscn the CI now succeeds and this is ready for review.

I do not want to also change the `build-custom.yml` workflow now, I think we should only do this if needed.

davidscn (commented):
Hm, what this workflow does is essentially encapsulated in the [setup precice](https://github.com/precice/setup-precice-action) workflow. There are also future plans to extend the action using caching etc.

Apart from the feature side of development, the syntax would simplify a lot, maybe even making the whole workflow (as a separate file) too much.

MakisH:
@davidscn good point, I forgot about the `setup-precice-action`!

I now replaced the custom script with that action, renamed `install-dependencies.sh` to `install-openfoam.sh`, and bumped `checkout` and `cache` to v3.

PR #305, opened 2023-08-17
Pass the correct number of entries to `writeData`

### LINKED ISSUE #303: Cannot run fluid-fluid tutorials with preCICE develop
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Ubuntu 22.04_
preCICE Version: f7289a67

**Describe the problem**

Trying to merge `develop` into `ff-develop`, and while running some tutorials, I noticed the following output:

```
---[preciceAdapter] Loaded the OpenFOAM-preCICE adapter - v1.2.3 + unreleased changes.
---[preciceAdapter] Reading preciceDict...
---[precice]  This is preCICE version 3.0.0
---[precice]  Revision info: v2.5.0-413-gf7289a67
---[precice]  Build type: Debug
---[precice]  Configuring preCICE with configuration "../precice-config.xml"
---[precice]  I am participant "Fluid2"
---[precice] ERROR:  Input sizes are inconsistent attempting to write 1D data "Pressure" to mesh "Fluid2-Mesh". You passed 41 vertices and 82 data components, but we expected 41 data components (1 x 41).
```

This comes from `tutorials/flow-over-heated-plate-partitioned-flow/fluid2-openfoam`.

Similarly from `tutorials/partitioned-backwards-facing-step/fluid1-openfoam`:

```
---[precice] ERROR:  Input/Output sizes are inconsistent attempting to read 1D data "Pressure" from mesh "Fluid1-Mesh". You passed 32 vertices and 96 data components, but we expected 32 data components (1 x 32).
```

Affected tutorials:

- flow-over-heated-plate-partitioned-flow
- partitioned-backwards-facing-step

Potentially affected:

- partitioned-pipe

Not affected:

- flow-over-heated-plate
- perpendicular-flap

**Additional context**

The tutorial `partitioned-pipe` does not throw this error, but instead runs into an assertion:

```
---[preciceAdapter] Loaded the OpenFOAM-preCICE adapter - v1.2.3 + unreleased changes.
---[preciceAdapter] Reading preciceDict...
---[precice]  This is preCICE version 3.0.0
---[precice]  Revision info: v2.5.0-413-gf7289a67
---[precice]  Build type: Debug
---[precice]  Configuring preCICE with configuration "../precice-config.xml"
---[precice]  I am participant "Fluid1"
---[precice]  Setting up primary communication to coupling partner/s
---[precice]  Primary ranks are connected
---[precice]  Setting up preliminary secondary communication to coupling partner/s
---[precice]  Prepare partition for mesh Fluid1-Mesh
---[precice]  Gather mesh Fluid1-Mesh
---[precice]  Send global mesh Fluid1-Mesh
---[precice]  Receive global mesh Fluid2-Mesh
---[precice]  Setting up secondary communication to coupling partner/s
---[precice]  Secondary ranks are connected
---[precice]  Computing "nearest-neighbor" mapping from mesh "Fluid2-Mesh" to mesh "Fluid1-Mesh" in "read" direction.
---[precice]  Mapping distance min:0 max:7.17448e-15 avg: 2.20885e-15 var: 4.2847e-30 cnt: 500
---[precice]  Mapping "Pressure" for t=0 from "Fluid2-Mesh" to "Fluid1-Mesh"
---[precice]  Mapping "Pressure" for t=1 from "Fluid2-Mesh" to "Fluid1-Mesh"
ASSERTION FAILED
Location:   void precice::impl::DataContext::mapData()
File:       /home/gc/repos/precice/precice/src/precice/impl/DataContext.cpp:95
Expression: context.fromData->stamples().size() > 0
Rank:       0
Arguments:  none
```

Maybe related to https://github.com/precice/openfoam-adapter/pull/297, to #255, or to https://github.com/precice/precice/pull/1742. I guess that somewhere in the code we are mixing up some data and mesh dimensions, or that in the `Interface.C` we are giving the wrong number of locations.

cc @thesamriel

davidscn:
Closes #303 

I added now a return type to `CouplingDataUser::write()` to tell writeData how many data entries are actually relevant. Adding this return type is also a nice consistency check for the `CouplingDataUser::write()` function and not uncommon for such functions.

TODO list:

### DIFF
--- a/Interface.C
+++ b/Interface.C
@@ -545,12 +545,16 @@ void preciceAdapter::Interface::readCouplingData(double relativeReadTime)
 
         // Make preCICE read vector or scalar data
         // and fill the adapter's buffer
+        std::size_t nReadData = vertexIDs_.size() * precice_.getDataDimensions(meshName_,couplingDataReader->dataName());
+        // We could add a sanity check here
+        // nReadData == vertexIDs_.size() * (1 + (dim_ - 1) * static_cast<int>(couplingDataReader->hasVectorData()));
+
         precice_.readData(
             meshName_,
             couplingDataReader->dataName(),
             vertexIDs_,
             relativeReadTime,
-            dataBuffer_);
+            {dataBuffer_.data(), nReadData});
 
         // Read the received data from the buffer
         couplingDataReader->read(dataBuffer_.data(), dim_);
--- a/Interface.C
+++ b/Interface.C
@@ -571,14 +571,14 @@ void preciceAdapter::Interface::writeCouplingData()
             couplingDataWriter = couplingDataWriters_.at(i);
 
         // Write the data into the adapter's buffer
-        couplingDataWriter->write(dataBuffer_.data(), meshConnectivity_, dim_);
+        auto nWrittenData = couplingDataWriter->write(dataBuffer_.data(), meshConnectivity_, dim_);
 
         // Make preCICE write vector or scalar data
         precice_.writeData(
             meshName_,
             couplingDataWriter->dataName(),
             vertexIDs_,
-            dataBuffer_);
+            {dataBuffer_.data(), nWrittenData});
     }
     // }
 }
--- a/Interface.C
+++ b/Interface.C
@@ -545,7 +545,7 @@ void preciceAdapter::Interface::readCouplingData(double relativeReadTime)
 
         // Make preCICE read vector or scalar data
         // and fill the adapter's buffer
-        std::size_t nReadData = vertexIDs_.size() * precice_.getDataDimensions(meshName_,couplingDataReader->dataName());
+        std::size_t nReadData = vertexIDs_.size() * precice_.getDataDimensions(meshName_, couplingDataReader->dataName());
         // We could add a sanity check here
         // nReadData == vertexIDs_.size() * (1 + (dim_ - 1) * static_cast<int>(couplingDataReader->hasVectorData()));
 
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -18,7 +18,7 @@ preciceAdapter::CHT::HeatFlux::HeatFlux(
     dataType_ = scalar;
 }
 
-void preciceAdapter::CHT::HeatFlux::write(double* buffer, bool meshConnectivity, const unsigned int dim)
+std::size_t preciceAdapter::CHT::HeatFlux::write(double* buffer, bool meshConnectivity, const unsigned int dim)
 {
     int bufferIndex = 0;
 
@@ -68,6 +68,7 @@ void preciceAdapter::CHT::HeatFlux::write(double* buffer, bool meshConnectivity,
             }
         }
     }
+    return bufferIndex;
 }
 
 void preciceAdapter::CHT::HeatFlux::read(double* buffer, const unsigned int dim)
--- a/CHT/HeatFlux.H
+++ b/CHT/HeatFlux.H
@@ -33,7 +33,7 @@ public:
 
     //- Compute heat flux values from the temperature field
     //  and write them into the buffer
-    void write(double* buffer, bool meshConnectivity, const unsigned int dim) final;
+    std::size_t write(double* buffer, bool meshConnectivity, const unsigned int dim) final;
 
     //- Read heat flux values from the buffer and assign them to
     //  the gradient of the temperature field
--- a/CHT/HeatTransferCoefficient.C
+++ b/CHT/HeatTransferCoefficient.C
@@ -20,7 +20,7 @@ preciceAdapter::CHT::HeatTransferCoefficient::HeatTransferCoefficient(
 }
 
 
-void preciceAdapter::CHT::HeatTransferCoefficient::write(double* buffer, bool meshConnectivity, const unsigned int dim)
+std::size_t preciceAdapter::CHT::HeatTransferCoefficient::write(double* buffer, bool meshConnectivity, const unsigned int dim)
 {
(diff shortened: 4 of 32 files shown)


MakisH:
Thanks for the PR! While it goes into the right direction, I still get the following (different) output from `flow-over-heated-plate-partitioned-flow/fluid1-openfoam`:

```
---[preciceAdapter] preCICE was configured and initialized
---[precice] ERROR:  Input/Output sizes are inconsistent attempting to read 1D data "Pressure" from mesh "Fluid1-Fluid-Mesh". You passed 41 vertices and 82 data components, but we expected 41 data components (1 x 41).
```

Similarly for `partitioned-backwards-facing-step/fluid1-openfoam`:

```
---[preciceAdapter] preCICE was configured and initialized
---[precice] ERROR:  Input/Output sizes are inconsistent attempting to read 1D data "Pressure" from mesh "Fluid1-Mesh". You passed 32 vertices and 96 data components, but we expected 32 data components (1 x 32).
```

For the `partitioned-pipe/fluid1-openfoam-pimplefoam`, I still get an assertion failure:

```
---[precice]  Computing "nearest-neighbor" mapping from mesh "Fluid2-Mesh" to mesh "Fluid1-Mesh" in "read" direction.
---[precice]  Mapping distance min:0 max:7.17448e-15 avg: 2.20885e-15 var: 4.2847e-30 cnt: 500
---[precice]  Mapping "Pressure" for t=0 from "Fluid2-Mesh" to "Fluid1-Mesh"
---[precice]  Mapping "Pressure" for t=1 from "Fluid2-Mesh" to "Fluid1-Mesh"
ASSERTION FAILED
Location:   void precice::impl::DataContext::mapData()
File:       /home/gc/repos/precice/precice/src/precice/impl/DataContext.cpp:95
Expression: context.fromData->stamples().size() > 0
Rank:       0
Arguments:  none
```

davidscn:
Had to adjust the buffer reading as well, as reading and writing share the same buffer.

> For the partitioned-pipe/fluid1-openfoam-pimplefoam, I still get an assertion failure:

I get the same, but I don't see how the changes here touch anything in the logical order of reading and writing. Seems to be related to @BenjaminRodenberg's work.

Interface.C
```diff
@@ -545,12 +545,16 @@ void preciceAdapter::Interface::readCouplingData(double relativeReadTime)
 
         // Make preCICE read vector or scalar data
         // and fill the adapter's buffer
+        std::size_t nReadData = vertexIDs_.size() * precice_.getDataDimensions(meshName_,couplingDataReader->dataName());
+        // We could add a sanity check here
+        // nReadData == vertexIDs_.size() * (1 + (dim_ - 1) * static_cast<int>(couplingDataReader->hasVectorData()));
+
```
MakisH, line 551: This should already be a unit test for `getDataDimensions`, right? So, I don't see the added value. I would rather remove it.

```suggestion
        std::size_t nReadData = vertexIDs_.size() * precice_.getDataDimensions(meshName_,couplingDataReader->dataName());

```

MakisH (commented):
The changes make sense, but I cannot yet confirm that it solves the issue. We need to first check if there is an issue in the core library.

```
ASSERTION FAILED
Location:   void precice::impl::DataContext::mapData()
File:       /home/gc/repos/precice/precice/src/precice/impl/DataContext.cpp:95
Expression: context.fromData->stamples().size() > 0
Rank:       0
Arguments:  none
```

davidscn:
> but I cannot yet confirm that it solves the issue.

The exception above is only triggered for the partitioned-pipe. You could try one of the other examples. I wouldn't wait here for changes in precice/precice.

MakisH:
Indeed, the partitioned flow over heated plate works. But the partitioned BFS shows strange results:

Screenshot from 2023-09-13 15-07-17
Screenshot from 2023-09-13 15-07-53

davidscn:
Interesting. First of all, running the case using the current precice@develop and the default configuration, no convergence within individual time-steps can be observed, i.e., we perform 100 implicit iterations within each time-step and move on afterwards.

However, using in the configuration file `<exchange ... substeps="0" />` for all data exchanges shows proper convergence behavior and the result looks fine as well
Screenshot from 2023-09-13 16-49-14

To me, this seems unrelated to the changes here and should go into a dedicated issue either here or in the tutorials repo

uekerman:
For the assertions that you run into, are there already issues in preCICE? They look definitely like edge cases we have not yet covered with the waveforms. We should try to reproduce them with integration tests. Having the config in the issues could already be enough to create these tests.

davidscn:
> @davidscn Why not an issue in the core library? Do you suspect we are reading/writing something wrong?

I was referring to the lacking convergence, not to any assertion we hit. Once we change the substeps default back to `false`, everything runs fine. However, the `substeps=true` option doesn't work and it could be an issue on how this case was set up or some interpolation issue in the waveform. I'm not completely sure and the tutorials seem to be the proper place to discuss.

MakisH:
Opened an issue for the convergence issues with substeps: https://github.com/precice/tutorials/issues/373

BenjaminRodenberg:
Be aware that quasi Newton schemes are currently not supported, if `substeps="true"`. A corresponding error message is on the way https://github.com/precice/precice/pull/1721.

For the FEniCS adapter the partitioned heat equations works with underrelaxation and `subcycling="true"` (slow, but good for testing). See https://github.com/precice/tutorials/pull/281. Do you have an OpenFOAM case with underrelaxation that you could use for testing the new subcycling feature with your adapter?

MakisH (approved):
@davidscn all previously failing tutorials seem to work now. Feel free to merge.

PR #308, opened 2023-11-17
Add phase variable and flux coupling for interFOAM in the FF module

thesamriel:
TODO list:

- Added coupling variables for `Alpha` and `AlphaGradient`, which can be used for coupling of `interFoam`.
- Added coupling variable `Phi`, which couples the face flux on the surface field `phi`.

### Coupling of `Alpha`:
Coupling of interFoam solvers does not work perfectly. Below screenshots show a very basic test scenario of a laminar flow through a 2-D pipe with the top half being air and the bottom half being incoming water. Gravity is set to 0 to avoid the creation of vortices. The top image always shows the monolithic simulation, the bottom one the coupled simulation.

| Time |  Alpha | Pressure  |
|---|---|---|
|5s|Selection_091 | Selection_092  |
| 9.5s|Selection_090 | Selection_087 |
 |20s| Selection_089| Selection_088 |

Coupling of the phase variable `Alpha` works with Dirichlet-Neumann coupling, however there are some flow artefacts when the air-water-interface crosses the coupling interface. However, I suspect the problem to be not in the coupling perse, but rather in the interFoam outlet boundary conditions. This is because, the fluid also behaves strange at the outlet of the monolithic version.

### Coupling of `Phi`:
Coupling was tested by adding it to normal FF-coupling. I also used it for better stability in the[ breaking dam scenario presented at Coupled Conference](https://www.researchgate.net/publication/375534619_Partitioned_Flow_Simulations_with_preCICE_and_OpenFOAM).
I don't have/know a proper scenario, where coupling of `Phi` can be demonstrated. Nonetheless, the functionality is there and might be useful to some users.

### Test case:
The interfoam pipe test case from above can be found on the [interfoam-pipe](https://github.com/precice/tutorials/tree/interfoam-pipe/interfoam-pipe) branch of the tutorials repository.
It also includes the option to additionally couple phi, which will change the results marginally.
Note how the water is held back or reflected at the outlet as well. Figuring out the correct outlet boundary condition might help improve the coupling as well. Found a discussion on this [here](https://www.cfd-online.com/Forums/openfoam-solving/95524-outlet-boundary-condition-interfoam.html), but not a proper solution.


MakisH:
Thanks for the PR, @thesamriel! Nice to finally have it here!
Could you please also add a testcase in the tutorials?

docs/config.md
```diff
@@ -217,7 +217,7 @@ solver              displacementLaplacian;
 
 #### FF
 
-The fluid-fluid coupling module supports reading and writing `Pressure`, `Velocity`, `PressureGradient`, `VelocityGradient`, `FlowTemperature` and `FlowTemperatureGradient`.
+The fluid-fluid coupling module supports reading and writing `Pressure`, `Velocity`, `PressureGradient`, `VelocityGradient`, `FlowTemperature`, `FlowTemperatureGradient`, `Alpha`, `AlphaGradient` and the face flux `Phi`.
```
MakisH, line 220: Nice! We also need such a mention in the docs README: https://github.com/precice/openfoam-adapter/blob/alpha-coupling/docs/README.md

FF/Alpha.C
```diff
@@ -0,0 +1,116 @@
...
+                const labelList& cells = overlapRegion.toc();
+
+                for (const auto& currentCell : cells)
+                {
+                    // Copy the alpha valus into the buffer
+                    buffer[bufferIndex++] = Alpha_->internalField()[currentCell];
+                }
+            }
+        }
+    }
```
MakisH, line 42: Very nice that you are also considering the volume coupling here! :hugs:

MakisH (approved):
The changes make sense to me, I would just like to try it before merging, which needs a testcase.

Thanks a lot for pushing this, it's still perfectly on time for the release! And thanks a lot for considering volume coupling here.

davidscn (commented):
Not all that much to add from my side. Changes look good and give a couple of more reasons to finally introduce generic readers and writers.

The only thing I don't like is the name of the variables, but from what I understand from the code these are already set by OpenFOAM. Still, it could be helpful to add some code documentation on what these variables are and where to use them. Alpha is a volume fraction for multi-phase flows?

thesamriel:
> The only thing I don't like is the name of the variables, but from what I understand from the code these are already set by OpenFOAM. Still, it could be helpful to add some code documentation on what these variables are and where to use them. Alpha is a volume fraction for multi-phase flows?

I'm also not too convinced about the names, these are just the terms used by OpnFOAM. `Phi` could be renamed to `Flux` or `FaceFlux`. `Alpha` could be renamed to `VolumeFraction` but I'm not so sure how much I like that.

MakisH:
> I added the test case to the tutorials. See at the bottom of the updated description of the PR here.

Please also start a PR, so that I can add comments there (and merge it). It currently seems to only be a branch.

docs/config.md
MakisH: Added.

MakisH:
> > The only thing I don't like is the name of the variables, but from what I understand from the code these are already set by OpenFOAM. Still, it could be helpful to add some code documentation on what these variables are and where to use them. Alpha is a volume fraction for multi-phase flows?
> 
> I'm also not too convinced about the names, these are just the terms used by OpnFOAM. `Phi` could be renamed to `Flux` or `FaceFlux`. `Alpha` could be renamed to `VolumeFraction` but I'm not so sure how much I like that.

This is the first adapter that supports multiphase flows in flow coupling, and these names are clear in the OpenFOAM community. I think we can keep them like this for now, and deprecate them in the future, if a need pops up.

MakisH (approved):
I updated the README, added a changelog entry, and merged back from develop.
I can run the accompanying test case in < 3 min and get the expected results.

I think this is now ready to merge! :rocket:

PR #309, opened 2023-12-14
Added a test workflow

valentin-seitz:


MakisH (commented):
Nice! Seems to be very small and elegant. How can we test it? Can you please add more details in the description?

MakisH (approved):
Approving to test the event trigger.

We generally don't really like that it would only start once the whole PR is approved, but that's maybe fine since this is only the system tests.

MakisH (approved):

.github/workflows/run-tests-on-pr.yaml
```diff
@@ -0,0 +1,13 @@
...
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  testing-approved-pr:
+    if: github.event.review.state == 'APPROVED'
+    uses: precice/tutorials/.github/workflows/run_testsuite_workflow.yml@develop
+    with:
+      suites: openfoam_adapter_pr
+      build_args: TUTORIALS_REF:develop,PRECICE_REF:develop,OPENFOAM_EXECUTABLE:openfoam2306,OPENFOAM_ADAPTER_REF:${{ github.event.pull_request.head.sha }}
```
MakisH, line 11: Now that we only support commits and tags, we need a way to specify that we need the latest develop branches.

MakisH:
@valentin-seitz is this ready to merge?

.github/workflows/run-tests-on-pr.yaml
```diff
@@ -0,0 +1,13 @@
+on:
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  testing-approved-pr:
+    if: github.event.review.state == 'APPROVED'
```
MakisH, line 7: A better approach here would be that we check if the PR has a label `ready to test`. Since only maintainers can edit labels, this should be relatively secure and convenient, without messing with the actual approval of the PR.

Suggested by @jpthiele in his talk at deRSE24.

.github/workflows/run-tests-on-pr.yaml
MakisH: This currently gives:

```
build_args=TUTORIALS_REF:develop,PRECICE_REF:develop,OPENFOAM_EXECUTABLE:openfoam2306,OPENFOAM_ADAPTER_REF:7d19f75264028f575e44d980f7e91715f3d21222 --suites=openfoam_adapter_pr --log-level=DEBUG
```

The adapter is correct, but the develop branches of the other components are probably not the very latest. Maybe the simplest would be to have a workflow or cron job that updates/invalidates the cached develop images.

MakisH: I now used https://github.com/marketplace/actions/get-latest-commit to get the latest develop branch for the tutorials.

I had to split the job into two: one for gathering the references and one for triggering the system tests with these references.

The jobs are sharing outputs: https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

MakisH:
The system tests are now triggered by adding (or removing and re-adding) the `trigger-system-tests` label to the pull request.

They use:

- preCICE v3.0.0
- OpenFOAM v2306
- latest `develop` of tutorials
- `HEAD` of the pull request

They currently fail, as expected, with:

```
CRITICAL: Fieldcompare returned non zero exit code, therefore Flow over heated plate (fluid-openfoam, solid-openfoam) failed
```

This is something to fix in the tutorials repository.

MakisH (approved):

PR #312, opened 2024-01-23
Introduce tolerances for triggering dt related warnings

### LINKED ISSUE #311: Adapter autodetecs subcycling erroneously
Using the current state of the develop adater ([99421f7](https://github.com/precice/openfoam-adapter/commit/99421f77f85e91262f921cf0bbdc3724d0a9cb20)) and the preCICE v3 release candidate, we run into the following two warnings when running our perpendicular flap tutorial:

```bash
--> FOAM Warning : 
    in file Utilities.C at line 17
    Warning in the preCICE adapter: 
The solver's timestep cannot be larger than the coupling timestep. Adjusting from 0.010000 to 0.010000
```

or 

```
--> FOAM Warning : 
    in file Utilities.C at line 17
    Warning in the preCICE adapter: 
The adapter does not fully support subcycling for FSI and instabilities may occur.
```

which have essentially the same origin, namely the floating point comparison regarding the time-step size in the adapter

https://github.com/precice/openfoam-adapter/blob/99421f77f85e91262f921cf0bbdc3724d0a9cb20/Adapter.C#L655-L685

In the tutorial itself, the time-step size is supposed to be fixed, so none of these warnings should pop up. However, using (as an example) an adjustable time-step size for writing result files might lead to noteworthy changes of the time-step size in OpenFOAM. In principle, the warning could be ignored as we adjust the time-step size in each if branch (of the code above) according to the allowed range. However, auto-enabling subcycling leads to additional checkpointing, which is undesired and we should understand the origin of the issue better to trigger the warnings, when they are actually helpful. 

There were recently changes regarding `dt` in the adapter (see #298 ) and also in preCICE. Investigating this issue requires some effort.

davidscn:
See #311 for details. preCICE v3 operates now with a higher precision for handling time-step sizes (by default e-14), such that the comparisons up to machine precision are triggered unintentionally. With this changes, we have now a comparison with tolerances.

Resolves #311

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -651,7 +651,7 @@ void preciceAdapter::Adapter::adjustSolverTimeStepAndReadData()
        If the solver tries to use a bigger timestep, then it needs to use
        the same timestep as the one determined by preCICE.
     */
-    double tolerance = 1e-12;
+    double tolerance = 1e-14;
     if (timestepPrecice_ - timestepSolverDetermined > tolerance)
     {
         // Add a bool 'subCycling = true' which is checked in the storeMeshPoints() function.
--- a/Adapter.C
+++ b/Adapter.C
@@ -651,8 +651,8 @@ void preciceAdapter::Adapter::adjustSolverTimeStepAndReadData()
        If the solver tries to use a bigger timestep, then it needs to use
        the same timestep as the one determined by preCICE.
     */
-
-    if (timestepSolverDetermined < timestepPrecice_)
+    double tolerance = 1e-12;
+    if (timestepPrecice_ - timestepSolverDetermined > tolerance)
     {
         // Add a bool 'subCycling = true' which is checked in the storeMeshPoints() function.
         adapterInfo(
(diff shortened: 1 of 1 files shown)


Adapter.C
```diff
@@ -651,8 +651,8 @@ void preciceAdapter::Adapter::adjustSolverTimeStepAndReadData()
        If the solver tries to use a bigger timestep, then it needs to use
        the same timestep as the one determined by preCICE.
     */
-
-    if (timestepSolverDetermined < timestepPrecice_)
+    double tolerance = 1e-12;
```
MakisH, line 654: Why e-12 if preCICE sets e-14?
We should probably add a comment.

Ideally, preCICE should provide this.

MakisH (approved):
Makes sense. I think we only need a comment.

This only affects the warnings and has nothing to do with detecting the end of the simulation, correct?

Adapter.C
davidscn: > Ideally, preCICE should provide this.

If you think this is valuable, please comment here https://github.com/precice/precice/issues/1902

Overall, we just use it to inform the user and I don't see a specific need or reason to ask preCICE for this.

> Why e-12 if preCICE sets e-14?

Putting it the other way around: for which differences do you want to emit a warning? I just picked what I thought could be of practical relevance, but we can definitely change it.

MakisH: Does the issue originate primarily from error accumulation in preCICE, or in OpenFOAM? Is it controlled by any accuracy set (at compile-time or runtime) by preCICE?

If it depends on the accuracy set in preCICE, then it just sounds reasonable to me to use that directly here, as one can then make the connection.

But since this only affects warnings and the behavior does not change, then it does not really matter.

No need to change anything in this context.

davidscn: It is a compile time constant in preCICE, but there is no guarantee that our time-step computation is accurate up to these many digits, but ok then let me pick 14.

PR #315, opened 2024-01-23
Fix connectivity definition in the adapter

### LINKED ISSUE #313: Mesh connectivity not defined correctly
Using the all current develop states
- https://github.com/precice/precice/tree/release-v3.0.0 [9729b7c](https://github.com/precice/precice/commit/9729b7ca028a8e32cd2beb564f642fad5ef37693)
- [99421f7](https://github.com/precice/openfoam-adapter/commit/99421f77f85e91262f921cf0bbdc3724d0a9cb20)

and running our [flow-over-heated-plate nearest projection](https://github.com/precice/tutorials/tree/master/flow-over-heated-plate-nearest-projection) tutorial gives me the following export mesh

image

which is obviously wrong. Either the export in preCICE or our definition in the adapter is either in the last or first iteration in the loop wrong, I suspect.

davidscn:
Closes #313 

This was already anticipated in https://github.com/precice/openfoam-adapter/pull/297#discussion_r1273447571
`std::map` is sorted according to the key, which leads to wrong key-value pairs in the previous implementation, as the keys are resorted before we insert the (not resorted) values.

### DIFF
--- a/Interface.C
+++ b/Interface.C
@@ -249,10 +249,6 @@ void preciceAdapter::Interface::configureMesh(const fvMesh& mesh, const std::str
                 {
                     vertices[verticesIndex++] = faceNodes[i][d];
                 }
-                if (meshConnectivity_)
-                {
-                    verticesMap.emplace(std::make_tuple(faceNodes[i][0], faceNodes[i][1], faceNodes[i][2]), -1);
-                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


MakisH:
Wouldn't it be simpler to replace the `std::map` with an `[std::unordered_map](https://en.cppreference.com/w/cpp/container/unordered_map)`?

davidscn:
Considering, which if these versions is more compact, I like the current version more. An unordered container would solve the problem, too, of course.

MakisH (approved):
The changes make sense, I have not tested anything. I also like this implementation better, as it moves the creation of the map after the `setMeshVertices` call.

We could also update the example on https://precice.org/website-testing/couple-your-code-defining-mesh-connectivity.html to use an `unordered_map`, as I guess more people will overlook this detail in their implementations.

PR #319, opened 2024-02-08
Release v2.0.0

davidscn:
TODO list:

- [ ] check for missing changelog entries

I don't know what to do with the foundation versions yet. With the changes and new features, it's noteworthy that the `faceTriangulation`/connectivity definition is different, the volume coupling changes seem incompatible, and also the handling of time/delta t is different (#298), which leads to many merge conflicts. Without some testing automation, following our typical rebase strategy runs into a lot of merge conflicts, which would require manual testing.

So, open question: which versions do we want to support in the upcoming release?


uekerman:
Why going for a breaking release? A breaking dependency is normally no sufficient reason. Maybe I am overlooking sth.

davidscn:
> Why going for a breaking release? A breaking dependency is normally no sufficient reason. Maybe I am overlooking sth.

Docs are here https://precice.org/adapter-openfoam-get.html#what-does-the-adapter-version-mean. Which version do you suggest?

MakisH:
We have defined the configuration file format of the adapter (`preciceDict`) as the "stable API" of the adapter. Since we don't change that, this is still a v1.x. We discussed the "what would happen if we release a new preCICE major version" when creating the versioning system and we decided that it would not affect the adapter version. Definitely not a perfect system, but the explicit "`preciceDict` is what we need to keep stable" helps as a guideline.

New features are, for example, the volume coupling (https://github.com/precice/openfoam-adapter/pull/255) and the extensions to the fluid-fluid module (https://github.com/precice/openfoam-adapter/pull/281, https://github.com/precice/openfoam-adapter/pull/308).

For these reasons, I propose v1.3.0.
We should still clearly mention the preCICE v3 requirements in the release notes.

If we decide to indeed not have v2, then I would actually suggest to also rename the branch, to avoid confusion in the future.

> I don't know what to do with the foundation versions yet.

> So, open question: which versions do we want to support in the upcoming release?

I would not attempt to make our releases dependent on supporting new Foundation versions. We release our adapter when we are ready, and we can always release more ports to other OpenFOAM versions.

There are still a few things I would like to check before releasing, I will approve after that, but not before next week.

PR #320, opened 2024-02-09
Release v1.3.0

davidscn:
TODO list:

- [x] check for missing changelog entries

I don't know what to do with the foundation versions yet. With the changes and new features, it's noteworthy that the `faceTriangulation`/connectivity definition is different, the volume coupling changes seem incompatible, and also the handling of time/delta t is different (#298), which leads to many merge conflicts. Without some testing automation, following our typical rebase strategy runs into a lot of merge conflicts, which would require manual testing.

So, open question: which versions do we want to support in the upcoming release?


MakisH (approved):

MakisH:
I updated the documentation and the changelog.
I have been testing (almost) this state for a while now.

PR #322, opened 2024-03-10
Remove unnecessary `timestepPrecice_`

BenjaminRodenberg:
This is a pure refactoring to simplify the code if determining the time step size.

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -548,7 +548,6 @@ void preciceAdapter::Adapter::initialize()
 
     DEBUG(adapterInfo("Initializing preCICE data..."));
     precice_->initialize();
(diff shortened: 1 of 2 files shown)


MakisH (approved):
This indeed looks like a relic.
Is it only unnecessary, or was it also a bug?

BenjaminRodenberg:
> This indeed looks like a relic. Is it only unnecessary, or was it also a bug?

I think it was only unnecessarily complicated.

But it's always hard to pin bugs down if subcycling is involved, so I cannot promise I have not accidentally fixed a bug :smirk: Therefore I also changed this to have a version of the code where I'm quite optimistic it's not buggy.

PR #325, opened 2024-03-19
Various documentation and CI updates

MakisH:
- Bump default OpenFOAM version from v2306 to v2312.
- Update badges, add a link to the forum in the badges, to direct people there for questions, instead of the issues.

TODO list:

- I added a changelog entry in `changelog-entries/` (create directory if missing) -> N/A


docs/get.md
```diff
@@ -8,7 +8,8 @@ summary: "Get the code from GitHub and run ./Allwmake. If this fails, look into
 To build the adapter, you need to install a few dependencies and then execute the `Allwmake` script.
 
 1. Install [a compatible OpenFOAM distribution](https://precice.org/adapter-openfoam-support.html).
-2. Install [preCICE v3](https://precice.org/installation-overview.html). In case you need preCICE v2, please install an older version of the adapter (v1.2.3 is the latest release to support preCICE v2).
+2. Install [preCICE v3](https://precice.org/installation-overview.html).
+    * In case you need preCICE v2, please install an older version of the adapter (v1.2.3 was the last release to support preCICE v2).
```
davidscn, line 12: It would be good to add a link to the release archive, as finding it might not be trivial in case you don't know where to look for older releases.

MakisH:
I also switched from using the preCICE Debian packages to the setup-precice-action in the build-custom: https://github.com/precice/openfoam-adapter/actions/runs/8364095082

MakisH:
Something seems to be completely broken in regarding the step outputs in GitHub Actions. Even though we don't seem to have changed anything related, the value of `openfoam_exec` we set in the `InstallOpenFOAM` step does not propagate to the adapter installation step.

Documentation: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

I suspect that there is either a bug with GitHub Actions, or something changed and is not documented.

Example job that fails: https://github.com/precice/openfoam-adapter/actions/runs/8365204318/job/22902472048

Note that I could install OpenFOAM v2206 on Ubuntu 22.04, but several other combinations seemed to lead to various sourceforge-related issues (404 for various subpackages).

MakisH:
The workflow is broken even on develop. This PR has nothing to do with it. I give up and merge.

PR #330, opened 2024-06-19
Skip velocity object from the initializer list

MakisH:
@Fujikawas noticed that the [volume-coupled flow tutorial](https://precice.org/tutorials-volume-coupled-flow.html) was failing early on, complaining that it could not find `U_vol` when adding coupling data readers.

In the original implementation (https://github.com/precice/openfoam-adapter/pull/270), the velocity was not initialized in the initializer list of the `Velocity` constructor, but it was deferred to the body, where it was created if needed. This is actually needed when reading source terms, since we need a temporary object to read data from preCICE.

In the follow-up extension https://github.com/precice/openfoam-adapter/pull/281, the OpenFOAM velocity object was added to the initializer list, at which point it tries to find an object that (intentionally) does not exist at that point. @thesamriel do I overlook any other reason this was added again into the initializer list?

TODO list:

### DIFF
--- a/FF/Velocity.C
+++ b/FF/Velocity.C
@@ -8,20 +8,19 @@ preciceAdapter::FF::Velocity::Velocity(
     const std::string nameU,
     const std::string namePhi,
     bool fluxCorrection)
-: U_(
-    const_cast<volVectorField*>(
-        &mesh.lookupObject<volVectorField>(nameU))),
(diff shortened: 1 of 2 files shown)


MakisH:
I assumed so, makes sense. Thank you for the quick reply!

Fujikawas (approved):
The changes make sense for me and solved the error in the tutorial case.

PR #334, opened 2024-06-26
Document the type of gradients we support

### LINKED ISSUE #543: Gradients: Surface-normal (scalars) vs complete (vector)
In our tutorials and in the OpenFOAM adapter, we currently treat pressure gradient and velocity gradient as surface-normal gradients, and thus scalar fields. This already created some confusion in some discussion with @Fujikawas.

Isn't this a bit weird to assume that gradients are always surface-normal?

MakisH:
Closes https://github.com/precice/tutorials/issues/543

TODO list:

- I added a changelog entry in `changelog-entries/` (create directory if missing) -> not needed


MakisH (commented):
Discussed with @uekerman, this is already ok.

Issue #328, opened 2024-04-26
Policy on support for OpenFOAM.org and other versions


MakisH:
As the ESI / OpenCFD (.com) and CFD Direct / OpenFOAM Foundation (.org) versions are diverging more and more, it is becoming impossible to maintain the same level of support we started with. The amount of effort required is extremely high, especially since every new version of OpenFOAM.org brings several breaking changes.

There seem to be some users relying mainly on the latest versions of OpenFOAM.org, mainly judging from the [download numbers of the release artifacts](https://somsubhra.github.io/github-release-stats/?username=precice&repository=openfoam-adapter&page=1&per_page=5) and PRs such as #310 (and similarly for foam-extend #302).

At the same time, we currently maintain commented-out configuration variants in some of our tutorials, and since OpenFOAM 10, a [complete branch](https://github.com/precice/tutorials/tree/OpenFOAM10).

Motivation for supporting multiple versions is mainly because people are often restricted from custom codes and what version is installed on their local system. We couple codes, so we want to enable diversity. However, supporting old versions as well is becoming extremely complex and keeps the whole project behind.

As a way forward, and to (a) simplify our release process and (b) facilitate community contributions for other versions, I suggest the following:

- Our latest release only supports the latest OpenFOAM.com version, as we mostly already do. This means:
   - No more release artifacts for older versions
   - No more "if this version do this, else do that" (see #32)
   - If someone needs an older version, they can get an older version of the adapter.
- We host one branch for OpenFOAM.org, aiming to support the latest release of OpenFOAM.org.
  - The community can contribute to this branch. 
  - Only compatibility updates can be contributed to these branches. All features and general fixes must be contributed to the develop branch.
  - We generate release artifacts for the latest supported version in every release.
  - If this is not ready, we proceed with our release, and attach additional artifacts later.
  - We only attach an artifact if all tests (see below) pass, and all preCICE features of the main adapter are supported or documented in the code and the user documentation.
- We similarly host one branch for foam-extend, etc. 

For tutorials:

- We consider each OpenFOAM variant to be a separate code. This means that it will get its own case directory in some tutorials: `fluid-openfoam-org`.
- We only provide such variants for a couple of tutorials (`partitioned-heat-conduction`, `flow-over-heated-plate`, and `perpendicular-flap`, potentially also `partitioned-pipe`). If someone wants to port more cases, they can compare the changes.
- We add these cases to the system tests.

Any suggestions?

hoehnp:
[...]

PR #335, opened 2024-06-29
OpenFOAM 10: Fixes adapter not reading coupling data

simrnjandu:
Added missing call to readCouplingData(), which ensures that data is read from the precice buffer.

This fixes the issue of perpendicular-flap tutorial showing no displacement of the interface on openfoam's side.

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -687,7 +687,7 @@ void preciceAdapter::Adapter::adjustSolverTimeStepAndReadData()
     // which also triggers the functionObject's adjustTimeStep())
     // TODO: Keep this in mind if any relevant problem appears.
     const_cast<Time&>(runTime_).setDeltaTNoAdjust(timestepSolver_);
-
+    readCouplingData(runTime_.deltaT().value());
     return;
 }
 


MakisH:
Thank you for contributing!
Important context (it took me a bit to realize): this is targeting the `OpenFOAM10` branch and is not an issue on `develop`/`master`:

https://github.com/precice/openfoam-adapter/blob/develop/Adapter.C#L698

This is the result of an incomplete rebasing of the related .org branches, which I still need to fix at some point. But this is interesting now: if you apply this fix, does everything else work? What have you tried?

For context, see #328. Please comment there regarding your needs.

simrnjandu:
> Important context (it took me a bit to realize): this is targeting the `OpenFOAM10` branch and is not an issue on `develop`/`master`

I forgot to add a link to the forum topic I created regarding this: [https://precice.discourse.group/t/openfoam-org-adapter-doesnt-read-celldisplacement/1971](https://precice.discourse.group/t/openfoam-org-adapter-doesnt-read-celldisplacement/1971)



> But this is interesting now: if you apply this fix, does everything else work? What have you tried?

After this fix, these tutorials work as expected: `perpendicular-flap` (fluid-openfoam and solid-dealii) and `flow-over-heated-plate` (fluid-openfoam and solid-openfoam)
(after necessary modification of the openfoam case files: [https://github.com/precice/tutorials/commit/d5911c9c4745a1df027c30b99e25d39b3aeb339e](https://github.com/precice/tutorials/commit/d5911c9c4745a1df027c30b99e25d39b3aeb339e) ).

I have not tried the other tutorials, but I believe they would work as well.

> For context, see #328. Please comment there regarding your needs.

I am not using any custom solver, so the latest openfoam org version that is supported by precice does the job for me. I need the org version because the ESI one does not have the required rheology models: [https://doc.cfd.direct/openfoam/user-guide-v10/transport-rheology](https://doc.cfd.direct/openfoam/user-guide-v10/transport-rheology)
For now, OpenFOAM10 is working perfectly fine for me!

PR #338, opened 2024-08-19
Use setup-openfoam action in workflows

gerlero:
Switches to the [setup-openfoam](https://github.com/gerlero/setup-openfoam) action (disclaimer: I made the action) for installing OpenFOAM in the build workflows.

I hope you can agree that this simplifies the workflows quite a bit.

TODO list:

### DIFF
--- a/.github/workflows/build-custom.yml
+++ b/.github/workflows/build-custom.yml
@@ -20,22 +20,22 @@ on:
         required: true
         type: choice
         options:
-          - OpenFOAMv2406
-          - OpenFOAMv2312
-          - OpenFOAMv2306
-          - OpenFOAMv2212
-          - OpenFOAMv2206
-          - OpenFOAMv2112
-          - OpenFOAMv2106
-          - OpenFOAMv2012
-          - OpenFOAMv2006
-          - OpenFOAMv1912
-          - OpenFOAM10
-          - OpenFOAM9
-          - OpenFOAM8
-          - OpenFOAM7
-          - OpenFOAM6
-          - OpenFOAM5
+          - 2406
+          - 2312
+          - 2306
+          - 2212
+          - 2206
+          - 2112
+          - 2106
+          - 2012
+          - 2006
+          - 1912
+          - 10
+          - 9
+          - 8
+          - 7
+          - 6
+          - 5
       versionpreCICE:
         description: 'Version of preCICE to build with'
         default: 'v3.0.0'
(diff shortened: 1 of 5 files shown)


gerlero:
I've updated the changelog directly, but now I'm not sure that's the correct workflow. Also, I assume docs changes are not needed for this.

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ Read more details in the issue [#52: Releases and versioning](https://github.com
 
 <!-- markdownlint-configure-file {"MD024": { "siblings_only": true } } -->
 
+## [v1.3.2] Unreleased
+
+### Changed
+
+- Use the [setup-openfoam](https://github.com/marketplace/actions/setup-openfoam) action to install OpenFOAM within CI workflows [#338](https://github.com/precice/openfoam-adapter/pull/338).
```
davidscn: This line would go into a separate file according to the number of the PR `338.md` into a directory called `changelog-entries` (to be created). The file edited here is only edited throughout the release. See also https://github.com/precice/openfoam-adapter/blob/develop/CONTRIBUTING.md#changelog

davidscn (commented):
This looks like a simplification, let's see what the CI says.

.github/workflows/build.yml
```diff
@@ -25,23 +25,9 @@ jobs:
...
-    - name: Setup caches
-      uses: actions/cache@v4
-      id: cache-id
+    - name: Set up OpenFOAM
+      uses: gerlero/setup-openfoam@v1
       with:
-        path: ${{ runner.temp }}/cache-directory
-        key: ${{ runner.os }}-cache-dependencies-v1.0
-    - name: Install OpenFOAM (considering cache)
-      uses: airvzxf/cache-anything-new-action@v1.0.1
```
MakisH, line 39: Does the `setup-openfoam` action apply any caching?
The main idea here was to not spam the OpenFOAM deb repositories, but it is not that this repository is extremely active at the moment.

.github/workflows/build-custom.yml
```diff
@@ -187,28 +98,28 @@ jobs:
         if ${{ github.event.inputs.runTutorialHeatedPlate }}
         then
           cd tutorials/flow-over-heated-plate/fluid-openfoam
-          ${{steps.installOpenFOAM.outputs.openfoam_exec}} ./run.sh | tee fluid-openfoam.log 2>&1 &
+          ./run.sh | tee fluid-openfoam.log 2>&1 &
```
MakisH, line 101: Have you tested if this workflow indeed works? I remember it was tricky to expose the environment (but I don't remember why anymore).

MakisH (commented):
This is a very useful addition, thank you very much @gerlero!

On a side note, this would make for a nice presentation at the OpenFOAM RSE group: https://wiki.openfoam.com/Research_Software_Engineering_Special_Interest_Group

Docs changes are indeed not needed. For the changelog, see what David wrote (or I can apply the change directly).

CHANGELOG.md
gerlero: @davidscn Thanks. I'll take care of it.

gerlero:
> On a side note, this would make for a nice presentation at the OpenFOAM RSE group: https://wiki.openfoam.com/Research_Software_Engineering_Special_Interest_Group

Sure, no problem. I've been aware of those meetings for a couple of months but couldn't make it to the last one. The reason I'm making this stuff is that I'm running OpenFOAM-based CIs myself for a few projects, so our interests should overlap quite a bit. I'll make sure to attend the next one.

.github/workflows/build.yml
gerlero: It does attempt to cache the deb via the [apt-deb-cache](https://github.com/marketplace/actions/cache-and-install-apt-deb-packages) action, but currently this doesn't work when running inside a container (https://github.com/gerlero/setup-openfoam/pull/13). My goal is to fix this in the near future though.

.github/workflows/build-custom.yml
gerlero: Keeping the environment for future steps is not straightforward, but the action manages to do it by tracking the OpenFOAM variables and exposing them via `$GITHUB_ENV` and `$GITHUB_PATH`.

I've tested the workflow with my fork now and it seems to work okay: https://github.com/gerlero/openfoam-adapter/actions/runs/10489014497/job/29052697825

CHANGELOG.md
gerlero: Done.

.github/workflows/build.yml
MakisH: Alright, happy to not have to maintain the caching mechanism here anyway! :smile: 
If you add more details in the issue, I am happy to help debug it.

This is not a blocking issue, in my opinion. I think the benefit is much higher.

.github/workflows/build-custom.yml
MakisH: Good point about the `$GITHUB_ENV` and `$GITHUB_PATH`, thanks!
Apparently, it works. We can discover any corner cases in develop.

MakisH (approved):

.github/workflows/build-custom.yml
MakisH: Also works in the main repo:

- OpenFOAM v2106 on Ubuntu 20.04, using the develop branch: https://github.com/precice/openfoam-adapter/actions/runs/10507653425
- OpenFOAM 7 on Ubuntu 20.04, using the OpenFOAM7 branch: https://github.com/precice/openfoam-adapter/actions/runs/10507736208 (the adapter building fails as currently known)

.github/workflows/build.yml
gerlero: Thanks. I'm waiting for https://github.com/eth-pkg/apt-deb-cache/pull/3 before trying to debug it again (that PR should fix some issues). I'm also playing with rolling my own caching action if that doesn't work. OpenFOAM is kinda finicky so most other "apt cache" actions I've tried do not work.

gerlero: Not to miss it, the cache-anything-new-action should work but the requirement that you pass a script (instead of simple commands) makes it more difficult to configure. Not that it wouldn't be possible to do so, but that's what pushed me towards trying to roll my own.

MakisH: For me, the cache-anything-new-action was just an easy way to say "I don't know what exactly I need to cache, so this automated method should work".

gerlero: @MakisH caching is now working even in containers (I ended up rolling my own @ [gerlero/apt-install](https://github.com/gerlero/apt-install)).

Issue #158, opened 2021-04-01
Different dimensions for dimensions Issue in FSI when turbulence model is used - OpenFOAM adapter


dntjudme:
Hi there,

This issue was found when I was doing a FSI simulation using OpenFOAM v7 / CalculiX v2.15 / preCICE v2.0.0.
Please let me lead you to this [discourse-post](https://precice.discourse.group/t/different-dimensions-for-dimensions-issue/535) for more details.

As suggested by @MakisH , a brief introduction to the settings of turbulence model in OpenFOAM is to simply turn on the turbulence by setting  `/constant/turbulenceProperties` file, using:
```
simulationType RAS;
RAS
{
    RASModel        kOmegaSST;
    turbulence      on;
}
```
Of course, the corresponding initial conditions such as `k`, `omega`, `nut`, `p` and `U` are well defined in the `0/` directory.

When the turbulence model is turned on, the FSI simulation stops at the 2nd iteration of the first time step of the coupling (shown in the figure in the [post](https://precice.discourse.group/t/different-dimensions-for-dimensions-issue/535/2))

Interestingly however, when I turned off the turbulence (switch to `laminar`):

`simulationType laminar;`

The FSI simulation went well and no errors are reported. However, the results are definitely not I'm expecting since the turbulent behaviours cannot be well resolved in my FSI simulation.

As @MakisH said, I do also suspect this issue might be originated from the OpenFOAM adapter side, so I'm looking for help from developers.

Should you requires more information please contact me directly and I'm happy to assists.

DavidSCN:
I cannot reproduce this issue with OpenFOAM v2012 and since I have at the moment no foundation version installed I cannot try with another OpenFOAM version. Are you tied to use OpenFOAM v7?

dntjudme:
Yes, I use OpenFOAM v7.
New updates: I also implemented other 3 types of turbulence models : SA one equation / k-epsilon / k-omega.
None of them throw back any dimension mismatch errors.

ernestoricciardi:
I'm having the same issue with OpenFOAM v7 and kOmegaSST for turbolence. 

I also tried with kOmega and I get MPI_ERR_TRUNCATE.

There are any news for fixing the dimensions error for kOmegaSST?

martinsaravia:
I am having the same issue with OF10 and Calculix using kOmegaSST turbulence model. Is there any news regarding this?

MakisH:
[...]

PR #344, opened 2024-11-20
Fix different dimensions error  / Checkpointing issue #158

vidulejs:
In issue #158 it was discovered that using the kOmegaSST turbulence model together with implicit coupling causes a dimension mismatch error in OpenFOAM, when starting the second timestep / reading the checkpointed fields. The perpetrator was found to be the field 'grad(U)', which was created in the first timestep during the execution of the kOmegaSST turbulence model.  The field is created only temporarily and afterwards is not stored in the OpenFOAM registry / fields database, thus causing an error when trying to read it in the next timesteps. A failing case was posted by a user:

>Hi makis, I have made a case related to this problem, you may run this case to catch the error.
>[3000_40Mpa_orig.zip](https://github.com/user-attachments/files/17321380/3000_40Mpa_orig.zip)

_Originally posted by @c7888026 in https://github.com/precice/openfoam-adapter/issues/158#issuecomment-2403864736_

However the problem is more general since other functions can create temporary checkpointable fields, such as 'yPsi' used for 'wallDist' calculation as posted by user:

>... The log is attached above , I found that the variable **yPSi** that is calculated in the _fvSolution_ and required by the method of the wall distance in the _fvSchemes_  is the problem, so it behaves as the gradU as you described and not passed correctly through the check point . I changed the wallDist method to meshWave as @c7888026 did in his case , and it worked fine even with K-omega sst and in parallel setting using mpirun , so it is matter of variables that are not passing well ( may be ) through the checkpoint step.

_Originally posted by @KariimAhmed in https://github.com/precice/openfoam-adapter/issues/158#issuecomment-2468859857_ 

I could not yet replicate this issue, because for some reason OpenFOAM is still selecting 'meshWave' method for 'wallDist' instead of 'Poisson', which I specified in fvSchemes. Also there are no preCICE tutorials without 'meshWave'.

The general solution I found is to check which fields are stored in the OpenFOAM database and which ones are checkpointed, then remove those checkpointed fields which are no longer used by OpenFOAM. This new function 'pruneCheckpointedFields()' is executed before the `readCheckpoint()` function, where the copy was loaded `*field = *fieldCopy;`. I am not sure, but I assume the field pointer was set to another field by OpenFOAM, otherwise it would be a memory error and not a dimensions error.  

```C++
void preciceAdapter::Adapter::pruneCheckpointedFields()
{
    // Check if checkpointed fields exist in OpenFOAM database
    // If not, remove them from the checkpointed fields list
	// [...]
}
```
```C++
void preciceAdapter::Adapter::readCheckpoint()
{
	// [...]
	
    // Reload all the fields of type volVectorField
    for (const auto& kv : volVectorFields_)
    {
        const std::string& key = kv.first;
        volVectorField* field = kv.second;
        volVectorField* fieldCopy = volVectorFieldCopies_.at(key);

        // Load the volume field
        *field = *fieldCopy;

        int nOldTimes = field->nOldTimes();
        if (nOldTimes >= 1)
        {
            field->oldTime() = fieldCopy->oldTime();
        }
        if (nOldTimes == 2)
        {
            field->oldTime().oldTime() = fieldCopy->oldTime().oldTime();
        }
    }
	// [...]
}
```

Originally I thought it was an issue with storing the field pointers in a vector and using the wrong index to access this vector, so I tried to change the storage to an unordered_map and use the field name as the key:

```C++
    //- Checkpointed volScalarField fields
    std::unordered_map<std::string, Foam::volScalarField*> volScalarFields_;
```

However, the problem remained unchanged. Thus I can revert the storage back to vectors for simplicity and access the name of the field from the field object when accessing it from the vector.  

Also, the new function 'pruneCheckpointedFields()'  uses a macro and is not the prettiest, so I am open for a refactor. I included the '<algorithm>' header in 'Adapter.H' to search the fields database, but it can be done without it. When a field is removed from checkpointing, a new debug message is logged:

```Bash
[2] ---[preciceAdapter] [DEBUG] Removed volTensorField : grad(U) from the checkpointed fields list.
```

TODO list:

- [x] Revert the fields storage from vectors to unordered_map  ?
- Can be done without using macro (it's fine as it is - @MakisH)
- [x] Can be done without including <algorithm> header
- [x] Replicate and test issue with a different wallDist method than meshWave (e.g. Poisson)

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -437,6 +437,7 @@ void preciceAdapter::Adapter::execute()
     // Read checkpoint if required
     if (requiresReadingCheckpoint())
     {
+        pruneCheckpointedFields();
         readCheckpoint();
     }
 
@@ -924,17 +925,26 @@ void preciceAdapter::Adapter::setupCheckpointing()
     ACCUMULATE_TIMER(timeInCheckpointingSetup_);
 }
 
-void preciceAdapter::Adapter::printFieldCounts()
+void preciceAdapter::Adapter::printFieldCountsDB()
 {
+
+// Print fields in the OpenFOAM database
+// don't print the fields ending with '0' aka oldTime
+
 std::string fields;
 int count = 0;
+int strlen = 0;
 
 #undef doLocalCode
 #define doLocalCode(GeomField)                                           \
     for (const word& obj : mesh_.sortedNames<GeomField>())               \
     {                                                                    \
-        fields += obj + " ";                                             \
-        count++;                                                         \
+        strlen = obj.size();                                             \
+        if (obj[strlen - 1] != '0')                                      \
+        {                                                                \
+            fields += obj + " ";                                         \
+            count++;                                                     \
+        }                                                                \
     }                                                                    \
     DEBUG(adapterInfo(std::to_string(count) + " : " #GeomField + " : " + fields));
     fields = "";
@@ -953,11 +963,54 @@ int count = 0;
     doLocalCode(pointVectorField);
     doLocalCode(pointTensorField);
 
-    // NOTE: Add here other object types to checkpoint, if needed.
-
 #undef doLocalCode
 }
 
+void preciceAdapter::Adapter::pruneCheckpointedFields()
+{
+    // Check if checkpointed fields exist in OpenFOAM database
+    // If not, remove them from the checkpointed fields list
+
+    word obj;
+    std::vector<word> fieldNames;
+    std::vector<std::string> toRemove;
+#undef doLocalCode
+#define doLocalCode(GeomField, GeomFieldMap, GeomFieldCopiesMap)           \
+    fieldNames.clear();                                                    \
+    toRemove.clear();                                                      \
+    for (const word& obj : mesh_.sortedNames<GeomField>())                 \
+    {                                                                      \
+        fieldNames.push_back(obj);                                         \
+    }                                                                      \
+    for (const auto& kv : GeomFieldMap){                                   \
+        obj = kv.first;                                                    \
+        if (std::find(fieldNames.begin(), fieldNames.end(), obj) == fieldNames.end())  \
+        {                                                                  \
+            toRemove.push_back(static_cast<std::string>(obj));             \
+        }                                                                  \
+    }                                                                      \
+    for (const auto& obj : toRemove) {                                     \
+        GeomFieldMap.erase(obj);                                           \
+        delete GeomFieldCopiesMap[obj];                                    \
+        GeomFieldCopiesMap.erase(obj);                                     \
+        DEBUG(adapterInfo("Removed " #GeomField " : " + obj + " from the checkpointed fields list.")); \
+    }
+
+doLocalCode(volScalarField, volScalarFields_, volScalarFieldCopies_);
+doLocalCode(volVectorField, volVectorFields_, volVectorFieldCopies_);
+doLocalCode(volTensorField, volTensorFields_, volTensorFieldCopies_);
+doLocalCode(volSymmTensorField, volSymmTensorFields_, volSymmTensorFieldCopies_);
+
+doLocalCode(surfaceScalarField, surfaceScalarFields_, surfaceScalarFieldCopies_);
+doLocalCode(surfaceVectorField, surfaceVectorFields_, surfaceVectorFieldCopies_);
+doLocalCode(surfaceTensorField, surfaceTensorFields_, surfaceTensorFieldCopies_);
+
+doLocalCode(pointScalarField, pointScalarFields_, pointScalarFieldCopies_);
+doLocalCode(pointVectorField, pointVectorFields_, pointVectorFieldCopies_);
+doLocalCode(pointTensorField, pointTensorFields_, pointTensorFieldCopies_);
+
+#undef doLocalCode
+}
 
 // All mesh checkpointed fields
 
@@ -1102,7 +1155,7 @@ void preciceAdapter::Adapter::readCheckpoint()
     //  for efficiency.
     DEBUG(adapterInfo("Reading a checkpoint..."));
 
-    printFieldCounts();
+    printFieldCountsDB();
 
     // Reload the runTime
     reloadCheckpointTime();
@@ -1328,7 +1381,7 @@ void preciceAdapter::Adapter::writeCheckpoint()
 
     DEBUG(adapterInfo("Writing a checkpoint..."));
 
-    printFieldCounts();
+    printFieldCountsDB();
 
     // Store the runTime
     storeCheckpointTime();
--- a/Adapter.C
+++ b/Adapter.C
@@ -3,6 +3,7 @@
 #include "Utilities.H"
 
 #include "IOstreams.H"
+#include <algorithm>
 
 using namespace Foam;
 
@@ -896,12 +897,12 @@ void preciceAdapter::Adapter::setupCheckpointing()
     DEBUG(adapterInfo("Adding in checkpointed fields..."));
 
 #undef doLocalCode
-#define doLocalCode(GeomField)                                           \
-    /* Checkpoint registered GeomField objects */                        \
-    for (const word& obj : mesh_.sortedNames<GeomField>())               \
-    {                                                                    \
-        addCheckpointField(mesh_.thisDb().getObjectPtr<GeomField>(obj)); \
-        DEBUG(adapterInfo("Checkpoint " + obj + " : " #GeomField));      \
+#define doLocalCode(GeomFieldType)                                           \
+    /* Checkpoint registered GeomFieldType objects */                        \
+    for (const word& obj : mesh_.sortedNames<GeomFieldType>())               \
+    {                                                                        \
+        addCheckpointField(mesh_.thisDb().getObjectPtr<GeomFieldType>(obj)); \
+        DEBUG(adapterInfo("Checkpoint " + obj + " : " #GeomFieldType));      \
     }
 
     doLocalCode(volScalarField);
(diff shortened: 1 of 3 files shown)


KariimAhmed:
Thank you Daniels for this work. Do you suggest trying your modified Adapter.C and .H  for my Poisson wall distance issue now ? or there still other modifications to do ? 
since i will run it on a cluster so i have to be sure before changing the original Precice installed there.

vidulejs:
Hello, Karim! 
As I wrote before I was not able to test the bugfix yet with the 'yPsi' Poisson wall distance issue. I tried to modify the fvSchemes wallDist method to Poisson, but for some reason OpenFOAM still selects meshWave. I did not yet create a seperate test case for this specifically. I might try to do this tonight.

If you wish, you may try recompiling with this code and try to run your case. But there may be changes in the final pull request.

Adapter.H
```diff
@@ -177,68 +180,67 @@ private:
     //- Checkpointed volScalarField mesh fields (copies)
     std::vector<Foam::volScalarField::Internal*> volScalarInternalFieldCopies_;
 
-    // Vectors of pointers to the checkpointed fields and their copies
+    // Maps of pointers to checkpointed fields and their copies
```
MakisH: As discussed, this might not be necessary.
Let's try a quick implementation using `std::vector` and, if this gets uglier, then a map should be ok as well.

Adapter.C
```diff
@@ -923,6 +925,53 @@ void preciceAdapter::Adapter::setupCheckpointing()
...
+void preciceAdapter::Adapter::pruneCheckpointedFields()
+{
+    // Check if checkpointed fields exist in OpenFOAM database
+    // If not, remove them from the checkpointed fields list
+
+    word obj;
+    std::vector<word> fieldNames;
+    std::vector<std::string> toRemove;
+#undef doLocalCode
+#define doLocalCode(GeomField, GeomFieldMap, GeomFieldCopiesMap)                                       \
```
MakisH: I think it is fine having this as a macro for now. We can consider refactoring later, in a broader context.

MakisH (commented):
Already good that it solves the original issue. If you cannot replicate the second one, I would not make that a road-blocker. We will figure out eventually if that is a separate issue.

Adapter.C
```diff
@@ -923,6 +924,69 @@ void preciceAdapter::Adapter::setupCheckpointing()
...
+        }                                                                                                             \
+        if (!found)                                                                                                   \
+        {                                                                                                             \
+            toRemove.push_back(obj);                                                                                  \
+        }                                                                                                             \
+    }                                                                                                                 \
+    if (!toRemove.empty())                                                                                            \
+    {                                                                                                                 \
+        for (int i = toRemove.size() - 1; i >= 0; i--)                                                                \
```
MakisH: You explained to me why you chose to iterate in reverse order here. I think that would be a useful comment to add.

MakisH (approved):
The code looks clean, and the provided test case seems to work, at least for 10 time windows. In that case, the `grad(U)` is removed.

I ran the flow-over-heated-plate and the perpendicular-flap tutorials and they also seem to work.

@davidscn do you want to check anything else?

Adapter.C
davidscn: C++ has also reverse iterators

```suggestion
for (auto it = toRemove.rbegin(); it != toRemove.rend(); ++it)
```

Adapter.C
```diff
@@ -923,6 +924,69 @@ void preciceAdapter::Adapter::setupCheckpointing()
...
+                found = true;                                                                                         \
+                break;                                                                                                \
+            }                                                                                                         \
+        }                                                                                                             \
+        if (!found)                                                                                                   \
+        {                                                                                                             \
+            toRemove.push_back(obj);                                                                                  \
+        }                                                                                                             \
+    }                                                                                                                 \
```
davidscn:
```suggestion
    /* Iterate through checkpointed fields */                                                                         
  for (const auto& geomFieldCopy : GeomFieldCopies_) {                                  
     auto obj = geomFieldCopy->name();
     if (std::find(regFields.begin(), regFields.end(), obj) == regFields.end()) {
         toRemove.push_back(obj);
     }
 }                                                                                              
```
Maybe an STL version of the code would be more readable?

vidulejs: I agree this is more readable and I even had this approach first, but then I would have to include the 'algorithms' header. Is that alright? It's not used elsewhere.

Adapter.C
vidulejs: I actually need the index not the element. Then I would use std::distance to get the index.

```C++
    if (!toRemove.empty())                                                                                            \
    {                                                                                                                 \
        for (auto it = toRemove.rbegin(); it != toRemove.rend(); ++it)                                                \
        {                                                                                                             \
            auto index = std::distance(toRemove.rbegin(), it);                                                        \
            GeomField_.erase(GeomField_.begin() + index);                                                             \
            delete GeomFieldCopies_.at(index);                                                                        \
            GeomFieldCopies_.erase(GeomFieldCopies_.begin() + index);                                                 \
            DEBUG(adapterInfo("Removed " #GeomField " : " + *it + " from the checkpointed fields list."));            \
        }                                                                                                             \
    }
```

Adapter.C
MakisH: Including the algorithms header is fine, and we should be using the STL wherever it makes sense. Here, it indeed does, as there was already some uneasiness on the index-based loops.

Generally, the fact that there is not so much STL in the code is not by design. Feel free to suggest better implementations wherever you see the potential. The only thing we need to be aware is keeping the C++ standard compatible with the preCICE and OpenFOAM requirements. C++14 should be safe and this is defined by the WMake rules in the respective OpenFOAM installation (`/usr/lib/openfoam/openfoam2406/wmake/rules/General/Gcc/c++`). They [announced](https://www.openfoam.com/news/main-news/openfoam-v2312) moving to C++17 at some point in 2024, could be in the upcoming v2412.

vidulejs: Thanks for the reply! I will then quickly refactor and test the code.

Adapter.C
```diff
@@ -923,6 +924,60 @@ void preciceAdapter::Adapter::setupCheckpointing()
...
+    /* Iterate through fields in OpenFOAM registry */                                                                  \
+    for (const word& obj : mesh_.sortedNames<GeomField>())                                                             \
+    {                                                                                                                  \
+        regFields.push_back(obj);                                                                                      \
+    }                                                                                                                  \
+    /* Iterate through checkpointed fields */                                                                          \
+    for (uint i = 0; i < GeomFieldCopies_.size(); i++)                                                                 \
+    {                                                                                                                  \
+        obj = GeomFieldCopies_.at(i)->name();                                                                          \
```
MakisH: What is the main reason we need the index here?
I guess we should be able to work with iterators, even if we only remove specific elements.
This already feels safer than before, but I think that the code can be further simplified and avoid the index calculations below.

vidulejs:
Hello again and sorry for the delay! I briefly had an issue with my system and couldn't run preCICE, but now I resolved it and could test the code.  Thanks for your comments David @davidscn !  I added the STL algorithm header in Adapter.h and made the changes you suggested.

Adapter.C
vidulejs: Good catch! I renamed the 'obj' variable to 'fieldObj' for clarity, and changed this loop to do without the index, because we do not actually need it here. Only place where the index is needed is in the last loop of this function.

Adapter.C
```diff
@@ -923,6 +924,60 @@ void preciceAdapter::Adapter::setupCheckpointing()
...
+{
+    // Check if checkpointed fields exist in OpenFOAM registry
+    // If not, remove them from the checkpointed fields vector
+
+    word fieldName;
+    std::vector<word> regFields;
+    std::vector<word> toRemove;
+
+#undef doLocalCode
+#define doLocalCode(GeomField, GeomField_, GeomFieldCopies_)                                                           \
```
MakisH: Add a comment on what the three arguments are.
Or rename `GeomField` to `GeomFieldType`?

Adapter.H
```diff
@@ -24,6 +24,8 @@
 // preCICE Solver Interface
 #include <precice/precice.hpp>
 
+#include <algorithm>
```
MakisH: This one should be in the `Adapter.C`.

Adapter.C
MakisH: This is now implemented as in the comment.

Adapter.C
```diff
@@ -930,37 +931,41 @@ void preciceAdapter::Adapter::pruneCheckpointedFields()
...
+        for (auto it = toRemoveIndices.rbegin(); it < toRemoveIndices.rend(); ++it)                                                           \
+        {                                                                                                                                     \
+            index = *it;                                                                                                                      \
+            DEBUG(adapterInfo("Removed " #GeomFieldType " : " + GeomFieldCopies_.at(index)->name() + " from the checkpointed fields list.")); \
+            GeomField_.erase(GeomField_.begin() + index);                                                                                     \
+            delete GeomFieldCopies_.at(index);                                                                                                \
+            GeomFieldCopies_.erase(GeomFieldCopies_.begin() + index);                                                                         \
+        }                                                                                                                                     \
```
vidulejs: I changed the iteration over `fieldName`(s) to iteration over the indices, since I need to access the index in `GeomField_` and `GeomFieldCopies_`. Now there is no confusion about getting the index from the reverse iterator.

Adapter.C
```diff
@@ -923,6 +925,64 @@ void preciceAdapter::Adapter::setupCheckpointing()
...
+    {                                                                                                                                         \
+        /* Iterate in reverse to avoid index shifting */                                                                                      \
+        for (auto it = toRemoveIndices.rbegin(); it != toRemoveIndices.rend(); ++it)                                                          \
+        {                                                                                                                                     \
+            index = *it;                                                                                                                      \
+            DEBUG(adapterInfo("Removed " #GeomFieldType " : " + GeomFieldCopies_.at(index)->name() + " from the checkpointed fields list.")); \
+            GeomField_.erase(GeomField_.begin() + index);                                                                                     \
+            delete GeomFieldCopies_.at(index);                                                                                                \
```
davidscn, line 966: Where does the corresponding deep copy of the object happen? Did you check for memory safety?

davidscn (approved):
Looks good, I haven't tested it though.

Adapter.C
vidulejs: The deep copies happen during `setupCheckpointing()` at the first simulation timestep. This pull request fixes an issue where preCICE tries to restore a checkpointed field, but it doesn't exist in the OpenFOAM registry anymore, i.e., OpenFOAM does not store this field anymore. Deleting the deep copy of the field and the pointer to the OpenFOAM object should be fine. The field copy memory is freed in the same way as in the destructor.

MakisH: Everything that happens here is only about indices. The deep copy only happens at the actual writing of the checkpoint. In this part of the code, we remove elements that we are not going to use anymore, so overall, I think we are safe.

MakisH (approved):
After some iterations of improvements, I think this is now ready to merge. I tested a previous state of this and it worked, and I ran the system tests as well (which do not trigger the actual issue). @vidulejs tested the latest state as well.

I am triggering the system tests once more for completeness, and merging.

MakisH:
I forgot that the system tests cannot currently run in forks (see https://github.com/precice/tutorials/issues/580). I still think that this is fine to merge. I will trigger the system tests in develop.

changelog-entries/344.md
```diff
@@ -0,0 +1,2 @@
+- Fixed a bug where a field was checkpointed, but doesn't exist anymore in OpenFOAM registry, causing a crash when reloading last timestep into OpenFOAM [#344](https://github.com/precice/openfoam-adapter/pull/344)
+- Added new function 'pruneCheckpointedFields()' which is called before readCheckpoint().
```
MakisH: This is not a user-facing change, we can remove it from the changelog:

```suggestion
- Fixed a bug where a field was checkpointed, but doesn't exist anymore in OpenFOAM registry, causing a crash when reloading last timestep into OpenFOAM. This is done by pruning checkpointed fields that do not (anymore) appear in the registry of objects at that timestep before reading the checkpoint. [#344](https://github.com/precice/openfoam-adapter/pull/344)
```

PR #346, opened 2025-01-10
Generic Read Write module

vidulejs:
This pull request implements a Generic module, which allows reading and writing fields by their OpenFOAM name. The surface and volume coupling of scalar and vector fields is implemented.

The new dictionary-based schema allows for additional options for each field, such as `solver_name`, `operation` and `flip-normal`. The default operation is `value`, which is just reading or writing the field directly. The operation `gradient` is currently not fully implemented, because of additional complexities. That should be a separate pull request.

For now the `operation surface-normal-gradient;` is implemented for surface scalar field writing. I we implement `operation gradient;` the data type will change from scalar to vector (and vector fields would become tensor fields). The adapter will throw an error if the gradient operation is not supported for that configuration.

Similarly, `flip-normal true` is also not yet implemented. However, that would be a more global change in the adapter. This also warrants a separate pull request. 

TODO list:

- [x] Vector and Scalar fields determined automatically
- [x] Volume coupling
- [x] Surface coupling
- [x] Surface scalar field normal gradient write 
- [x] Boundary conditions
- [x] Documentation v1
- [x] Merge develop into this branch


Make/options
```diff
@@ -7,6 +7,7 @@ EXE_INC = \
     -I$(LIB_SRC)/transportModels/twoPhaseMixture/lnInclude \
     -I$(LIB_SRC)/transportModels/interfaceProperties/lnInclude \
     -I$(LIB_SRC)/transportModels/immiscibleIncompressibleTwoPhaseMixture/lnInclude \
+    -I$(LIB_SRC)/thermoTools/derivedFvPatchFields/temperatureCoupledBase \
```
vidulejs: This is not necessary, artifact of past experimentation

docs/config.md
```diff
@@ -252,6 +252,39 @@ The FF module is still experimental and the boundary conditions presented here h
...
+
+As of OpenFOAM-adapter version 1.3.1, the old schema for readData and writeData is still used in `system/preciceDict`. This means that the coupled fields are specified by a single keyword, where the keywords are mapped to their OpenFOAM objects explicitly in the adapter, e.g., `Velocity` -> `U` or `Pressure` -> `p`. This is like a translation table between preCICE and OpenFOAM. In the future, changes to the schema will allow for the coupling data name and the solver name, i.e., the name as it appears in the solver, to be specified separately. In the meanwhile, to use the GENERIC module, it's required to use one keyword for both. For example, to use the GENERIC module in the channel-transport tutorial case, the coupling data has to be renamed from `Velocity` to `U` (case-sensitive) in `precice-config.xml`, `fluid-openfoam/system/preciceDict` and the other participant `transport-nutils/transport.py`.
```
MakisH: This now does not need the new schema, right? Comment needs update.

GENERIC/GENERIC.C
```diff
@@ -0,0 +1,183 @@
+#include "GENERIC.H"
```
vidulejs: Let me know if we need to change the capitalization of the module name.

GENERIC/GENERIC.C
```diff
@@ -0,0 +1,183 @@
...
+    return true;
+}
+
+bool preciceAdapter::GENERIC::GenericInterface::readConfig(const IOdictionary& adapterConfig)
+{
+    DEBUG(adapterInfo("Reading GENERIC module configuration..."));
+
+    // const dictionary& moduleDict = adapterConfig.subOrEmptyDict("GENERIC");
+
+    // TODO: Reduce code duplication and get the interfaces configuration from preciceAdapter:Adapter
```
vidulejs: This function below re-implements the preciceDict parsing from Adapter.C. I did that because I (1) didn't want to modify Adapter.C and (2) I will eventually need to get the new schema fields for `name`,  `solver_name` and `operation`. Changes to Adapter.C would come with the schema PR.

GENERIC/GENERIC.C
```diff
@@ -0,0 +1,183 @@
...
+
+            DEBUG(adapterInfo("    writeData    : "));
+            const wordList& writeDataDict = interfaceDict.get<wordList>("writeData");
+
+            for (const word& writeDatumEntry : writeDataDict)
+            {
+                word dataName = writeDatumEntry;
+
+                // Assume dataName is same as solver_name, see code for new schema below
+                struct fieldConfig fieldConfig;
```
vidulejs: Implemented a struct called fieldConfig to hold the new options for each coupled read and write datum.

GENERIC/GENERIC.C
```diff
@@ -0,0 +1,183 @@
...
+                DEBUG(adapterInfo("      - " + fieldConfig.name));
+            }
+            interfacesConfig_.push_back(interfaceConfig);
+        }
+    }
+
+    return true;
+}
+
+bool preciceAdapter::GENERIC::GenericInterface::addWriters(std::string dataName, Interface* interface)
```
vidulejs: This method implements searching for the field in the OpenFOAM registry based on the 'solver_name'. Currently I check whether it's a scalar or vector field.

GENERIC/GENERIC.C
```diff
@@ -0,0 +1,183 @@
...
+{
+    bool found = true; // Set to false later, if needed.
+
+    // Determine type of field
+    bool isScalarField;
+
+    // TODO: refactor to get rid of the inner loop
```
vidulejs: The inner loop is searching for where dataName is equal to dataName in the interfacesConfig. Could do without the loop with a "find" one-liner.

GENERIC/GENERIC.C
```diff
@@ -0,0 +1,183 @@
...
+            const wordList& writeDataDict = interfaceDict.get<wordList>("writeData");
+
+            for (const word& writeDatumEntry : writeDataDict)
+            {
+                word dataName = writeDatumEntry;
+
+                // Assume dataName is same as solver_name, see code for new schema below
+                struct fieldConfig fieldConfig;
+                fieldConfig.name = dataName;
+                fieldConfig.solver_name = dataName;
```
vidulejs: `solver_name` is currently equal to `name`. The new schema will support differentiating between the data name and the solver name.

GENERIC/GENERIC.C
```diff
@@ -0,0 +1,183 @@
...
+                    interface->addCouplingDataWriter(
+                        dataName,
+                        new ScalarFieldCoupler(mesh_, fieldConfig));
+                }
+                else
+                {
+                    interface->addCouplingDataWriter(
+                        dataName,
+                        new VectorFieldCoupler(mesh_, fieldConfig));
+                }
```
vidulejs: I don't even need the `bool isScalarField`. This could be just one if else if condition.

GENERIC/ReadWrite.C
```diff
@@ -0,0 +1,395 @@
...
+    const_cast<volScalarField*>(
+        &mesh.lookupObject<volScalarField>(fieldConfig.solver_name))),
+  mesh_(mesh),
+  fieldConfig_(fieldConfig)
+{
+    dataType_ = scalar;
+}
+
+
+// In the GENERIC module, only the value operation is supported for now, i.e., cannot write gradients.
```
vidulejs: This is related to the current schema limitation. In the new schema we will be able to specify `operation = gradient`. I wrote about this also in the documentation in config.md. There isn't a problem for reading on the OF adapter side though. Because when reading we must conform to the boundary condition, if it's `fixedValue` we can only write the value, same for the gradient BC.

module-generic/ReadWrite.C
```diff
@@ -0,0 +1,395 @@
...
+            {
+                boundaryPatch.gradient()[i] = buffer[bufferIndex++];
+            }
+        }
+        else
+        {
+            FatalErrorInFunction << "Unsupported boundary condition type " << bc.type() << exit(FatalError);
+        }
+
+        // // evaluate the boundary condition, i.e., do some calculation to obtain the actual value provided refValue
```
vidulejs, line 181: IIRC this is not needed, because the solver will call these methods for us.

GENERIC/ReadWrite.C
```diff
@@ -0,0 +1,395 @@
...
+
+        // // evaluate the boundary condition, i.e., do some calculation to obtain the actual value provided refValue
+        // boundaryPatch.updateCoeffs();
+        // boundaryPatch.evaluate();
+    }
+}
+
+bool preciceAdapter::GENERIC::ScalarFieldCoupler::isLocationTypeSupported(const bool meshConnectivity) const
+{
+    // TODO
```
vidulejs: Mesh connectivity is not tested, but it should work - just the same as in the other modules.

Adapter.C
```diff
@@ -138,17 +143,19 @@ bool preciceAdapter::Adapter::configFileRead()
                     }
 
                     DEBUG(adapterInfo("    writeData    : "));
-                    auto writeData = interfaceDict.get<wordList>("writeData");
-                    for (auto writeDatum : writeData)
+                    const wordList& writeDataDict = interfaceDict.get<wordList>("writeData");
+                    for (const word& writeDatumEntry : writeDataDict)
```
vidulejs: This can be reverted. I had changed it in one of the WIP commits. I just specified the types.

vidulejs (commented):
Most of the code is the same as in the other modules (FF and CHT). I added comments on where things differ and where there's new implementations.

docs/config.md
MakisH: To address after merging back from develop, to integrate #354

GENERIC/GENERIC.C
MakisH: Yes, let's better do `module-generic/Generic.H`.

Also rename in the documentation.

For the preciceDict, maybe `modules (generic);`.

GENERIC/GENERIC.C
MakisH: Update when integrating #354

GENERIC/ReadWrite.C
```diff
@@ -0,0 +1,395 @@
+#include "ReadWrite.H"
+#include "primitivePatchInterpolation.H"
+#include "fvCFD.H"
+
+#include "fixedValueFvPatchFields.H"
+#include "fixedGradientFvPatchFields.H"
+// #include "mixedFvPatchFields.H"
```
MakisH: Let's start without this, remove the commented-out header, and open an issue for Robin conditions, or otherwise keep track of this in the general module issue.

module-generic/ReadWrite.C
MakisH: If you think this needs further action, please open an issue.

GENERIC/ReadWrite.C
MakisH: Then let's start without it, but document the limitation and add an error, if not too complicated.

GENERIC/ReadWrite.C
```diff
@@ -0,0 +1,395 @@
...
+        else
+        {
+            FatalErrorInFunction << "Unsupported boundary condition type " << bc.type() << exit(FatalError);
+        }
+    }
+}
+
+bool preciceAdapter::GENERIC::VectorFieldCoupler::isLocationTypeSupported(const bool meshConnectivity) const
+{
+    // TODO
```
MakisH: Implement error message for non-supported locations/connectivity.

Adapter.C
```diff
@@ -290,22 +306,15 @@ void preciceAdapter::Adapter::configure()
...
-                    inModules++;
-                }
+                if (CHTenabled_ && CHT_->addWriters(dataName, interface)) inModules++;
 
                 // Add FSI-related coupling data writers
-                if (FSIenabled_ && FSI_->addWriters(dataName, interface))
-                {
-                    inModules++;
-                }
+                if (FSIenabled_ && FSI_->addWriters(dataName, interface)) inModules++;
```
MakisH: The braces in the current develop branch are in purpose (#175)

Adapter.H
```diff
@@ -122,6 +127,8 @@ private:
 
     // NOTE: Add here a pointer for your new module object
 
+    GENERIC::GenericInterface* GENERIC_ = NULL;
+
```
MakisH: Swap order

module-generic/ReadWrite.C
```diff
@@ -0,0 +1,395 @@
...
+    dataType_ = scalar;
+}
+
+
+// In the GENERIC module, only the value operation is supported for now, i.e., cannot write gradients.
+std::size_t preciceAdapter::GENERIC::ScalarFieldCoupler::write(double* buffer, bool meshConnectivity, const unsigned int dim)
+{
+    int bufferIndex = 0;
+
+    if (this->locationType_ == LocationType::volumeCenters)
```
MakisH, line 64: In our meeting, you pointed out that we should document that both surface and volume coupling are possible for a single interface (cell centers, face centers, and nodes).

This is already true for other modules as well (such as the CHT). Feel free to update the rest of the documentation if you see needed.

Adapter.C
vidulejs: BTW, we have braces for the writeData block, but we're missing them for the readData block just below.

GENERIC/GENERIC.C
vidulejs: Fixed! Now parsing is done only in Adapter.C.

Adapter.C
vidulejs, line 390: I think we can modify the logic here for the Generic module. If the other modules can read the same field as the Generic module, then give priority to them. 

For example, if CHT can couple Temperature, then don't couple it with the Generic module. The change is easy, just add an if condition:

```cpp
    // Only add Generic interface if not found in other modules
    if (inModules == 0)
    {
            if (genericModuleEnabled_ && Generic_->addWriters(FieldConfig, interface))
            {
                inModules++;
            };
    }
```

module-generic/Generic.C
```diff
@@ -0,0 +1,118 @@
...
+            interface->addCouplingDataWriter(
+                FieldConfig.name,
+                new VectorFieldCoupler(mesh_, FieldConfig));
+        }
+        else
+        {
+            found = false;
+            std::string msg = "Generic module: Data \"" + FieldConfig.name + "\", solver name: \"" + FieldConfig.solver_name + "\" not found!\n";
+            msg += "Available fields: " + availableVolScalarFields + availableVolVectorFields;
+            adapterInfo(msg, "warning");
```
vidulejs, line 71: I thought I would add a message here if we cannot find "solver_name" in the OpenFOAM field list. Not sure if it should be at level "warning" or "info". Either way we will exit with an error in `Adapter::configure()`, because we cannot couple this field.

```log
---[preciceAdapter] [DEBUG] Adding coupling data writers...
---[preciceAdapter] [DEBUG] Added writer: Velocity
---[preciceAdapter] [DEBUG] Added writer: Pressure
--> FOAM Warning : 
    in file Utilities.C at line 18
    Warning in the preCICE adapter: 
Generic module: Data "TestData", solver name: "Test" not found!
Available fields: nu p U 

---[preciceAdapter] [DEBUG] Destroying the preCICE solver interface...
---[precice]  Implicitly finalizing in destructor
---[preciceAdapter] [DEBUG] Deleting the interfaces...
---[preciceAdapter] [DEBUG] Destroying the Generic module...


--> FOAM FATAL ERROR: (openfoam-2406 patch=241212)
Error in the preCICE adapter: 
I don't know how to write "TestData". Maybe this is a typo or maybe you need to enable some adapter module?
```

MakisH: I would say that, if a user asks for a field that is not available, there should be an error. A warning might go unnoticed.

Adapter.C
vidulejs: Implemented.

module-generic/ReadWrite.C
```diff
@@ -0,0 +1,399 @@
...
+            // Get the surface normal gradient boundary patch
+            scalarField gradientPatch(scalarField_->boundaryField()[patchID].snGrad());
+
+            // For every cell of the patch
+            forAll(gradientPatch, i)
+            {
+                buffer[bufferIndex++] = gradientPatch[i];
+            }
+        }
+        return bufferIndex;
```
vidulejs, line 61: Allow writing scalar field surface normal gradients

Adapter.H
```diff
@@ -37,15 +39,6 @@ class Adapter
...
 private:
-    //- field options
-    struct FieldConfig
-    {
-        std::string name;
-        std::string solver_name;
-        std::string operation;
-        bool flip_normal;
-    };
-
```
vidulejs, line 48: Moved struct definition to Interface.H

module-generic/ReadWrite.C
```diff
@@ -21,12 +21,41 @@ preciceAdapter::Generic::ScalarFieldCoupler::ScalarFieldCoupler(
     dataType_ = scalar;
 }
 
+void preciceAdapter::Generic::ScalarFieldCoupler::initialize()
+{
+    if (FieldConfig_.operation == "gradient")
```
MakisH: Let's rename the option to `surface-normal-gradient`. We also have other fields that would be a full gradient.

docs/config.md
```diff
@@ -259,37 +259,51 @@ When coupling face flux `Phi`, usually no specific boundary condition needs to b
 
 ## Generic module
 
-The Generic module is an experimental addition, which would theoretically allow for any field to be coupled by its name. Currently, volume and surface coupling of scalar and vector fields is supported, for example, Velocity and Pressure. Explicitly setting `locations volumeCenters;` in preciceDict is required for volume coupling. If omitted, it defaults to surface coupling on faceCenters. When writing to OpenFOAM, the Generic module automatically detects the boundary condition type (`fixedValue`, `fixedGradient`) of the coupled patch to apply the received data.
+The Generic module is an experimental addition that allows coupling any field by its name. It supports volume and surface coupling of scalar and vector fields (e.g., `Velocity`, `Pressure`, `Temperature`). If multiple modules are enabled, the Generic module will handle only those fields which have not been defined in the other modules. For example, if FF is also enabled, then fields whose data name starts with `Velocity` will be handled by the FF module instead.
```
MakisH, line 262: We can also state somewhere that in the future we plan to have something like this as the default, with edge cases explicitly implemented.

docs/config.md
```diff
@@ -259,37 +259,51 @@ When coupling face flux `Phi`, usually no specific boundary condition needs to b
...
 
     writeData
     (
-        U
+      Velocity1
+      {
+        name        Velocity;   // Data name as defined in preCICE config
+        solver_name U;          // Optional: Field name in OpenFOAM (defaults to same as 'name')
+        operation   value;      // Optional: Operation to perform on data (defaults to 'value')
+        flip-normal false;      // Optional: Flip the normal direction (defaults to 'false')
```
MakisH, line 287: We need a note that this is not yet implemented. As you said, we can continue with the `flip-normal` in the next PR.

module-generic/ReadWrite.C
```diff
@@ -0,0 +1,399 @@
...
+        {
+            auto& boundaryPatch = refCast<fixedGradientFvPatchScalarField>(bc);
+            forAll(boundaryPatch, i)
+            {
+                boundaryPatch.gradient()[i] = buffer[bufferIndex++];
+            }
+        }
+        else
+        {
+            FatalErrorInFunction << "Unsupported boundary condition type " << bc.type() << exit(FatalError);
```
MakisH: Here and maybe in other places, we could use the adapter utility for reporting errors.

docs/config.md
```diff
@@ -257,6 +257,53 @@ The FF module is still experimental and the boundary conditions presented here h
...
+    );
+  };
+};
+```
+
+Explicitly setting `locations volumeCenters;` in preciceDict is required for volume coupling. If omitted, it defaults to surface coupling on `faceCenters`. Volume coupling data is always assumed as values for now, i.e., cannot set volume gradient.
+
+When reading on the OpenFOAM side, the Generic module automatically detects the boundary condition type of the coupled patch to apply the received data. The boundary condition must be respected, therefore the data received is applied either as a `fixedValue` or `fixedGradient` boundary condition, determined by the type set in the `0/` files.
+
+The `operation gradient;` is currently only supported for scalar field surface writing, in which case it actually performs the surface normal gradient operation (data type is still scalar).
```
MakisH: Will need to be updated to `surface-normal-gradient`.

Adapter.C
MakisH: Inconsistent, indeed. Please add.

docs/config.md
vidulejs: Added a sentence at the end. Did you mean the blue html formatted note?

module-generic/ReadWrite.C
vidulejs: Renamed. Also changed in the documentation.

docs/config.md
MakisH: I can just mention this in the release notes, no need for anything else.

MakisH (approved):
I had another quick look, I think this is fine for now! Triggering the system tests just in case, since the config parsing was touched. I will merge after they succeed.

MakisH:
@vidulejs the longer system tests suite [failed overnight](https://github.com/precice/tutorials/actions/runs/22337832217) and the error for the [multiple-perpendicular-flaps tutorial](https://github.com/precice/tutorials/tree/develop/multiple-perpendicular-flaps) is:

```
--> FOAM FATAL ERROR: (openfoam-2512)
Size of newPoints 0 does not correspond to current mesh points size 9528

    From virtual void Foam::polyMesh::movePoints(const Foam::pointField&)
    in file meshes/polyMesh/polyMesh.C at line 1168.
```

Do you think this could be related to this PR?

[fluid-openfoam.log](https://github.com/user-attachments/files/25515831/fluid-openfoam.log)

PR #350, opened 2025-01-30
Document running actions locally and update custom build CI options

MakisH:
Before we can maintain the adapter for different OpenFOAM versions again, we need a system to build the adapter for different versions locally. We already have a CI workflow for that, but it is a bit cumbersome running it always on GitHub.

This PR:

- Documents how to run GitHub Actions and specifically this workflow locally with act.
- Adds an options file for the arguments, to not always pass them via the CLI.
- Documents which OpenFOAM versions are available in which Ubuntu versions. This is something I often have to look up again.
- Updates the default argument values for the workflow.

TODO list:

@vidulejs maybe a nice first PR to review and try out the workflow locally. I am looking forward to your comments on what works, what not, and what I could improve.


CONTRIBUTING.md
```diff
@@ -28,6 +28,25 @@ There are also a few additional workflows that can be triggered manually:
...
+
+You can also run GitHub Actions locally with [act](https://github.com/nektos/act).
+
+List the workflows with `act --list`. To simulate a push event, run `act push`. Some workflows might work on GitHub Actions but not on act, in which case you could try running specific workflows with, for example, `act -j 'build'`.
+
+To trigger the custom build workflow:
+
+1. Switch to the directory `.github/workflows/`
+2. Edit the inputs in the `build-custom.input` file.
+3. [Generate a GitHub access token](https://github.com/settings/personal-access-tokens) and add it to a `.secrets` file with content `GITHUB_TOKEN=<your-token>`.
```
MakisH, line 41: This step should not be needed, but there seems to be a related [issue in act](https://github.com/nektos/act/issues/233).

CONTRIBUTING.md
```diff
@@ -28,6 +28,25 @@ There are also a few additional workflows that can be triggered manually:
 
 Members of the repository can trigger these workflows in the "Actions" tab.
 
+### Running the CI locally
+
+You can also run GitHub Actions locally with [act](https://github.com/nektos/act).
+
+List the workflows with `act --list`. To simulate a push event, run `act push`. Some workflows might work on GitHub Actions but not on act, in which case you could try running specific workflows with, for example, `act -j 'build'`.
```
MakisH, line 35: In particular, I could run the first tutorial, but not more. It was as if the second solver was not starting.

.github/workflows/build-custom.yml
```diff
@@ -4,13 +4,13 @@ on:
...
         description: 'Virtual Environment'
-        default: 'ubuntu-18.04'
+        default: 'ubuntu-22.04'
         required: true
         type: choice
         options:
+          - ubuntu-24.04
           - ubuntu-22.04
           - ubuntu-20.04
-          - ubuntu-18.04
```
MakisH, line 13: [GitHub-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners) with Ubuntu 18.04 are not available anymore, hence we cannot test OpenFOAM 5 and 6 anymore.

Locally, there is a Docker image that is picked:

```
Start image=catthehacker/ubuntu:act-18.04
```

but I am later on getting another error:

```
[Custom build (manual)/build] ⭐ Run Main Check out repository
[Custom build (manual)/build]   🐳  docker cp src=/home/gc/.cache/act/actions-checkout@v4/ dst=/var/run/act/actions/actions-checkout@v4/
[Custom build (manual)/build]   🐳  docker exec cmd=[node /var/run/act/actions/actions-checkout@v4/dist/index.js] user= workdir=
| OCI runtime exec failed: exec failed: unable to start container process: exec: "node": executable file not found in $PATH: unknown
[Custom build (manual)/build]   ❌  Failure - Main Check out repository
[Custom build (manual)/build] exitcode '126': failure
```

I think we will have to live without testing OpenFOAM.org earlier than 7 for now.

CONTRIBUTING.md
```diff
@@ -28,6 +28,25 @@ There are also a few additional workflows that can be triggered manually:
...
+
+To trigger the custom build workflow:
+
+1. Switch to the directory `.github/workflows/`
+2. Edit the inputs in the `build-custom.input` file.
+3. [Generate a GitHub access token](https://github.com/settings/personal-access-tokens) and add it to a `.secrets` file with content `GITHUB_TOKEN=<your-token>`.
+4. Start the build using:
+
+   ```shell
+   act -W build-custom.yml --input-file build-custom.input --secret-file .secrets --artifact-server-path $PWD/.artifacts
```
vidulejs, line 46: artifact-server-path sounds misleading. Sounds like there is some remote server, but it saves logs locally. 
Without this argument the log files will not be saved. This seems like a known issue of act, see issue ["environment variables are missing"](https://github.com/nektos/act/issues/329)

Without this argument I get the error:
```bash
[Custom build (manual)/build]   ❗  ::error::Unable to get the ACTIONS_RUNTIME_TOKEN env variable
[Custom build (manual)/build]   ❌  Failure - Main Archive logs
[Custom build (manual)/build] exitcode '1': failure
```

CONTRIBUTING.md
vidulejs: The quickstart tutorial is not advancing because the other solver cannot be compiled. It says it could not find cmake. I can take a look into this.

.github/workflows/build-custom.yml
vidulejs: Why use a container with act? Wouldn't it make more sense to get OpenFOAM containers and install act after? Ah, I see there is a [setup-openfoam](https://github.com/gerlero/setup-openfoam) repo for GitHub actions.

CONTRIBUTING.md
MakisH: Clarified this, thanks!

.github/workflows/build-custom.yml
MakisH: Exactly, using the `setup-openfoam` action allows us to not maintain such Docker images.

Note that now GitHub has also removed the `ubuntu-20.04` runners. However, I get the impression that these are supported longer in act, so I would keep this option offered for now.

CONTRIBUTING.md
MakisH: I added a step to install CMake for the Quickstart tutorial. Thanks for spotting and reporting this.

.github/workflows/build-custom.yml
MakisH: The exact issue with starting this is related to the checkout action.
I would need to look into why this works in the newer containers but not in ubuntu-18.04.

MakisH: The setup-openfoam action does not officially support Ubuntu 18.04 anyway. I consider this a wont-fix for now, and will move on to try to port to OpenFOAM 7, which is available on Ubuntu 20.04.

Alternatively, we will need a custom workflow to build from source for older versions.

MakisH: One way to get a newer nodejs would have been:

```yaml
    - name: nodejs
      if: ${{ contains(github.event.inputs.virtualEnvironment, 'ubuntu-18.04') }} 
      run: |
        curl -fsSL https://deb.nodesource.com/setup_20.x | bash - 
        apt-get install -y nodejs
        node -v
```

but this stumbles upon missing dependencies.

PR #352, opened 2025-03-06
Set `ADAPTER_TARGET_DIR` and `ADAPTER_CFLAGS` (`ADAPTER_PREP_FLAGS`) via environment variables

BenjaminRodenberg:
## Changes

* Will use value of `ADAPTER_TARGET_DIR` if set
* Will use `FOAM_USER_LIBBIN` as `ADAPTER_TARGET_DIR` if `ADAPTER_TARGET_DIR` is unset

## Motivation

I tried to install the adapter as root user under `FOAM_LIBBIN` and this did not work (some docker environment). As a workaround I'm currently temporarily overwriting `FOAM_USER_LIBBIN` with `FOAM_LIBBIN`.

TODO list:

### DIFF
--- a/Allwmake
+++ b/Allwmake
@@ -6,7 +6,7 @@ set -e -u
 # Optional: Preprocessor flags
 # "-DADAPTER_DEBUG_MODE" enables debug messages
 # "-DADAPTER_ENABLE_TIMINGS" enables time measurements
-ADAPTER_PREP_FLAGS="${ADAPTER_PREP_FLAGS:-}"
+ADAPTER_PREP_FLAGS="${PRECICE_OPENFOAM_PREP_FLAGS:-}"
 
 # Build command and options
 # In order to compile with multiple threads, set the environment variable WM_NCOMPPROCS,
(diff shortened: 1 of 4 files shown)


MakisH:
Sounds like a good idea, thanks!

Since this is an external environment variable, I would suggest a more descriptive name, such as `OPENFOAM_ADAPTER_TARGET_DIR`.

I also see this more as a support for something new (getting this information from the environment), not as a fix for something broken. The current workflow is to manually change these variables in the header of the file.

If you want, I can take this over and directly modify things.

MakisH (approved):
I extended this with to address the general question "how do we make the build configurable with external environment variables".

For this reason, I renamed the variables to start with `PRECICE_OPENFOAM_`.

Also, I never liked the name `ADAPTER_PREP_FLAGS` because it is completely non-standard and inaccurate. I renamed it to `ADAPTER_CFLAGS`, mimicking the pkg-config naming. They still don't mean the same thing, but I think this is a better name.

I also shortened the documentation a bit to make it more skimmable.

PR #354, opened 2025-03-20
Parse both legacy and schema-based preciceDict configuration files

vidulejs:
The new preCICE schema will have [additional options for fields](https://github.com/precice/preeco-orga/pull/38), to specify `solver_name` and `operation`. This means that readData and writeData in preciceDict would be [dictionaries instead of lists](https://github.com/precice/preeco-orga/issues/40), so that these additional options can be parsed. 

EDIT 05/02/2026: We decided that the new schema for read/writeData should be a list of dictionaries (instead of dicts of dicts). 

 For example:

```openfoam
  Interface1
  {
    mesh              Fluid-Mesh;
    patches           (interface);
    
    readData
    (
      Heat-Flux1
      {
      	name      	      	Heat-Flux;
      	solver_name      	T;
      	operation      	    gradient;
        flip-normal         true;
      }
    );
    
    writeData
    (
      Temperature1
      {
      	name      	      	Temperature;
      	solver_name      	T;
      	operation      	    value;
      }

      Temperature2
      {
      	name      	      	Temperature;
      }
    );
  };
```

I implemented a new `struct fieldConfig` to hold `name` (data name), `solver_name`, and `operation`. 

~~I tried to look into ways to keep backwards compatibility with the lists, but this may not be desirable. The new schema warrants breaking changes. In order to read old config files, we'd have to check if it's a list or a dict. OpenFOAM does not make this easy... I tried to use the `ITstream` (Input Token Stream) to check whether the token is `(` for list or `{` for dict. However the `ITstream & dictionary.lookup(word &keyword)` works only if i'm looking up a list and not a dict. I would need to write a custom parser. [Similar issue on cfd-online forums.](https://www.cfd-online.com/Forums/openfoam-solving/60439-read-new-constant-solver.html)~~

TODO list:

- [x] Parsing lists of dictionaries
- [x] `name` field is mandatory
- [x] Backwards compatibility with lists of words (wordList)
- [x] Added bool `flip-normal` 
- [ ] Draft PR in the tutorials


MakisH:
For the review, I think we would need also a corresponding draft PR in the tutorials, porting at least one case (e.g., the flow-over-heated-plate).

MakisH:
Based on the latest state (list of dictionaries), it looks like we need a different, solution as described in https://github.com/precice/preeco-orga/pull/38#issuecomment-2751151039

MakisH:
As `operation`, we could distinguish between `gradient` and `normal-gradient` (writing to be checked for consistency), to account for https://github.com/precice/openfoam-adapter/pull/380

MakisH:
To release it with v1.4.0, we need to implement it in a backwards-compatible way.

vidulejs:
I have tested the new schema parsing of lists of dicts. I found a subtle issue that might be encountered by a lot of users. In OpenFOAM, the last list entry cannot have a trailing semicolon after the closing brace `};`

For example, here the last closing bracket on Velocity2 has the semicolon, which will cause an error.
```preciceDict
    readData
    (
      Velocity1
      {
        name Velocity;
        solver_name U;
        operation value;
      };

      Velocity2
      {
        name Velocity;
        operation gradient;
      };

    );
```


It will cause a cryptic error without pointing the user towards the preciceDict:

```log
---[preciceAdapter] Reading preciceDict...
---[preciceAdapter] [DEBUG]   precice-config-file : ../precice-config.xml
---[preciceAdapter] [DEBUG]   participant name    : Fluid
---[preciceAdapter] [DEBUG]   modules requested   : 
---[preciceAdapter] [DEBUG]   - FF

---[preciceAdapter] [DEBUG]   interfaces : 
---[preciceAdapter] [DEBUG]   - mesh         : Fluid-Mesh
---[preciceAdapter] [DEBUG]     locations    : volumeCenters
---[preciceAdapter] [DEBUG]     connectivity : 0
---[preciceAdapter] [DEBUG]     patches      : 
---[preciceAdapter] [DEBUG]     cellSets      : 
---[preciceAdapter] [DEBUG]       - box1
---[preciceAdapter] [DEBUG]     writeData    : 
---[preciceAdapter] [DEBUG]     readData     : 
--> FOAM Warning : 
    in file Utilities.C at line 50
    Error (deferred - will exit later) in the preCICE adapter: 
attempt to read beyond EOF
```

Is there a way we could catch this error and print something more meaningful?

EDIT: With the new token stream parsing function we could catch the semicolons and do something.

Adapter.C
```diff
@@ -16,6 +16,59 @@ preciceAdapter::Adapter::Adapter(const Time& runTime, const fvMesh& mesh)
...
+            DEBUG(adapterInfo("        operation  : " + fieldConfig.operation));
+        }
+        stream >> _t; // Last token ')'
+    }
+    else
+    {
+        adapterInfo(listName + " must be a list", "error");
+    }
+}
+
```
vidulejs: Created a new member function `Adapter::readFieldConfigs`, which will handle parsing logic of writeData and readData lists. Since I am parsing a token stream I have much more flexibility now. For example, I now have backwards compatibility with `wordList`s. I tested a few tutorial cases (with the current schema) and it seems to work fine.

vidulejs:
@MakisH I seem to have solved the backwards compatibility issue. Could we run system tests on this branch? Although I added additional debug statements to the log output:
```cpp
            DEBUG(adapterInfo("        solver_name: " + fieldConfig.solver_name));
            DEBUG(adapterInfo("        operation  : " + fieldConfig.operation));
```
That could fail the system tests (?)

MakisH:
> @MakisH I seem to have solved the backwards compatibility issue. Could we run system tests on this branch? Although I added additional debug statements to the log output:
> 
> ```c++
>             DEBUG(adapterInfo("        solver_name: " + fieldConfig.solver_name));
>             DEBUG(adapterInfo("        operation  : " + fieldConfig.operation));
> ```
> 
> That could fail the system tests (?)

Triggered (just by adding, or later removing and re-adding the `trigger-system-tests` label on this PR).

The log files are not compared, only the exported interface meshes.

Note that there are some conflicts on `Adapter.C` at the moment. I have not yet looked at anything else.

edit: Looks like the conflicts need to be resolved first for the system tests to start.

MakisH:
Reviewed in a videocall. Still needs a changelog entry and an example in the documentation, mentioning that this is not yet used.

Also needs to implement the `flip-normal` boolean.

The system tests pass, and this works now both with legacy configs, as well as with a list of dictionaries.

MakisH:
Question: Based on this PR, How would you write a new-style `preciceDict` to couple a field known to preCICE as `DISPLACEMENTS` but that the adapter needs to bind to `FSI/Displacement.C`? The `solver_name` is not enough in this case, right?

Adapter.C
```diff
@@ -133,21 +196,20 @@ void preciceAdapter::Adapter::configFileRead()
                     return;
                 }
 
-                DEBUG(adapterInfo("    writeData    : "));
-                auto writeData = interfaceDict.lookupOrDefault<wordList>("writeData", wordList());
-                for (auto writeDatum : writeData)
+                if (interfaceDict.found("writeData"))
```
vidulejs, line 199: The reason I replaced `auto writeData = interfaceDict.lookupOrDefault<wordList>("writeData", wordList());` with `if (interfaceDict.found("writeData"))` is because it's no longer a `wordList`, ~~but a new type.~~ EDIT: It's not actually a new type, or rather, such a type does not exist in OpenFOAM. OpenFOAM can have the `wordList` type or `PtrList` type (for lists of objects of the same type, like the list of dictionaries). I have here just a string of tokens for which I have the custom parser for, to support the word list, dictionary list or both simultaneously.

vidulejs:
> Question: Based on this PR, How would you write a new-style `preciceDict` to couple a field known to preCICE as `DISPLACEMENTS` but that the adapter needs to bind to `FSI/Displacement.C`? The `solver_name` is not enough in this case, right?

Yes, it's not enough. In `Adapter::configure()`, when we add the interfaces for the existing modules, we have, e.g., 
```cpp
if (FSIenabled_ && FSI_->addWriters(dataName, interface)
```
where `dataName` would be `DISPLACEMENTS` in this case. The existing modules don't use `solver_name` (yet). I suggest we open another issue for this case.

docs/config.md
```diff
@@ -142,6 +142,25 @@ The `writeData` and `readData` names are case-insensitive since v1.4.0. This mea
...
+        solver_name U;          // Optional: Field name in OpenFOAM (defaults to same as 'name')
+        operation   value;      // Optional: Operation to perform on data (defaults to 'value')
+        flip-normal false;      // Optional: Flip the normal direction (defaults to 'false')
+    }
+
+    // Legacy word entry
+    Temperature
+);
+```
+
```
vidulejs: Let me know what you think about this documentation. I would like to stress more that the new options are not yet functionally supported, but just simply parsed. If the GENERIC module makes it to v1.4.0, then I could update this statement. But the new schema has to be merged before the GENERIC module in order to make the options functional in the new module. 

Also, the placement of this documentation may be awkward. Maybe you have a different idea on where to put this?

docs/config.md
```diff
@@ -142,6 +142,25 @@ The `writeData` and `readData` names are case-insensitive since v1.4.0. This mea
 Additionally, since earlier versions, only the beginning of the name needs to match: `Temperatures0` is valid and matched to the temperature reader/writer, for example.
 {% endtip %}
 
+Since v1.4.0, `readData` and `writeData` support parsing the new dictionary schema. Additional options (`solver_name`, `operation` and `flip-normal`) can be specified in the dictionaries . For example, the data `name` as known by preCICE can be different than the `solver_name` known by OpenFOAM. However, the new options are not yet functionally supported by the current modules FF, CHT and FSI. The legacy word list format is still supported, and both formats can even be mixed:
```
```suggestion
We are currently working in porting the adapter configuration file to the new [adapter configuration schema](https://github.com/precice/preeco-orga/tree/main/adapter-config-schema). Since v1.4.0, `readData` and `writeData` support parsing both the new format. Additional options (`solver_name`, `operation` and `flip-normal`) can be specified in the dictionaries . For example, the data `name` as known by preCICE can be different than the `solver_name` known by OpenFOAM. However, the new options are not yet functionally supported by the current modules FF, CHT and FSI. The legacy word list format is still supported, and both formats can even be mixed:
```

docs/config.md
MakisH: Good point, I moved the documentation to the bottom, next to the current comment on porting to the current configuration format, to give this less prominence, but also to point to the future.

MakisH (approved):

MakisH:
> Draft PR in the tutorials

I don't think we need this at the moment. Not before we have actual functionality implemented.

PR #356, opened 2025-03-20
Extend porting guide with more points

MakisH:
Still not sure how `writeEntry` is supposed to look like in OpenFOAM6. See https://github.com/OpenFOAM/OpenFOAM-7/commit/d41166187a0c32ab712026c995b0790cb4471194.


MakisH:
Merging to get the information in the documentation, which still needs to be further extended. The compatibility with other versions remains a larger issue to work on.

PR #358, opened 2025-03-21
Avoid ambiguous type conversion when accessing tmp<vectorField>

MakisH:
Reported by a user using the Intel compiler (OneAPI 2023.2, RHEL 8, OpenFOAM v2406) in the [forum](https://precice.discourse.group/t/failing-to-compile-the-openfoam-adapter-at-moduleff-o/2240/3).

Resolves errors such as:

```
In file included from FF/ModuleFF.C:3:
FF/Velocity.C:95:25: error: conversion from 'tmp<vectorField>' (aka 'tmp<Field<Vector<double>>>') to 'vectorField' (aka 'Field<Vector<double>>') is ambiguous
            vectorField n = U_->boundaryField()[patchID].patch().nf();
                        ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/leonardo/prod/spack/05/install/0.21/linux-rhel8-icelake/oneapi-2023.2.0/openfoam-2406-psbdrns5bo35czf4hemtwyre65rfdf32/src/OpenFOAM/lnInclude/tmp.H:320:9: note: candidate function
        operator const T&() const { return cref(); }
        ^
/leonardo/prod/spack/05/install/0.21/linux-rhel8-icelake/oneapi-2023.2.0/openfoam-2406-psbdrns5bo35czf4hemtwyre65rfdf32/src/OpenFOAM/lnInclude/Field.H:266:16: note: candidate constructor
        inline Field(const tmp<Field<Type>>& tfld);
               ^
```

PR #359, opened 2025-03-21
Avoid ambiguous type conversion when accessing tmp<vectorField>

MakisH:
Reported by a user using the Intel compiler (OneAPI 2023.2, RHEL 8, OpenFOAM v2406) in the [forum](https://precice.discourse.group/t/failing-to-compile-the-openfoam-adapter-at-moduleff-o/2240/3).

Resolves errors such as:

```
In file included from FF/ModuleFF.C:3:
FF/Velocity.C:95:25: error: conversion from 'tmp<vectorField>' (aka 'tmp<Field<Vector<double>>>') to 'vectorField' (aka 'Field<Vector<double>>') is ambiguous
            vectorField n = U_->boundaryField()[patchID].patch().nf();
                        ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/leonardo/prod/spack/05/install/0.21/linux-rhel8-icelake/oneapi-2023.2.0/openfoam-2406-psbdrns5bo35czf4hemtwyre65rfdf32/src/OpenFOAM/lnInclude/tmp.H:320:9: note: candidate function
        operator const T&() const { return cref(); }
        ^
/leonardo/prod/spack/05/install/0.21/linux-rhel8-icelake/oneapi-2023.2.0/openfoam-2406-psbdrns5bo35czf4hemtwyre65rfdf32/src/OpenFOAM/lnInclude/Field.H:266:16: note: candidate constructor
        inline Field(const tmp<Field<Type>>& tfld);
               ^
```

(same as #358, different repository)

Edit: The system tests pass after fixing the Git reference of preCICE, but I realize that they don't test anything related to the fluid-fluid module yet. I have, however, run the partitioned-pipe tutorial locally and it looks correct.

### DIFF
--- a/FF/BoundaryConditions/coupledPressure/coupledPressureFvPatchField.C
+++ b/FF/BoundaryConditions/coupledPressure/coupledPressureFvPatchField.C
@@ -99,7 +99,7 @@ void Foam::coupledPressureFvPatchField::updateCoeffs()
     const scalarField& phip = phi->boundaryField()[this->patch().index()];
     const Foam::volVectorField* U = &db().lookupObject<volVectorField>(uName_);
     const vectorField& Up = U->boundaryField()[this->patch().index()];
-    const vectorField n = this->patch().nf();
+    const vectorField n = this->patch().nf().cref();
 
     int t0 = this->patch().boundaryMesh().mesh().time().startTimeIndex();
     int t = this->patch().boundaryMesh().mesh().time().timeIndex();
--- a/FF/BoundaryConditions/coupledVelocity/coupledVelocityFvPatchField.C
+++ b/FF/BoundaryConditions/coupledVelocity/coupledVelocityFvPatchField.C
@@ -95,7 +95,7 @@ void Foam::coupledVelocityFvPatchField::updateCoeffs()
     }
     const Foam::surfaceScalarField* phi = &db().lookupObject<surfaceScalarField>(phiName_);
     const scalarField& phip = phi->boundaryField()[this->patch().index()];
-    const vectorField n = this->patch().nf();
+    const vectorField n = this->patch().nf().cref();
 
     int t0 = this->patch().boundaryMesh().mesh().time().startTimeIndex();
     int t = this->patch().boundaryMesh().mesh().time().timeIndex();
--- a/FF/Velocity.C
+++ b/FF/Velocity.C
@@ -92,7 +92,7 @@ std::size_t preciceAdapter::FF::Velocity::write(double* buffer, bool meshConnect
         if (fluxCorrection_)
         {
             scalarField phip = phi_->boundaryFieldRef()[patchID];
-            vectorField n = U_->boundaryField()[patchID].patch().nf();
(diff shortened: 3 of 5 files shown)


MakisH:
@vidulejs I am not exactly sure if `cref()` is the right function to call here (suggested by the compiler error). Does this change make sense to you?

FF/BoundaryConditions/coupledPressure/coupledPressureFvPatchField.C
```diff
@@ -99,7 +99,7 @@ void Foam::coupledPressureFvPatchField::updateCoeffs()
     const scalarField& phip = phi->boundaryField()[this->patch().index()];
     const Foam::volVectorField* U = &db().lookupObject<volVectorField>(uName_);
     const vectorField& Up = U->boundaryField()[this->patch().index()];
-    const vectorField n = this->patch().nf();
+    const vectorField n = this->patch().nf().cref();
```
vidulejs, line 102:
```C++
const Foam::tmp<Foam::vectorField> n = this->patch().nf();
```

FF/BoundaryConditions/coupledVelocity/coupledVelocityFvPatchField.C
```diff
@@ -95,7 +95,7 @@ void Foam::coupledVelocityFvPatchField::updateCoeffs()
     }
     const Foam::surfaceScalarField* phi = &db().lookupObject<surfaceScalarField>(phiName_);
     const scalarField& phip = phi->boundaryField()[this->patch().index()];
-    const vectorField n = this->patch().nf();
+    const vectorField n = this->patch().nf().cref();
```
vidulejs, line 98:
```C++
const Foam::tmp<Foam::vectorField> n = this->patch().nf();
```

FF/Velocity.C
```diff
@@ -92,7 +92,7 @@ std::size_t preciceAdapter::FF::Velocity::write(double* buffer, bool meshConnect
         if (fluxCorrection_)
         {
             scalarField phip = phi_->boundaryFieldRef()[patchID];
-            vectorField n = U_->boundaryField()[patchID].patch().nf();
+            const vectorField n = U_->boundaryField()[patchID].patch().nf().cref();
```
vidulejs, line 95:
```C++
const Foam::tmp<Foam::vectorField> n = U_->boundaryField()[patchID].patch().nf();
```

vidulejs (changes requested):
`cref()` returns a const reference (`const vectorField&`) to the field inside of the `tmp` wrapper. The `tmp` wrapper is essentially a smart pointer, which will deallocate the temporary field once they are no longer in use.

Since the `nf()` or "normal face" method returns a `tmp` field, why not just store it as a `tmp` field? This `tmp` field was apparently implicitly converted and the Intel compiler rightly complained about it. 

I think both, calling `cref()` on the `tmp` field or just storing it as a `tmp` field, is functionally the same in this case. But I prefer storing it as a `tmp` field, because I think it's a better programming practice. I compiled and tested this approach.

vidulejs (approved):

Issue #363, opened 2025-04-02
Ignoring user-defined solverType in the FSI module


MakisH:
At least in the FSI module, defining a custom solver type is ignored, at least when in a known type can be deduced.

To reproduce, add a custom solver type in the tutorial `quickstart/fluid-openfoam/system/preciceDict`:

```
FSI
{
  rho rho [1 -3 0 0 0 0 0] 1000;
  solverType basic;
}
```

I encountered this while checking the error handling in the constructor of `ForceBase`. The solver type is determined in `FSI/FSI.C` (functions `readConfig` **and** `configure`). Compare with `CHT/CHT.C`.

PR #364, opened 2025-04-02
Simplify error handling (remove error-deferred workaround)

MakisH:
Currently, we have a workaround to force OpenFOAM to stop when encountering an error in the configuration phase. This is not needed anymore since [OpenFOAM v2012](https://www.openfoam.com/news/main-news/openfoam-v20-12/post-processing#post-processing-function-object-error-handling), which introduces an additional option `errors` for the function object, which if set to `strict`, acts as we would expect in this scenario. See [related OpenFOAM issue](https://develop.openfoam.com/Development/openfoam/-/issues/1779#note_57407).

The `errors` option is now in OpenFOAM long enough that we can assume we will not trouble too many users.

This PR:

- Documents the option
- Removes the `error-deferred` option from `Utilities.C`
- Replaces all instances of `error-deferred` with `error`
- Replaces the `FataErrorInFunction` in `Utilities.C` with a `FatalError` (the `FatalErrorInFunction` did not seem to work in some cases, and it does not seem appropriate for a central logging function)
- Replaces a few explicit calls to `FatalErrorInFunction` with calls to the `adapterInfo` function, clarifying the respective error messages, if needed.

Ideally, I would go as far as to rename `adapterInfo` to `adapterLog`, but this would touch too many files and potentially frustrate some users that maintain their own forks.

I have tested the following:

- [x] Adapter function object does not exist
- [x] Adapter exists, preCICE does not
- [x] Typo in `patchName`
- [x] Typo in `readData`
- [x] Missing `rho` -> a default OpenFOAM error is raised during the lookup
- [x] Wrong solver type -> see #363 

Reviewers: Filter out the whitespace changes. I indented-out a large part of `Adapter.C` because I removed a try-catch construct. 

TODO list:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -357,6 +357,7 @@ functions
     preCICE_Adapter
     {
         type preciceAdapterFunctionObject;
+        errors strict; // optional
     }
 }
 ```
@@ -364,6 +365,8 @@ functions
 This directs the solver to use the `preciceAdapterFunctionObject` function object,
 which is part of the `libpreciceAdapterFunctionObject.so` shared library.
 The name `preCICE_Adapter` can be arbitrary. It is important that the library is loaded outside the `functions` dictionary when you want to use the custom boundary conditions that we provide with the FF module.
+
+The `errors strict` option is optional and [available since OpenFOAM v2012](https://www.openfoam.com/news/main-news/openfoam-v20-12/post-processing#post-processing-function-object-error-handling). Since the adapter is necessary to do a coupled simulation, this option instructs OpenFOAM to stop in case it faces issues with loading the adapter. For OpenFOAM versions that don't support this, remove the option.
 
 If you are using other function objects in your simulation, add the preCICE adapter to the end of the list. The adapter will then be executed last, which is important, as the adapter also controls the end of the simulation. When the end of the simulation is detected, the adapter also triggers the `end()` method of all function objects.
 
--- a/Adapter.C
+++ b/Adapter.C
@@ -230,20 +230,7 @@ bool preciceAdapter::Adapter::configFileRead()
 void preciceAdapter::Adapter::configure()
 {
     // Read the adapter's configuration file
-    if (!configFileRead())
-    {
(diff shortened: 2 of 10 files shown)


MakisH:
The system tests now pass: https://github.com/precice/openfoam-adapter/actions/runs/14307651426/job/40095066009?pr=364

Now only waiting for review.

docs/config.md
```diff
@@ -365,6 +366,8 @@ This directs the solver to use the `preciceAdapterFunctionObject` function objec
 which is part of the `libpreciceAdapterFunctionObject.so` shared library.
 The name `preCICE_Adapter` can be arbitrary. It is important that the library is loaded outside the `functions` dictionary when you want to use the custom boundary conditions that we provide with the FF module.
 
+The `errors strict` option is optional and [available since OpenFOAM v2012](https://www.openfoam.com/news/main-news/openfoam-v20-12/post-processing#post-processing-function-object-error-handling). Since the adapter is necessary to do a coupled simulation, this option instructs OpenFOAM to stop in case it faces issues with loading the adapter. For OpenFOAM versions that don't support this, remove the option.
```
davidscn, line 369: So, if not set, my OpenFOAM solver will just keep going, yes?

davidscn (approved):
Looks good, I didn't try to run it myself.

docs/config.md
MakisH: Correct, same as when OpenFOAM cannot find the adapter.
Do you think this is particularly risky?
Should we take any further actions beyond mentioning this prominently in the changelog?

vidulejs (approved):
Looks good! I noticed that the `errorsInConfigure` variable is not used anymore. Can be removed from Adapter.h and Adapter.c.

fsimonis (approved):

MakisH (commented):
Actually simplified the `configFileRead()` as well and tried a few cases (missing `preciceDict`, wrong module-field combination, no known module). All worked as expected.

PR #366, opened 2025-04-08
Handle potential exceptions from preCICE calls

### LINKED ISSUE #362: Not handling preCICE exceptions
Since preCICE v3.2.0, preCICE has more refined error handling via exceptions. However, these might give a very bad impression to the user, if not handled correctly, as the (often very clear) error message is accompanied by a stack trace that is not relevant for the user (and for this reason, I am adding the `bug` label here).

Example:

[wild-exception.txt](https://github.com/user-attachments/files/19532878/wild-exception.txt)

We should wrap the calls to the preCICE API with exception handlers. Example (by @fsimonis in the developers chat):

```c++
void myfunction() 
try {
  // boom code
} except (precice::Error& e) {
  // make error digestible for solver user
  std::exit(1);
}
```

This would be a backwards-compatible change.

MakisH:
Closes #362.

Every exception is just handled by throwing an error message, but this still leads to nicer output than the complete call stack. For example, simply:

```
---[precice] ERROR:  Sending data to another participant (using sockets) failed with a system error: write: Broken pipe. This often means that the other participant exited with an error (look there).
```

compared to previous state:

```
---[precice] ERROR:  Receiving data from another participant (using sockets) failed with a system error: read: End of file. This often means that the other participant exited with an error (look there).
terminate called after throwing an instance of 'precice::Error'
  what():  Receiving data from another participant (using sockets) failed with a system error: read: End of file. This often means that the other participant exited with an error (look there).
[tickly:49174] *** Process received signal ***
[tickly:49174] Signal: Aborted (6)
[tickly:49174] Signal code:  (-6)
[tickly:49174] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x42520)[0x71f76fc42520]
[tickly:49174] [ 1] /lib/x86_64-linux-gnu/libc.so.6(pthread_kill+0x12c)[0x71f76fc969fc]
[tickly:49174] [ 2] /lib/x86_64-linux-gnu/libc.so.6(raise+0x16)[0x71f76fc42476]
[tickly:49174] [ 3] /lib/x86_64-linux-gnu/libc.so.6(abort+0xd3)[0x71f76fc287f3]
[tickly:49174] [ 4] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xa2b9e)[0x71f7700a2b9e]
[tickly:49174] [ 5] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xae20c)[0x71f7700ae20c]
[tickly:49174] [ 6] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xae277)[0x71f7700ae277]
[tickly:49174] [ 7] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xae4d8)[0x71f7700ae4d8]
[tickly:49174] [ 8] /home/gc/repos/precice/precice/build/libprecice.so.3(+0x953ba)[0x71f76bc953ba]
[tickly:49174] [ 9] /home/gc/repos/precice/precice/build/libprecice.so.3(+0x7bf3e0)[0x71f76c3bf3e0]
[tickly:49174] [10] /home/gc/repos/precice/precice/build/libprecice.so.3(+0x53d408)[0x71f76c13d408]
[tickly:49174] [11] /home/gc/repos/precice/precice/build/libprecice.so.3(+0x5ac081)[0x71f76c1ac081]
[tickly:49174] [12] /home/gc/repos/precice/precice/build/libprecice.so.3(+0x5bbd0b)[0x71f76c1bbd0b]
[tickly:49174] [13] /home/gc/repos/precice/precice/build/libprecice.so.3(+0x5a5390)[0x71f76c1a5390]
[tickly:49174] [14] /home/gc/repos/precice/precice/build/libprecice.so.3(+0x697b0f)[0x71f76c297b0f]
[tickly:49174] [15] /home/gc/repos/precice/precice/build/libprecice.so.3(+0x67c9e9)[0x71f76c27c9e9]
[tickly:49174] [16] /home/gc/repos/precice/precice/build/libprecice.so.3(_ZN7precice11Participant7advanceEd+0x35)[0x71f76c265a2b]
[tickly:49174] [17] /home/gc/OpenFOAM/gc-v2412/platforms/linux64GccDPInt32Opt/lib/libpreciceAdapterFunctionObject.so(_ZN14preciceAdapter7Adapter7executeEv+0xa1)[0x71f76cd8e001]
[tickly:49174] [18] /home/gc/OpenFOAM/gc-v2412/platforms/linux64GccDPInt32Opt/lib/libpreciceAdapterFunctionObject.so(_ZN4Foam15functionObjects28preciceAdapterFunctionObject7executeEv+0x11)[0x71f76cdb5d61]
[tickly:49174] [19] /usr/lib/openfoam/openfoam2412/platforms/linux64GccDPInt32Opt/lib/libOpenFOAM.so(_ZN4Foam18functionObjectList7executeEb+0x100)[0x71f77081e0f0]
[tickly:49174] [20] /usr/lib/openfoam/openfoam2412/platforms/linux64GccDPInt32Opt/lib/libOpenFOAM.so(_ZNK4Foam4Time3runEv+0x1f1)[0x71f770833df1]
[tickly:49174] [21] pimpleFoam(+0x19d56)[0x5b694cf4ad56]
[tickly:49174] [22] /lib/x86_64-linux-gnu/libc.so.6(+0x29d90)[0x71f76fc29d90]
[tickly:49174] [23] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80)[0x71f76fc29e40]
[tickly:49174] [24] pimpleFoam(+0x21e35)[0x5b694cf52e35]
[tickly:49174] *** End of error message ***
Aborted (core dumped)
```

Not sure if there is any point in actually including the `.what()`. This should always be duplicated, right? Including it leads to this:

```
---[precice] ERROR:  Receiving data from another participant (using sockets) failed with a system error: read: End of file. This often means that the other participant exited with an error (look there).


--> FOAM FATAL ERROR: (openfoam-2412)
Error in the preCICE adapter: 
There was an issue while trying to advance preCICE:
Receiving data from another participant (using sockets) failed with a system error: read: End of file. This often means that the other participant exited with an error (look there).


    in file Utilities.C at line 32.

FOAM exiting
```

Also not sure if we want to actually trigger a `Foam::FatalError` via our logger. It does not really add much more information, besides that the error comes from the adapter.

@fsimonis can we assume that some API methods (such as the boolean checks) never throw? Would a `noexcept` keyword in these API methods make sense?

TODO list:

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -545,11 +545,33 @@ void preciceAdapter::Adapter::initialize()
     DEBUG(adapterInfo("Initializing the preCICE solver interface..."));
     SETUP_TIMER();
 
-    if (precice_->requiresInitialData())
+    bool requiresInitialData = false;
+    try
+    {
+        precice_->requiresInitialData();
+    }
+    catch (const std::runtime_error& e)
+    {
+        adapterInfo(std::string("An exception was raised while calling the preCICE method requiresInitialData:\n")
+                        + e.what(),
+                    "error");
+    }
+    if (requiresInitialData)
+    {
+        DEBUG(adapterInfo("Initializing preCICE data..."));
         writeCouplingData();
+    }
 
-    DEBUG(adapterInfo("Initializing preCICE data..."));
-    precice_->initialize();
+    try
+    {
+        precice_->initialize();
+    }
+    catch (const std::runtime_error& e)
+    {
+        adapterInfo(std::string("An exception was raised while calling the preCICE method initialize:\n")
+                        + e.what(),
+                    "error");
+    }
     preciceInitialized_ = true;
     ACCUMULATE_TIMER(timeInInitialize_);
 
@@ -566,7 +588,16 @@ void preciceAdapter::Adapter::finalize()
 
         // Finalize the preCICE solver interface
         SETUP_TIMER();
-        precice_->finalize();
+        try
+        {
+            precice_->finalize();
+        }
+        catch (const std::runtime_error& e)
+        {
+            adapterInfo(std::string("An exception was raised while calling the preCICE method finalize:\n")
+                            + e.what(),
+                        "error");
+        }
         ACCUMULATE_TIMER(timeInFinalize_);
 
         preciceInitialized_ = false;
@@ -587,7 +618,16 @@ void preciceAdapter::Adapter::advance()
     DEBUG(adapterInfo("Advancing preCICE..."));
 
     SETUP_TIMER();
-    precice_->advance(timestepSolver_);
+    try
+    {
+        precice_->advance(timestepSolver_);
+    }
+    catch (const std::runtime_error& e)
+    {
+        adapterInfo(std::string("An exception was raised while calling the preCICE method advance:\n")
+                        + e.what(),
+                    "error");
+    }
     ACCUMULATE_TIMER(timeInAdvance_);
 
     return;
@@ -652,7 +692,7 @@ void preciceAdapter::Adapter::adjustSolverTimeStepAndReadData()
        the same timestep as the one determined by preCICE.
     */
     double tolerance = 1e-14;
-    if (precice_->getMaxTimeStepSize() - timestepSolverDetermined > tolerance)
+    if (getMaxTimeStepSize() - timestepSolverDetermined > tolerance)
     {
         // Add a bool 'subCycling = true' which is checked in the storeMeshPoints() function.
         adapterInfo(
@@ -668,24 +708,24 @@ void preciceAdapter::Adapter::adjustSolverTimeStepAndReadData()
                 "warning");
         }
     }
-    else if (timestepSolverDetermined - precice_->getMaxTimeStepSize() > tolerance)
+    else if (timestepSolverDetermined - getMaxTimeStepSize() > tolerance)
     {
         // In the last time-step, we adjust to dt = 0, but we don't need to trigger the warning here
-        if (precice_->isCouplingOngoing())
+        if (isCouplingOngoing())
         {
             adapterInfo(
                 "The solver's timestep cannot be larger than the coupling timestep."
                 " Adjusting from "
-                    + std::to_string(timestepSolverDetermined) + " to " + std::to_string(precice_->getMaxTimeStepSize()),
+                    + std::to_string(timestepSolverDetermined) + " to " + std::to_string(getMaxTimeStepSize()),
                 "warning");
         }
-        timestepSolver_ = precice_->getMaxTimeStepSize();
+        timestepSolver_ = getMaxTimeStepSize();
     }
     else
     {
         DEBUG(adapterInfo("The solver's timestep is the same as the "
                           "coupling timestep."));
-        timestepSolver_ = precice_->getMaxTimeStepSize();
+        timestepSolver_ = getMaxTimeStepSize();
     }
 
     // Update the solver's timestep (but don't trigger the adjustDeltaT(),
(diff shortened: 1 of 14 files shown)


MakisH:
The system tests pass, the issues previously were related to the runner: https://github.com/precice/openfoam-adapter/actions/runs/14330310572/job/40172178566?pr=366

Adapter.C
```diff
@@ -545,11 +545,29 @@ void preciceAdapter::Adapter::initialize()
...
         writeCouplingData();
+    }
 
-    DEBUG(adapterInfo("Initializing preCICE data..."));
-    precice_->initialize();
+    try
+    {
+        precice_->initialize();
+    }
+    catch (const std::runtime_error& e)
```
davidscn: That's a rather verbose way of handling this, also a bit hard to read. Why not simply wrapping the whole code block into a `try catch`?

```cpp
try{
 if (precice_->requiresInitialData())
  precice_->initialize()
  } catch(...)
{}

}

File: Adapter.C [hunk 2]
```diff
@@ -587,7 +612,14 @@ void preciceAdapter::Adapter::advance()
...
 
     SETUP_TIMER();
-    precice_->advance(timestepSolver_);
+    try
+    {
+        precice_->advance(timestepSolver_);
+    }
+    catch (const std::runtime_error& e)
+    {
+        std::exit(1);
```
davidscn:
```suggestion
        std::exit(EXIT_FAILURE);
```

Adapter.C
```diff
@@ -702,6 +734,20 @@ void preciceAdapter::Adapter::adjustSolverTimeStepAndReadData()
     return;
 }
 
+double preciceAdapter::Adapter::getMaxTimeStepSize()
```
davidscn:
```suggestion
double preciceAdapter::Adapter::getMaxTimeStepSize() const
```
?

davidscn (commented):
> Not sure if there is any point in actually including the .what(). This should always be duplicated, right? Including it leads to this:

Actually no, I don't think so. It should only appear if we actually print (using `what`) in the `catch` block. It could be related to how we interface with boost log, I vaguely remember some issues we had with ASTE as well.

Regarding this issue. I am not 100% convinced of all the `try`'s which are now sprinkled across the entire code. A less scattered alternative would be to wrap highlevel functions in the `Adapter` class in a try-catch block.

We could still catch the precice exceptions using using `catch(::precice::Error)`. This would not work with former preCICE versions then, but we can use our version macros to distinguish and have something like

```cpp
#if PRECICE_VERSION_GTE(3,2,0)
using PreciceError = ::precice::Error
#else
using PreciceError = std::runtime_error
#endif
```

If the else branch becomes unnecessary at some point, we can remove it and have the clean solution.

Adapter.C
fsimonis: Or the whole function, as we do in the C bindings.

https://github.com/precice/precice/blob/3c34660420b3fcd1595d50158dac6b207c1c7322/extras/bindings/c/src/preciceC.cpp#L68-L74

MakisH:
Thank you both for the very useful comments. I applied all previous feedback, and the changes now are much fewer.

Adapter.C
davidscn, line 1: Why not the `configure` function? Should also capture the `Interface` functions then, no? What about the `teardown`, where we potentially destroy preCICE?

MakisH: Out of relevant functions, the `configure()` only calls `initialize()`, which already handles preCICE exceptions. The `finalize()` also calls `teardown()`, however it is `finalize()` that handles exceptions.

I now wrapped `configure()` and `teardown()` as well.

I generally tried to handle exceptions in all functions that interact with the `precice_` object.

The `Interface` has some functions wrapped already.

Anything else I am missing?

davidscn: > Out of relevant functions, the configure() only calls initialize(), which already handles preCICE exceptions. The finalize() also calls teardown(), however it is finalize() that handles exceptions.

Ah ok I didnt check that.

I think we dont need to double wrap: if a function down the stack throws, we catch it further upstream (stack unwinding). So, my question was more: do we need the catches in the `Interface`?

MakisH: Good point. I went one step further and removed the handling from every function that is not directly called by the `preciceAdapterFunctionObject` class, which is the entry point. We now only have:

- `configure()`
- `execute()`
- `adjustTimeStep()`
- `end()`
- `~Adapter()`

plus two minor unrelated fixes that happened during the process.

davidscn (approved):
Perfetto

PR #369, opened 2025-07-10
Mesh points & volumes checkpointing

vidulejs:
## Overview
- Removed comments and vectors to mesh fields which don't exist
- Fixed `meshOldPoints` checkpointing. Previously mesh would move from incorrect interim state to `meshPoints`, instead of `meshOldPoints` (leads to incorrect mesh flux).
- Mesh volume checkpointing (needed for higher order time derivative schemes)

#### Cleanup and refactor
This relates to the bulk of line removals in this branch.

Just like checkpointed fields of physical variables, there were different types of checkpointed mesh fields, however they were unnecessary and empty. There were no `meshVolVectorFields_` or `meshSurfaceVectorFields_` (none that needed checkpointing) . The only checkpointed mesh field is face motion flux `mesh_.phi()` of type `surfaceScalarField`. As pointed to by in a comment in `setupMeshCheckpointing()`:
> // The other mesh <type>Fields:
> //      C
> //      Cf
> //      Sf
> //      magSf
> //      delta
> // are updated by the function fvMesh::movePoints. Only the meshPhi needs checkpointing.

#### Mesh Points Checkpointing
- Store `meshPoints_` and `meshOldPoints_` as pointers instead of objects for consistency with the rest of the codebase. These are actually pointers to the copies managed by the adapter.
- ~~The main logic of restoring mesh checkpoints is like this:~~
<del>
<pre><code>
&gt; // const_cast&lt;pointField&amp;&gt;(mesh_.points()) = oldMeshPoints_;
&gt; 
&gt; // Reload mesh points
&gt; const_cast<Foam::fvMesh&>(mesh_).movePoints(*meshPoints_)
</code></pre>
</del>

~~Currently the first line is commented out and probably forgotten by the previous developer Derek (and I do not understand the meaning of this comment).~~
~~> // Switch oldpoints on for pure physics. (is this required?). Switch off for better mesh deformation capabilities?~~
~~> // const_cast<pointField&>(mesh_.points()) = oldMeshPoints_;~~

~~I had independently come to the conclusion that this needs to be enabled, when looking at the `fvMesh.movePoints()` method. Perhaps we need a longer discussion on this.~~

**EDIT***
- The main logic of restoring mesh checkpoints is like this:
```cpp
    // Reload mesh points
    const_cast<Foam::fvMesh&>(mesh_).movePoints(*meshPoints_);

    // polyMesh.movePoints will only update oldPoints
    // if (curMotionTimeIndex_ != time().timeIndex())
    // I.e., first implicit iteration
    const_cast<pointField&>(mesh_.oldPoints()) = *meshOldPoints_;
 ```
 
`mesh_.oldPoints()` will not get updated inside of `fvMesh.movePoints()` -> `polyMesh.movePoints()` in implicit iterations. See the discussion in the comments. Related issue - FSI: Subcycling with implicit coupling #58. 

#### Mesh Volumes Checkpointing

For solvers that use higher-order time derivatives (e.g., Euler ddtScheme), checkpoint cell volumes (`V, V0, V00`).
- Implement new methods `setupMeshVolCheckpointing,
  readMeshVolCheckpoint, writeMeshVolCheckpoint`
- A counter `volumeCheckpointCounter` is used to get `V0, V00`, when they become available. Otherwise `mesh_.V00()` access will throw an error.
- Only store copies of the volume fields (`meshVolFieldCopies_`). The references given by `mesh_.V0()` may be invalidated by OpenFOAM, so we cannot store pointers to them. Instead directly call `mesh_.V0()` and so on.
- `readMeshVolCheckpoint()` will const_cast the const reference and overwrite the memory, e.g.,
```cpp
const_cast<volScalarField::Internal&>(mesh_.V()) = *(meshVolFieldCopies_.at(0));
```

TODO list:


Adapter.H
```diff
@@ -138,45 +138,27 @@ private:
...
 
     // Vectors of pointers to the checkpointed mesh fields and their copies
 
-    //- Checkpointed surfaceScalarField mesh fields
+    //- Checkpointed mesh points, oldPoints
+    Foam::pointField* meshPoints_;
+    Foam::pointField* meshOldPoints_;
+
+    //- Checkpointed volScalarField mesh fields (V, V0, V00)
+    std::vector<volScalarField::Internal*> meshVolFieldCopies_;
```
MakisH, line 173: Please remind me, it's been a while: What is the reason that here (and in `addMeshCheckpointField()`) we only need to store the copies, but for `surfaceScalarFields` we needed two objects (main and copy)?

Can we get rid of one of the two in the other fields as well?

Adapter.C
```diff
@@ -754,23 +752,15 @@ void preciceAdapter::Adapter::reloadCheckpointTime()
...
-    // Add a bool 'subcycling' which can be evaluated every timestep.
-    if ( !oldVolsStored && mesh_.foundObject<volScalarField::Internal>("V00") ) // For Ddt schemes which use one previous timestep
+    if (!meshPoints_)
     {
-        setupMeshVolCheckpointing();
-        oldVolsStored = true;
+        DEBUG(adapterInfo("Storing mesh points..."));
+        // Add points and oldPoints
+        meshPoints_ = new Foam::pointField(mesh_.points());
+        // First timestep oldPoints doesn't exist yet
```
MakisH:
```suggestion
        // First timestep, oldPoints doesn't exist yet
```

But I don't understand: if it is an issue that it does not yet exist, how to copy that?

Adapter.C
```diff
@@ -780,7 +770,14 @@ void preciceAdapter::Adapter::storeMeshPoints()
             meshCheckPointed = true;
         }
         writeMeshCheckpoint();
-        writeVolCheckpoint(); // Does not write anything unless subcycling.
+
+        // For Ddt schemes which use up to two previous timesteps V0, V00
+        if (volumeCheckpointCounter < 3)
```
MakisH: @vidulejs you probably also want to change this one to some kind of more meaningful flag. Unless you think that a counter makes more sense here.

How does the code behave for >=3?

Adapter.C
```diff
@@ -792,100 +789,60 @@ void preciceAdapter::Adapter::reloadMeshPoints()
         return;
     }
 
-    // In Foam::polyMesh::movePoints.
-    // TODO: The function movePoints overwrites the pointer to the old mesh.
-    // Therefore, if you revert the mesh, the oldpointer will be set to the points, which are the new values.
-    DEBUG(adapterInfo("Moving mesh points to their previous locations..."));
+    // Invalidate on-demand fields
```
MakisH:
```suggestion
    // Invalidate derived fields, which should be calculated by movePoints()
```

Adapter.C
```diff
@@ -792,100 +789,60 @@ void preciceAdapter::Adapter::reloadMeshPoints()
...
+    // fvMesh::updateGeomNotOldVol()
```
MakisH:
```suggestion
    // See a similar implementation in fvMesh::updateGeomNotOldVol():
```

Adapter.C
```diff
@@ -792,100 +789,60 @@ void preciceAdapter::Adapter::reloadMeshPoints()
...
+    (void)mesh_.Sf();
+    (void)mesh_.magSf();
+    (void)mesh_.C();
+    (void)mesh_.Cf();
 
-    // TODO
-    // Switch oldpoints on for pure physics. (is this required?). Switch off for better mesh deformation capabilities?
-    // const_cast<pointField&>(mesh_.points()) = oldMeshPoints_;
-    const_cast<fvMesh&>(mesh_).movePoints(meshPoints_);
+    // fvMesh.movePoints overwrites the pointer to the oldPoints
```
MakisH: Expand this comment:
- this looks counter-intuitive (assigning old to current)
- but later on we will call movePoints

Adapter.C
```diff
@@ -792,100 +789,60 @@ void preciceAdapter::Adapter::reloadMeshPoints()
...
-    /*  // TODO This part should only be used when sybcycling. See the description in 'storeMeshPoints()'
-        // The if statement can be removed in this case, but it is still included for clarity
-    if ( oldVolsStored )
-    {
-        readVolCheckpoint();
-    }
-    */
+    DEBUG(adapterInfo("Moved mesh points to their previous locations."));
+
+    readMeshVolCheckpoint();
```
MakisH, line 876: We need a similar comment for the volumes, or to move the message after both operations and make it for both.

Adapter.C
```diff
@@ -792,100 +789,60 @@ void preciceAdapter::Adapter::reloadMeshPoints()
...
-        const_cast<volScalarField::Internal&>
-        (
-            mesh_.V00()
-        )
-    );*/
-#ifdef ADAPTER_DEBUG_MODE
-    adapterInfo(
-        "Added " + mesh_.V00().name() + " in the list of buffer checkpointed fields.");
-#endif
+    // Add V, V0, V00
```
MakisH:
```suggestion
    // Add the mesh volumes for the current and up to two previous time steps (V, V0, V00), if available
```

MakisH (commented):
Looks good overall, thank you for all the cleanup effort! This is an important step.

Please also:

- [ ] Add a changelog entry
- [ ] Update the PR description with the motivation and linking to related issues/PRs.

I am triggering the system tests mainly as a routine check, but we will (indeed) need to update the reference results after merging.

Adapter.C
```diff
@@ -792,100 +789,60 @@ void preciceAdapter::Adapter::reloadMeshPoints()
...
-    // TODO: The function movePoints overwrites the pointer to the old mesh.
-    // Therefore, if you revert the mesh, the oldpointer will be set to the points, which are the new values.
-    DEBUG(adapterInfo("Moving mesh points to their previous locations..."));
+    // Invalidate on-demand fields
+    // fvMesh::updateGeomNotOldVol()
+    // (void)mesh_.V();
+    (void)mesh_.Sf();
+    (void)mesh_.magSf();
+    (void)mesh_.C();
+    (void)mesh_.Cf();
```
MakisH: I am still not convinced that this is not just an implicit casting of the return value to `(void)`, which seems to be a way to discard the return value:

https://stackoverflow.com/questions/689677/why-cast-unused-return-values-to-void

I am also not yet convinced that invalidating the pointer would add some safety, if the method works correctly.

Please have a deeper look here.

vidulejs:
Thanks for the discussion @MakisH! I have now had a look with fresh eyes, it's been a while since I made the original commits. 

There's something I need to bring attention to. Although this PR is not about implicit subcycling (WIP PR #370), the two are closely related. Bear with me for a moment as I try to explain.

The old issue on *FSI: Subcycling with implicit coupling #58* mentions the private member variable `curTimeIndex_` of `fvMesh`, which will prevent the  `storeOldVol()` method from executing in subsequent implicit iterations. 
```cpp
 Foam::tmp<Foam::scalarField> Foam::fvMesh::movePoints(const pointField& p)
 {
  // Grab old time volumes if the time has been incremented
  // This will update V0, V00
  if (curTimeIndex_ < time().timeIndex())
  {
      storeOldVol(V());
  }
  // (...)
}
```
When we restore a checkpoint, we restore the `Time` object, which is in the past and less than `curTImeIndex_`. The `fvMesh` class is responsible for the cell volumes `V`, `V0`, `V00`. 
Similarly, the `polyMesh` class is responsible for `points` and `oldPoints`, and it also performs such a check:
```cpp
void Foam::polyMesh::movePoints(const pointField& newPoints)
{
    // (...)

    // Pick up old points
    if (curMotionTimeIndex_ != time().timeIndex())
    {

        // Mesh motion in the new time step
        oldPointsPtr_.reset(nullptr);
        oldPointsPtr_.reset(new pointField(points_));
        curMotionTimeIndex_ = time().timeIndex();
    }

    points_ = newPoints;

    // (...)
}
```
**Now, what does this mean?** In implicit coupling w/o subcycling this is inconsequential (!) We perform a solver step spanning the whole window. In the first coupling iteration `curTimeIndex_` is updated and we store the old volumes and old points corresponding to the start of the window. The new points are updated irregardless. In subsequent implicit coupling iterations the state of old volumes and old points is correct, i.e., corresponds to the start of the window. Also mesh face flux `meshPhi` is correct.

**However, for implicit coupling w/ subcycling**, `polyMesh::movePoints()` will store the old points from the last (unconverged/interim) subcycling step . 

We perform multiple subcycling steps within the window, the old points will always update since `curMotionTimeIndex_ != time().timeIndex()`. 

On the next implicit coupling iteration, the old points will be from the last subcycling step - not from the start of the window! The mesh face motion flux `meshPhi` will be incorrect and introduce spurious motion!

### Solution

Therefore, instead of relying on `fvMesh.movePoints()` to store the old points (sometimes correctly),  we have to make sure they are always updated correctly like this.
```diff
void preciceAdapter::Adapter::reloadMeshPoints()
{
    if (!mesh_.moving())
    {
        DEBUG(adapterInfo("Mesh points not moved as the mesh is not moving"));
        return;
    }

-    // fvMesh.movePoints overwrites the pointer to the oldPoints
-    const_cast<pointField&>(mesh_.points()) = *meshOldPoints_;

     // Reload mesh points
     const_cast<Foam::fvMesh&>(mesh_).movePoints(*meshPoints_);

+    // polyMesh.movePoints will only update oldPoints
+    // if (curMotionTimeIndex_ != time().timeIndex())
+    const_cast<pointField&>(mesh_.oldPoints()) = *meshOldPoints_;

     readMeshCheckpoint();
     
    DEBUG(adapterInfo("Moved mesh points to their previous locations."));

    readMeshVolCheckpoint();
}
```

Adapter.H
vidulejs: We do not need to explicitly store pointers to the `fvMesh.points()` and `fvMesh.oldPoints()` because we can just call these methods when we need to access them. The logic for these fields is handled specifically as opposed to generically. Same goes for the mesh volumes.

We store pointers to the OpenFOAM field objects and our copies for convenience. In `setupCheckpointing()` we loop over the OpenFOAM object registry to get all the fields, then in `addMeshCheckpointField()` we store the pointers in `*Fields_` and `*FieldCopies_` vectors for each type of field. 

The mesh face motion flux `meshPhi` is not indexed in the object registry, therefore it is added manually by calling it's method. 
```cpp
void preciceAdapter::Adapter::setupMeshCheckpointing()
{
    DEBUG(adapterInfo("Creating a list of the mesh checkpointed fields..."));
    // Add meshPhi (Face motion flux)
    addMeshCheckpointField(const_cast<surfaceScalarField&>(mesh_.phi()));

    DEBUG(adapterInfo("Added " + mesh_.phi().name() + " to the list of checkpointed fields."));
}
```
Afterwards it is accessed from `meshSurfaceScalarFields_` and `meshSurfaceScalarFieldCopies_`, however these vectors contain only `meshPhi` and no other `surfaceScalarFields`.

Perhaps we can refactor these parts for consistency with the rest of the codebase.

Adapter.C
vidulejs: The debug statements for mesh volumes are inside of `readMeshVolCheckpoint()`

Adapter.C
vidulejs: Yes, you are correct. We are just calling the methods and discarding the return value. I took this code block from ``fvMesh.updateGeomNotOldVol()``. There is a comment in the OpenFOAM source code on why they do this, but this code will execute unconditionally anyway when `fvMesh.movePoints()` is called. Anyway, we can safely remove these lines of code.

```cpp
void Foam::fvMesh::movePoints(const pointField& p)
{
	// (...)
    polyMesh::movePoints(p);

    // Update or delete the local geometric properties as early as possible so
    // they can be used if necessary. These get recreated here instead of
    // demand driven since they might do parallel transfers which can conflict
    // with when they're actually being used.
    // Note that between above "polyMesh::movePoints(p)" and here nothing
    // should use the local geometric properties.
    updateGeomNotOldVol();
    
    // (...)
```

As for deleting the on-demand fields, it is actually done elswhere. 

```cpp
void Foam::polyMesh::movePoints(const pointField& newPoints)
{
    // (...)

    // Update the mesh geometry (via fvGeometryScheme)
    // - updateGeom is virtual -> calls fvMesh::updateGeom (or higher)
    // - fvMesh::updateGeom defers to surfaceInterpolation::updateGeom(),
    //   which defers to fvGeometryScheme::movePoints()
    // - set the mesh flux
    // - clear out/recalculate stale geometry
    updateGeom();
    
    // (...)
 ```
 
 -->
 
 ```cpp
void Foam::primitiveMesh::clearGeom()
{
    if (debug)
    {
        Pout<< "primitiveMesh::clearGeom() : "
            << "clearing geometric data"
            << endl;
    }

    deleteDemandDrivenData(cellCentresPtr_);
    deleteDemandDrivenData(cellVolumesPtr_);
    deleteDemandDrivenData(faceCentresPtr_);
    deleteDemandDrivenData(faceAreasPtr_);
}
 ```

**Comment by vidulejs** [comment]
In the validation of this branch I found out something interesting. Namely, higher order timestepping with volume checkpoints already works on the develop branch. When the volume checkpointing is properly implemented in the adapter, there is no numerical difference between this and the develop branch. 

The reason is that this same if condition `if (curTimeIndex_ < time().timeIndex())` I mentioned before in `polyMesh.movePoints` is also present in `fvMesh.movePoints`, which prevents writing `V0` and `V00` when repeating the same timestep in implicit iterations. Therefore, the old mesh volumes are already checkpointed correctly by OpenFOAM itself. OpenFOAM will store the correct old mesh volumes corresponding to the start of the window. 

```cpp
void Foam::fvMesh::movePoints(const pointField& p)
{
    DebugInFunction << endl;

    // Grab old time volumes if the time has been incremented
    // This will update V0, V00
    if (curTimeIndex_ < time().timeIndex())
    {
        storeOldVol(V());
    }
\\ ...
```

The only difference will arise in implicit subcycling, when taking multiple steps within on time window and the old mesh volumes will correspond to the last substep of the first implicit iteration. That will cause an incorrect state when reloading to the start of the window. Work on subcycling is currently in progress and it would make more sense to get back to this once more pieces are in place.

Therefore, I do not think this PR should be merged yet. I made it a draft PR for now. The rest of the changes regarding the refactor can be merged, I made a separate PR for those in #387 .

PS: Now it makes more sense, why the original issue on volume checkpointing #58 talks about subcycling.

vidulejs:
I need to merge the latest version of develop into this branch and add a comment about subcycling, then this would be ready to merge.

MakisH:
> I need to merge the latest version of develop into this branch and add a comment about subcycling, then this would be ready to merge.

@vidulejs it would be great to include it in the next release, if you could do that over the next week. If not, or if you need a bit more time, please let me know.

Note that there are a few unresolved conversations, and this is currently marked as draft.

Adapter.C
vidulejs: Changed `volumeCheckpointCounter` to an enum `volumeCheckpointState_`.

Adapter.H
MakisH: I would say "if it's working now, let's not touch it". Let's get this one merged for now.

Adapter.H
```diff
@@ -347,6 +367,13 @@ private:
     //- Write the checkpoint - store the fields and time
     void writeCheckpoint();
 
+    // TODO Probably these can be included to the mesh checkpoints.
```
MakisH: Fine for now. Let's not introduce another TODO.


```suggestion
```

Adapter.C
```diff
@@ -1409,6 +1441,51 @@ void preciceAdapter::Adapter::writeMeshCheckpoint()
...
 }
 
+void preciceAdapter::Adapter::readMeshVolCheckpoint()
+{
+    // Reload V, V0, V00
+    if (volumeCheckpointState_ == VolumeCheckpointState::V_AND_V0_AVAILABLE || volumeCheckpointState_ == VolumeCheckpointState::ALL_AVAILABLE)
+    {
+        // Actually we don't need to reload V, as it is re-calculated in every time-step
+        // const_cast<volScalarField::Internal&>(mesh_.V()) = *(meshVolFieldCopies_.at(0));
+        // DEBUG(adapterInfo("Read mesh volume " + meshVolFieldCopies_.at(0)->name()));
```
MakisH: No reason to keep the dead code, it's clear how to add it if one would need it:

```suggestion
        // We do not need to reload V, as it is re-calculated in every time-step
```

Adapter.C
```diff
@@ -1498,6 +1575,15 @@ void preciceAdapter::Adapter::teardown()
...
 
+        for (uint i = 0; i < meshVolFieldCopies_.size(); i++)
+        {
+            delete meshVolFieldCopies_.at(i);
+        }
+        meshVolFieldCopies_.clear();
+
+        delete meshPoints_;
+        delete meshOldPoints_;
+
```
MakisH, line 1584: These have already been deleted above, leading to `double free or corruption error`.


```suggestion
```

MakisH (commented):
Thanks for the follow-up! I went through everything again, and I think this is ready to merge. Just a few comment suggestions, which I am applying directly.

We do not have any affected tutorials to verify that this now works better with higher-order schemes, but at least it does not seem to break any existing tutorials.

MakisH:
Triggered the system tests, which are now much more extensive: https://github.com/precice/openfoam-adapter/actions/runs/27304886579/job/80659972966?pr=369

I expect some of the FSI cases to fail, maybe not all of them.

MakisH (approved):

Issue #368, opened 2025-06-13
Modeling aspects of FSI with implicit coupling and subcycling


vidulejs:
While working on issue #58, ~~I discovered the real reason for instability in FSI simulations with subcycling. Initially it ticked me off, that the mesh was changing during subcycling steps. Maybe this was intended but I was not sure.~~

~~However, I realized that at each subcycling step the displacement is read again. This is the total displacement and not multiplied by subcycling dt. This means that for N subcycling steps, the total displacement will be amplified N times.~~

EDIT: What I thought before was wrong and not the cause of the issue. The core of the problem is that the implicit subcycling simulations converge to different solutions than the non-subcycling ones. See my new comment below.

davidscn:
> However, I realized that at each subcycling step the displacement is read again

For FSI and subcycling, this would be the expected behavior.

> This is the total displacement and not multiplied by subcycling dt.

That's the expected behavior either. preCICE takes care of the time interpolation between the beginning and the end of the time window, so each intermediate state will just be in between those two displacement states. The read displacement is total, yes, but the state of the fluid grid within a simulation is already deformed, so there is only an incremental portion left to reach this absolute state. I think this might go wrong when using relative displacements. But from your description, I understand you are dealing with absolute displacements, right?

> This is the total displacement and not multiplied by subcycling dt

Why should the displacement be multiplied by dt? What would be the unit of this quantity?

> This means that for N subcycling steps, the total displacement will be amplified N times.

I dont understand this and disagree, could explain this more in detail?

vidulejs:
Excuse the pasted text, I'm in a slight rush. I was in the process of doing a convergence study on the perpendicular flap with subcycling when I wrote most of this text.  

The idea was finding a timestep, time-window and subcycling timestep that converges to the same solution. The natural assumption is that as the timestep goes down, the solutions will converge. What I found instead was that there are two solutions that the simulations converge to;  (1) the non-implicit-subcycling cases and (2) the implicit-subcycling cases. See the plots below. 

Note that explicit subcycling converges to the first group IIRC, so the problem is strictly with implicit subcycling. However I cannot conclude whether the problem is on the OpenFOAM adapter side or somewhere else (the solid participant is not the problem though, same issue with Calculix and Deal II solid). 

# Timestep convergence analysis

[...]

PR #370, opened 2025-07-10
FSI implicit subcycling fix

vidulejs:
...

TODO list:

… at correct time within subcycling window. Not fully tested yet.


MakisH:
In your tests, keep in mind that the CalculiX adapter also has a similar issue: https://github.com/precice/calculix-adapter/issues/9

vidulejs:
This pull request was intended to fix issue #368, but it turns out the discussion there extends beyond the OpenFOAM adapter. Nothing to do on the adapter side.

Issue #295, opened 2023-07-18
Generic readers/writers with additional operations


MakisH:
In the context of #279 and similar issues, and looking at the current state of the adapter, I think that we could move away from the field-specific readers/writers to generic readers/writers very easily for most of the fields, as we are just extracting and injecting values at specific locations.

For the readers/writers that need to do additional conversions, we could potentially use something similar to the coded function objects that OpenFOAM provides. Either the user could provide details on what conversions to do, or (non-exclusive) we could apply the additional operations only on the fields that require this.

Merely documenting the idea for now, I don't yet have clear implementation directions.

MakisH:
A generic module has been introduced in https://github.com/precice/openfoam-adapter/pull/346, with follow-up extensions in https://github.com/precice/openfoam-adapter/pull/398 and https://github.com/precice/openfoam-adapter/pull/400. Find the documentation on https://precice.org/adapter-openfoam-config.html#generic-module

This is meant to become the default in a future v2 of the adapter, after gradually transitioning the tutorials to that.

The `with additional operations` part (meaning user-defined operations) is still open.

Issue #279, opened 2023-02-15
Generic writer/reader for volume coupling


MakisH:
In the context of #255, we discussed whether a generic reader/writer would make sense (@tirgendetwas). An additional use case (as discussed with Sergio Roget at the preCICE Workshop 2023) would be coupling forces over a porous medium. These forces should already be available in the objects registry, so a generic reader/writer should be able to access it.

To make it very generic, we could use a template for any (or even constrained) type.

Later on, we could also add a generic reader/writer for surface fields.

Potentially also related to #277 and related to #77, #150.

sroget:
Hello.
I comment my "progress" and my problems. I have to say that I am a newbie in this field and I don't understand perfectly what the code does.
What I want to do is to extract the drag and lift forces from a porous medium and apply them to a structural model. So this is a FSI case as I see it.

After talking with @MakisH  I have tried to replicate the structure of the CHT/Temperature.C file in the FSI/Force.C file, since the chanel-transport tutorial is the closest to what I want to do.

It compiles correctly so from a machine point of view it seems to be fine. I have created a simple case from the perpendicular-flap tutorial for testing. Where I create a openFOAM case with a porous medium in the middle of the domain. Also I modified the structural case made also in openFOAM to set the flap in the same position of the porous medium.
 It seems that there is communication between fluid-openfoam and solid-openfoam. When I open the cases with paraview I get the force field. The problem is that the values of that field are practically zero. I guess I am not reading the forces produced by the porous medium. I know that openfoam calculates them with the functionObject force so I understand that somehow it could be used with preCICE. 

Could you give me a path to follow and go forward in this project?

Thank you in advance

MakisH:
>  I have tried to replicate the structure of the CHT/Temperature.C file in the FSI/Force.C file, since the chanel-transport tutorial is the closest to what I want to do.

@sroget could you please link to your code? You can fork this repository and push to a branch. I am not sure I understand what you have done.

Which branches of the adapter are you based upon? We have not yet merged the volume coupling related changes (see https://github.com/precice/openfoam-adapter/pull/255 and https://github.com/precice/openfoam-adapter/pull/270, as well as https://github.com/precice/tutorials/pull/315).

> When I open the cases with paraview I get the force field. The problem is that the values of that field are practically zero. I guess I am not reading the forces produced by the porous medium. I know that openfoam calculates them with the functionObject force so I understand that somehow it could be used with preCICE.

[...]

Issue #138, opened 2020-08-28
Support an general data exchange (read/write) for field datas


TEFEdotCC:
Issue to support incompressible multiphase: [#127](https://github.com/precice/openfoam-adapter/issues/127)
Extracted from my actually problem: [precice.discourse.group](https://precice.discourse.group/t/need-help-with-icoreactingmul-solver/307)

**preCICE version:** 2.1.0
**openfoam-adapter git commit:** [c49b862](https://github.com/precice/openfoam-adapter/commit/c49b862aab4845ae79b8b7257a20e15ba5a0019c)
**OpenFOAM version:** v2006

Currently I'm not so deep in preCICE and OpenFOAM code base, but my basic thinking about the job of preCICE is to communicate with different solvers and transfer data between them.
The openfoam-adapter do also calculation steps, is interessted to turbulence models, ... and so on. But why? I think it is not the job of preCICE and there adapters. Maybe to implement missing thinks, ok. But for generall purpose? I would say no as long as the solver also has possibilites to setup and calculate thinks.

For my problem OpenFOAM can calculate the wallHeatFlux out of the box.
[config file](https://develop.openfoam.com/Development/openfoam/-/blob/master/etc/caseDicts/postProcessing/fields/wallHeatFlux), [header](https://develop.openfoam.com/Development/openfoam/-/blob/master/src/functionObjects/field/wallHeatFlux/wallHeatFlux.H) and [source](https://develop.openfoam.com/Development/openfoam/-/blob/master/src/functionObjects/field/wallHeatFlux/wallHeatFlux.C)
So, the code should do the job on all offical shiped solvers with heat transfer support.
If I want to use this OpenFOAM feature, I need an general data field reader / writer. That could be also nice to transfer thinks from own coeded field data in openfoam. e.g passiv scalar field.

The implementation of an general data reader could be easy like the [CHT/Temperature](https://github.com/precice/openfoam-adapter/blob/master/CHT/Temperature.C) code. In preciceDict should also be possible to set up an mapping between openfoam fields and precice-data.

**Expand precice-config.xml** to introduce an type entry to data (--> clean code)

    <data:scalar name="myShitNameForNothing" type="none|Temperature|Heat Flux|..." />  

**New module general** in preCICEDict? Like

    modules (GENERAL);
    
    GENERAL
    {
        mapping
        (
            ("data:name", "field name:OpenFOAM")
        );
    }

EDIT:
Or do all the shit in preCICEDict

[...]

PR #380, opened 2025-10-21
Add coupling fields used for CFD-DEM

davidscn:
Adds 

- the drag force
- explicit and implicit momentum fields
- writing the pressure gradient

### DIFF
--- a/FF/PressureGradient.C
+++ b/FF/PressureGradient.C
@@ -1,0 +1,72 @@
+#include "PressureGradient.H"
+
+using namespace Foam;
+
+preciceAdapter::FF::PressureGradient::PressureGradient(
+    const Foam::fvMesh& mesh,
+    const std::string nameP)
+: p_(
+    const_cast<volScalarField*>(
+        &mesh.lookupObject<volScalarField>(nameP)))
+{
+    dataType_ = scalar;
+}
+
+std::size_t preciceAdapter::FF::PressureGradient::write(double* buffer, bool meshConnectivity, const unsigned int dim)
+{
+    int bufferIndex = 0;
+
+    // For every boundary patch of the interface
+    for (uint j = 0; j < patchIDs_.size(); j++)
+    {
+        int patchID = patchIDs_.at(j);
+
+        // Get the pressure gradient boundary patch
+        const scalarField gradientPatch((p_->boundaryFieldRef()[patchID])
+                                            .snGrad());
+
+        // For every cell of the patch
+        forAll(gradientPatch, i)
+        {
+            // Copy the pressure gradient into the buffer
+            buffer[bufferIndex++] =
+                -gradientPatch[i];
+        }
+    }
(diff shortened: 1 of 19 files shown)


FF/PressureGradient.C
```diff
@@ -6,31 +6,91 @@ preciceAdapter::FF::PressureGradient::PressureGradient(
...
+          &mesh.lookupObject<volScalarField>(nameP))),
+  gradP_(IOobject(
+             "gradP",
+             mesh.time().timeName(),
+             mesh,
+             IOobject::NO_READ,
+             IOobject::NO_WRITE),
+         fvc::grad(*p_))
 {
-    dataType_ = scalar;
```
davidscn, line 12: This change is critical, but necessary. In previous versions, gradP was a scalar for some reason. Maybe we should introduce a normal gradient separately.

MakisH: According to our [versioning scheme](https://precice.org/adapter-openfoam-get.html#what-does-the-adapter-version-mean), this would be a breaking change / major version jump. But I also see the point that this is wrongly defined (or with a very restrictive assumption).

I also stumbled upon this in another context, and we should indeed make a separate normal gradient. We also have other gradients that need reconsideration or renaming (`AlphaGradient`, `TemperatureGradient`, `VelocityGradient`).

I would suggest that, as a first step, we rename this to `PressureGradientVector` (or a better name), so that it can be merged as-is and released in a v1.4.0, and then we revamp the fields in a v2 of the adapter.

MakisH (commented):
Many nice new features, thanks! Remember to also update the documentation.

I have not looked into details yet. Is there any test case already available?

Note that one of the major updates we were planning with @vidulejs for v2 was the generic readers/writers (#138, #279, #295, #346), and this PR gives us more examples to consider.

MakisH:
Since all the additions here are rather straight-forward, this would be a perfect use case for https://github.com/precice/openfoam-adapter/issues/295

@vidulejs has prepared a first draft in https://github.com/precice/openfoam-adapter/pull/346

and we will need to then distinguish between `gradient` and `normal-gradient`, for which we will need https://github.com/precice/openfoam-adapter/pull/354

We will get both in v1.4.0. Before that, there will at least be a v1.3.2 next week.

davidscn:
I don't really see the reason to not break the current state and do a major release soon, but ok, now it is backwards compatible. I named the field now

`FullPressureGradient`

Ready to be merged.

MakisH:
Thanks for trying to make this non-breaking!
Suggestion for the name: `PressureGradientFull` would better fit the current naming scheme (see, e.g., `DisplacementDelta`).

MakisH:
The new fields also need to be added to the documentation:

- https://github.com/precice/openfoam-adapter/blob/develop/docs/README.md
- https://github.com/precice/openfoam-adapter/blob/develop/docs/config.md

Issue #382, opened 2025-11-14
Remeshing support


MakisH:
As of v3.2.0, preCICE supports [dynamic meshes](https://precice.org/couple-your-code-moving-or-changing-meshes.html) with remeshing. This also needs the adapter to support it.

The support should be relatively easy, and we could add it soon.

We would need a simple test case, ideally based on one of our tutorials. OpenFOAM seems to offer the [dynamicRefineFvMesh](https://api.openfoam.com/2506/classFoam_1_1dynamicRefineFvMesh.html) class, I am not sure how well the adapter would already play with that. We could then use the perpendicular flap, for example.

Another interesting case could be the lid-driven cavity with a flexible bottom, which would be a nice validation example as well.

@fsimonis have you already tried anything with OpenFOAM and remeshing?

Triggered by https://precice.discourse.group/t/dynamic-coupling-interface-for-openfoam/2662

fsimonis:
This feature has also been requested by @Crazy-Rich-Meghan.

fsimonis:
Some notes we learned:

## Timing

Currently, the meshing is separately timed/controlled as the remeshing using `mesh.update()` is out of the control of preCICE. We need to align this with the requirement of preCICE to only remesh in the first iteration of a time window.

Using the new [`mesh.controlledUpdate()`](https://gitlab.com/openfoam/core/openfoam/-/blob/master/src/dynamicFvMesh/dynamicFvMesh/dynamicFvMesh.C?ref_type=heads#L167) could help. This is used in some solvers for beta testing like [pimpleFoam](https://gitlab.com/openfoam/core/openfoam/-/blob/master/applications/solvers/incompressible/pimpleFoam/pimpleFoam.C?ref_type=heads#L146).
More on [gitlab](https://gitlab.com/openfoam/core/openfoam/-/merge_requests/306)

## Checkpointing

The mesh update change the mesh as well as the dofs of fields defined on it. This directly influences the checkpointing, meaning the mesh update and the writing of the checkpoint needs to be synchronized.

If we assume that the remeshing is only allowed in the first iteration of a time window, then the procedure should looks as follows:

1. time window converges
2. read data from precice
3. update mesh if required which also takes care of remapping the read data to the new discretization. (needs confirmation)
4. create the checkpoint with the updated dofs.
6. perform the time loop

MakisH:
[...]

Issue #324, opened 2024-03-19
Not checkpointing non-geometric fields


MakisH:
Our `setupCheckpointing()` currently looks like this:

https://github.com/precice/openfoam-adapter/blob/ac79b78e4b094dc354bbbe6850d353f201012c63/Adapter.C#L890-L924

This looks for fields of various types and adds them to a list of fields to checkpoint. It works nicely for geometric fields, but there are more that we need to track.

In particular, in private communication, users have reported issues with the residuals. We currently don't checkpoint the general solution state metrics, which can lead OpenFOAM having a wrong impression on the actual convergence, and exiting earlier, in particular when using only one outer PIMPLE iteration. The relative tolerances are absolute zero in that case, and a workaround seems to be (if I understand correctly) to enforce multiple PIMPLE iterations (`nCorrectors`).

Adding `Info<< mesh_.thisDb().toc() << endl;` prints the table of contents of the objects registry. 

For the `fluid-openfoam` case (pimpleFoam) of the perpendicular flap, this prints:

```
30
(
solutionControl
Force
points
wallDist
neighbour
faceDiffusivity
MRFProperties
pointMesh
pointDisplacement
faces
U
turbulenceProperties
fvSchemes
preciceDict
fvOptions
faceZones
fvSolution
Uf
cumulativeContErr
nu
phi
owner
data
cellZones
boundary
dynamicMeshDict
p
cellDisplacement
pointZones
transportProperties
)

---[preciceAdapter] [DEBUG] Checkpoint nu : volScalarField
---[preciceAdapter] [DEBUG] Checkpoint p : volScalarField
---[preciceAdapter] [DEBUG] Checkpoint Force : volVectorField
---[preciceAdapter] [DEBUG] Checkpoint U : volVectorField
---[preciceAdapter] [DEBUG] Checkpoint cellDisplacement : volVectorField
---[preciceAdapter] [DEBUG] Checkpoint faceDiffusivity : surfaceScalarField
---[preciceAdapter] [DEBUG] Checkpoint phi : surfaceScalarField
---[preciceAdapter] [DEBUG] Checkpoint Uf : surfaceVectorField
---[preciceAdapter] [DEBUG] Checkpoint pointDisplacement : pointVectorField
---[preciceAdapter] [DEBUG] Writing a checkpoint...
---[preciceAdapter] [DEBUG] Stored time value t = 0.000000
---[preciceAdapter] [DEBUG] Storing mesh points...
[...]

PR #387, opened 2026-01-14
Mesh checkpointing refactor

vidulejs:
## Overview

This pull request removes redundant checkpointing methods, which were either unused or had no effect. More detail on this later. This results in the large amount of line removals of this PR. In addition I refactored and cleaned up the mesh points checkpointing. 

This pull request is one half of a larger refactor - PR #369, which also contains logic to correctly checkpoint mesh volumes, however it became clear that mesh volume checkpointing should not be merged at this time. 

All methods and comments related to volume checkpointing were completely removed, since they haven't served any purpose so far. In the future the volume checkpointing could be re-implemented based on PR #369, however the only effect would be on implicit subcycling (see discussion on that PR). 

**Changes:**
- Removed volume checkpointing methods and comments
- Removed comments and checkpoints for mesh fields which are unused (meshSurfaceVectorFields_, meshVolVectorFields_, volScalarInternalFields_) 
- Updated multi-line debug statements to one-liners.
- Mesh points checkpointing refactor
  - Changed the checkpoints from references to pointers, just like the rest of the checkpoints
  - Renamed oldMeshPoints to meshOldPoints (like fvMesh.oldPoints() method)
  - Fixed subtle issue in mesh oldPoints restoration, added code comment, see below

### Cleanup and refactor
This relates to the bulk of line removals in this branch.

Just like checkpointed fields of physical variables, there were different types of checkpointed mesh fields, however they were unnecessary and empty. There were no `meshVolVectorFields_` or `meshSurfaceVectorFields_` (none that needed checkpointing) . The only checkpointed mesh field is face motion flux `mesh_.phi()` of type `surfaceScalarField`. As pointed to by in a comment in `setupMeshCheckpointing()`:
> // The other mesh <type>Fields:
> //      C
> //      Cf
> //      Sf
> //      magSf
> //      delta
> // are updated by the function fvMesh::movePoints. Only the meshPhi needs checkpointing.

Edit: See new comment below with further explanation.

### Mesh Points Checkpointing
- Store `meshPoints_` and `meshOldPoints_` as pointers instead of objects for consistency with the rest of the codebase. These are actually pointers to the copies managed by the adapter.

- The main logic of restoring mesh point checkpoints is like this:
```cpp
    // Reload mesh points
    const_cast<Foam::fvMesh&>(mesh_).movePoints(*meshPoints_);

    // polyMesh.movePoints will only update oldPoints
    // if (curMotionTimeIndex_ != time().timeIndex())
    // I.e., first implicit iteration
    const_cast<pointField&>(mesh_.oldPoints()) = *meshOldPoints_;
 ```
 
Otherwise `mesh_.oldPoints()` would not get updated inside of `fvMesh.movePoints()` -> `polyMesh.movePoints()` in implicit iterations. 

Potentially this has can have no effect on the results, since we already checkpoint face motion flux `mesh_.phi()`, which is the delta between points and oldPoints. However, there will be a difference when doing implicit subcycling, see the discussion in PR #369. 

TODO list:

### CODE REVIEW

**Comment by vidulejs** [comment]
The reason why we do not need to checkpoint the mesh fields is that they are calculated on-demand once the mesh points move. This was also pointed to by the comment that was already in the OpenFOAM-adapter. Another reason why we don't need checkpoints for these fields is that they do not have `.oldTime()` method and thus do not participate in the time integration scheme. When we reload the checkpoint we will call `fvMesh::movePoints()`, which will in turn invalidate the old versions of these fields and recompute them in `fvMesh::updateGeomNotOldVol()`.

```cpp
void Foam::fvMesh::movePoints(const pointField& p)
{
   // [...] The comment below is not mine, but from the OpenFOAM source code.

    polyMesh::movePoints(p);

    // Update or delete the local geometric properties as early as possible so
    // they can be used if necessary. These get recreated here instead of
    // demand driven since they might do parallel transfers which can conflict
    // with when they're actually being used.
    // Note that between above "polyMesh::movePoints(p)" and here nothing
    // should use the local geometric properties.
    updateGeomNotOldVol();

   // [...]
}
```

```cpp
void Foam::fvMesh::updateGeomNotOldVol()
{
    const bool haveV = (VPtr_ != nullptr);
    const bool haveSf = (SfPtr_ != nullptr);
    const bool haveMagSf = (magSfPtr_ != nullptr);
    const bool haveCP = (CPtr_ != nullptr);
    const bool haveCf = (CfPtr_ != nullptr);

    clearGeomNotOldVol();

    // Now recreate the fields
    if (haveV)
    {
        (void)V();
    }

    if (haveSf)
    {
        (void)Sf();
    }

    if (haveMagSf)
    {
        (void)magSf();
    }

    if (haveCP)
    {
        (void)C();
    }

    if (haveCf)
    {
        (void)Cf();
    }
}
```

Adapter.C
```diff
@@ -1452,72 +1299,24 @@ void preciceAdapter::Adapter::writeMeshCheckpoint()
...
-}
+    DEBUG(adapterInfo("Storing mesh points..."));
 
-void preciceAdapter::Adapter::writeVolCheckpoint()
-{
-    DEBUG(adapterInfo("Writing a mesh volumes checkpoint..."));
+    // Store mesh points
+    // swap pointers
+    *(meshOldPoints_) = *(meshPoints_);
+    *(meshPoints_) = mesh_.points();
```
MakisH, line 1313: I guess this (and the previous way, if we use the same sizes) will break once we go for remeshing, correct?

#382 

Not necessary to solve it now, but as a first impulse for discussion.

MakisH (commented):
Thanks a lot for this long-overdue cleanup!

It's nice to see so much code deleted. Let's also add an overview of checkpointed and non-checkpointed fields in the documentation (as you had a draft in https://github.com/precice/openfoam-adapter/issues/324 and as discussed on Matrix). The most fitting place would be a new section in https://precice.org/adapter-openfoam-extend.html (formulated from the perspective "this is what the adapter currently checkpoints, you might need to add more for a different use case").

vidulejs:
Added documentation on checkpointed fields in extend.md

docs/images/docs-adapter-openfoam-implicit-loop.png
MakisH, line 1: What did you use to generate this figure? Is there a source file to attach as well?

docs/extend.md
```diff
@@ -28,3 +28,69 @@ section of the `Make/options`. Since the adapter is a shared library,
...
+
+### Checkpointed Fields
+
+The adapter automatically finds and checkpoints all objects registered in the OpenFOAM object registry (`mesh_.thisDb()`) that match the following types:
+
+#### Volume Fields
+
+- `volScalarField`: e.g., `p`, `T`, `k`, `omega`, `nut`, `nu`, `rho`
+- `volVectorField`: e.g., `U`, `Force`, `cellDisplacement`
+- `volTensorField`: e.g., `grad(U)` (if exists, relevant closed issue [#158](docs-adapter-implicit-loop))
```
MakisH:
```suggestion
- `volTensorField`: e.g., `grad(U)` (if exists, relevant closed issue [#158](https://github.com/precice/openfoam-adapter/issues/158))
```

docs/extend.md
```diff
@@ -28,3 +28,69 @@ section of the `Make/options`. Since the adapter is a shared library,
...
+- `surfaceVectorField`: e.g., `Uf`
+- `surfaceTensorField`
+
+#### Point Fields
+
+- `pointScalarField`
+- `pointVectorField`: e.g., `pointDisplacement`
+- `pointTensorField`
+
+The adapter will only create the checkpoint list once during initialization. I.e., after the first solver step it is assumed that no new fields have been created that need to be checkpointed.
```
MakisH:
```suggestion
The adapter will only create the checkpoint list once during initialization: After the first solver step, it is assumed that no new fields have been created that need to be checkpointed.
```

docs/extend.md
```diff
@@ -28,3 +28,69 @@ section of the `Make/options`. Since the adapter is a shared library,
 another missing library will trigger an "undefined symbol" runtime error.
 
 See also the notes and discussion in [issue #7: Create a module for fluid-structure interaction](https://github.com/precice/openfoam-adapter/issues/7).
+
+## Implicit Coupling Diagram
```
MakisH: We use sentence case (affects most new headings).

docs/extend.md
```diff
@@ -28,3 +28,69 @@ section of the `Make/options`. Since the adapter is a shared library,
...
+Certain fields are purposely not checkpointed, see issue [#324](https://github.com/precice/openfoam-adapter/issues/324) and explanation in [PR #387](https://github.com/precice/openfoam-adapter/pull/387). This is either because they are re-calculated when needed "on-demand" or because they are essentially static (dictionaries assumed unchanging).
+
+The following fields are recalculated by OpenFOAM when `fvMesh::movePoints` is called, so the adapter does not store them:
+
+- `C` (Cell Centers)
+- `Cf` (Face Centers)
+- `Sf` (Face Area Vectors)
+- `magSf` (Face Area Magnitudes)
+- `delta` (Face-Center to Cell-Center distances)
+- `V`, `V0`, `V00` (Cell Volumes): These are already properly checkpointed by OpenFOAM itself, except when subcycling (future pull request [#369](https://github.com/precice/openfoam-adapter/pull/369)).
```
MakisH: Can we link to an issue instead of a PR here?

docs/extend.md
MakisH, line 1: Overall, I think this still makes more sense as a standalone file page. We could start a page `OpenFOAM adapter developer documentation` (`adapter-openfoam-dev-docs`), right after `Extending`.

Easier alternative for now: Just add a sentence that makes it fit with "extending".

docs/extend.md
```diff
@@ -28,3 +28,69 @@ section of the `Make/options`. Since the adapter is a shared library,
...
+To handle mesh motion (ALE), the adapter explicitly checkpoints:
+
+- Mesh Points `mesh_.points()` and `mesh_.oldPoints()`
+- Mesh Motion Flux `mesh_.phi()`
+
+### Non-Checkpointed Fields
+
+Certain fields are purposely not checkpointed, see issue [#324](https://github.com/precice/openfoam-adapter/issues/324) and explanation in [PR #387](https://github.com/precice/openfoam-adapter/pull/387). This is either because they are re-calculated when needed "on-demand" or because they are essentially static (dictionaries assumed unchanging).
+
+The following fields are recalculated by OpenFOAM when `fvMesh::movePoints` is called, so the adapter does not store them:
```
MakisH, line 82:
```suggestion
#### Derived fields

The following fields are recalculated by OpenFOAM when `fvMesh::movePoints` is called, so the adapter does not store them:
```

docs/extend.md
```diff
@@ -28,3 +28,69 @@ section of the `Make/options`. Since the adapter is a shared library,
 another missing library will trigger an "undefined symbol" runtime error.
 
 See also the notes and discussion in [issue #7: Create a module for fluid-structure interaction](https://github.com/precice/openfoam-adapter/issues/7).
+
+## Implicit Coupling Diagram
+
+In implicit coupling, the adapter manages the coupling loop, which entails checkpointing the simulation state (fields and mesh), to restore the initial state at the beginning of each coupling time step. As the coupling data is exchanged, the adapter repeats the time step until convergence. See this diagram for an overview:
+
+![Implicit checkpointing diagram](images/docs-adapter-openfoam-implicit-loop.png)
```
MakisH, line 39: Let's drop the implicit coupling section and move the figure directly in Checkpointing, as context.

docs/extend.md
vidulejs: I only have the old issue [#58](https://github.com/precice/openfoam-adapter/issues/58) to link to here. I didn't make my own issue on volume checkpointing actually.

docs/images/docs-adapter-openfoam-implicit-loop.png
vidulejs: Added source file.

docs/extend.md
MakisH: I rephrased it as "to extend".

docs/extend.md
MakisH: Linked to the exact comment as "related discussion".

MakisH (approved):

PR #391, opened 2026-01-28
Allow for case-insensitive exchanged field names

### LINKED ISSUE #390: Exchanged fields are case-sensititve
Some solvers (in this case, OpenRadioss) currently only accept names such as `DISPLACEMENTS`. In the OpenFOAM adapter, we look for `Displacement` in the beginning of the string.

We could first convert the `dataName` to a consistent case, and compare that to the given list.

In the longer term, we should separate the name of the field from what it is connected to in the adapter, since we often have name clashes. In the short term, we could just handle different permutations.

Related to https://github.com/precice/preeco-orga/issues/36, would be superceded by #354.

MakisH:
closes #390

The system tests fail for unrelated reasons (https://github.com/precice/precice/issues/2432) but the part that concerns this PR is ok.

TODO list:

### DIFF
--- a/Utilities.C
+++ b/Utilities.C
@@ -66,9 +66,9 @@ void adapterInfo(const std::string message, const std::string level)
     return;
 }
 
-bool similar(std::string s, std::string with)
+bool matchingStrings(std::string s, std::string match)
 {
     std::transform(s.begin(), s.end(), s.begin(), [](unsigned char c) { return std::tolower(c); });
-    std::transform(with.begin(), with.end(), with.begin(), [](unsigned char c) { return std::tolower(c); });
-    return s.find(with) == 0;
+    std::transform(match.begin(), match.end(), match.begin(), [](unsigned char c) { return std::tolower(c); });
+    return s.find(match) == 0;
 }
--- a/Utilities.C
+++ b/Utilities.C
@@ -1,4 +1,5 @@
 #include "Utilities.H"
+#include <cctype>
 
 using namespace Foam;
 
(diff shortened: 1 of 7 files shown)


Utilities.C
```diff
@@ -64,3 +65,10 @@ void adapterInfo(const std::string message, const std::string level)
 
     return;
 }
+
+bool similar(std::string s, std::string with)
```
vidulejs: IMO confusingly vague function name. What about something like case_insensitive ?

vidulejs (commented):
Alternatively, why not instead modify the `CouplingDataUser` to store all data names in lower case, then check against that? Then we'd just modify the hard-coded names to lowercase as well.

```cpp
// apply lower casing in one of these methods.

void preciceAdapter::CouplingDataUser::setDataName(std::string dataName)
{
    dataName_ = std::move(dataName);
}

const std::string& preciceAdapter::CouplingDataUser::dataName()
{
    return dataName_;
}
```

Utilities.C
MakisH: Good point, renamed it to `matchingStrings`, with the assumption that `matching` does not mean `equal`, but `related to` (per your suggestion).

MakisH:
> Alternatively, why not instead modify the `CouplingDataUser` to store all data names in lower case, then check against that?

Refactoring the `CouplingDataUser` could be an alternative, but I would like to avoid a larger refactoring without significant benefit (makes it difficult to port to forks).

The long-term solution would be decoupling these and using the `solver_name` in the schema (https://github.com/precice/openfoam-adapter/pull/354). This solution should not be needed in v2 of the adapter.

MakisH:
Approved in a private message.

PR #392, opened 2026-01-29
Clarify field support for solvers in README

MakisH:
Replaces the list of fields that the adapter can read/write, with more detailed tables (one per module). The tables clarify which field can be written/read on which mesh locations, and whether mesh connectivity is provided.

It also mentions fields added with #380.

The tables need to be cross-checked. I am, in particular, unsure about the support of mesh connectivity.

The CI is currently failing for unrelated reasons:

- Building: See [issue with OpenFOAM packages](https://gitlab.com/openfoam/core/openfoam/-/issues/3496) (solved)
- Formatting: See [action issue](https://github.com/jidicula/clang-format-action/issues/267)


MakisH:
Overall, it would be easier to add a table with fields as rows and as columns:

- Points (column group)
  - Read
  - Write 
- Faces (column group)
  - Read
  - Write 
- Cells (column group)
  - Read
  - Write 

and then have `X` or similar to mark what is supported, or `Fluid`/`Solid` if only one of them is possible.

We should then also add the missing documentation from #380.

docs/config.md
```diff
@@ -217,7 +217,7 @@ solver              displacementLaplacian;
 
 #### FF
 
-The fluid-fluid coupling module supports reading and writing `Pressure`, `Velocity`, `PressureGradient`, `VelocityGradient`, `FlowTemperature`, `FlowTemperatureGradient`, `Alpha`, `AlphaGradient` and the face flux `Phi`.
+The fluid-fluid coupling module supports reading and writing `Pressure`, `Velocity`, `PressureGradient` (surface-normal gradient), `PressureGradientFull` (full pressure gradient), `VelocityGradient`, `FlowTemperature`, `FlowTemperatureGradient`, `Alpha`, `AlphaGradient`, `Phi` (face flux), `DragForce`, `ExplicitMomentum`, `ImplicitMomentum`.
```
davidscn: Maybe it would make sense to put the last three fields into the context of CFD-DEM, I am not sure how convenient such fields would be beyond that. Having them under fluid-fluid is also somewhat questionable.

docs/README.md
```diff
@@ -12,32 +12,58 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
-- Phase flux (phi) (read + write)
-
-In addition, the adapter supports the following fields in a volume coupling setup:
-
-- Temperature (write)
-- Pressure (write)
-- Velocity (read + write)
-
-All features of preCICE are supported, including implicit coupling and nearest-projection mapping. Even though OpenFOAM is 3D, this adapter can also work in the 2D mode of preCICE, defining only one layer of interface nodes (automatically).
+This adapter has been demonstrated on different use cases (conjugate heat transfer, fluid-structure interaction, fluid-fluid coupling), both in 2D (3D with one layer of cells in the z-axis) and 3D, and both for flow and solid OpenFOAM-based solvers (see [tutorials](https://precice.org/tutorials.html)).
```
davidscn:
```suggestion
This adapter has been demonstrated on different use cases (conjugate heat transfer, fluid-structure interaction, fluid-fluid coupling, CFD-DEM), both in 2D (3D with one layer of cells in the z-axis) and 3D, and both for flow and solid OpenFOAM-based solvers (see [tutorials](https://precice.org/tutorials.html)).
```

docs/README.md
```diff
@@ -12,32 +12,58 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
+- ***:** mesh connectivity supported (for e.g., nearest-projection mapping)
+
+### Module: Conjugate heat transfer
+
+| Field | Write | Read |  Config prefix |
+| --- | --- | --- | --- |
+| Heat flux | N*, F | N, F* | `Heat-Flux` |
+| Heat transfer coefficient | N*, F | N, F* | `Heat-Transfer-Coefficient` |
+| Sink temperature | N*, F | N, F* | `Sink-Temperature` |
+| Temperature | N*, F, C | N, F*, C | `Temperature` |
```
davidscn, line 32: I know where the confusion here stems from: the connectivity element in the OpenFOAM adapter was initially added with nearest-projection mappings in mind, for the above data fields, the following holds

- all these fields are mapped `consistent`
- for a consistent mapping, the `write` side needs to provide connectivity
- the fields in OpenFOAM itself live on the face centers, so one would read them on the face centers.
- the adapter generally only allows the definition of connectivity between face nodes

That said, for reading data, connectivity is not relevant, one would configure the face centers and leave the connectivity definition to the writing side. Internally, we only support reading on `F` (and `C`) for reading data and not `N`. Putting the `F*`  on the read side is pointless (the connectivity is never used), but also doesnt work in terms of the adapter. 

So correct would be

```
| Heat flux | N*, F | F | `Heat-Flux` |
```

See also: https://github.com/precice/tutorials/tree/develop/flow-over-heated-plate-nearest-projection

docs/README.md
```diff
@@ -12,32 +12,58 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
+
+All fields are supported for both flow (compressible/incompressible) and basic (e.g., laplacianFoam) solvers.
+
+### Module: Fluid-structure interaction
+
+| Field | Write | Read | Config prefix |
+| --- | --- | --- | --- |
+| Displacement: absolute | N*, F* | N*, F* | `Displacement` |
```
davidscn:
```suggestion
| Displacement: absolute | N*, [F*](https://github.com/precice/openfoam-adapter/issues/153) | N*, F* | `Displacement` |
```
Should we link here to the issue? For `Displacement`, `N*` and `F*` on the read side is technically correct, but would be similar pointless as discussed above.

docs/README.md
```diff
@@ -12,32 +12,58 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
+| Displacement: relative | N*, F* | N*, F* | `DisplacementDelta` |
+| Force | F* (flow) | F* | `Force` |
+| Stress | F* (flow) | F* | `Stress` |
```
davidscn, line 43:
```suggestion
| Force | F (flow) | F | `Force` |
| Stress | F (flow) |  | `Stress` |
```
What does flow here mean? I think reading Stress is not supported.

docs/README.md
```diff
@@ -12,32 +12,58 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
+
+Displacement reading and writing is supported both for flow (compressible or incompressible) and structure solvers.
+Force and stress writing in only supported for flow solvers, reading for both.
+
+### Module: Fluid-fluid coupling
```
davidscn, line 48: I think all `F*` in the following table need to be a `F`

vidulejs:
In Adapter.C,  line 178, it says FSI with meshConnectivity is unsupported and will throw an error.

```cpp
    // If the FSI module is enabled, create it, read the
    // FSI-specific options and configure it.
    if (FSIenabled_)
    {
        // Check for unsupported FSI with meshConnectivity
        for (uint i = 0; i < interfacesConfig_.size(); i++)
        {
            if (interfacesConfig_.at(i).meshConnectivity == true)
            {
                adapterInfo(
                    "You have requested mesh connectivity (most probably for nearest-projection mapping) "
                    "and you have enabled the FSI module. "
                    "Mapping with connectivity information is not implemented for FSI, only for CHT-related fields. "
                    "error");
                return;
            }
        }
```

docs/README.md
```diff
@@ -12,32 +12,58 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
+
+| Field | Write | Read | Config prefix |
+| --- | --- | --- | --- |
+| Drag force | F*, C | F*, C | `DragForce` |
+| Momentum: explicit | F*, C | F*, C | `ExplicitMomentum` |
+| Momentum: implicit | F*, C | F*, C | `ImplicitMomentum` |
+| Phase flux | F* | F* | `Phi` |
+| Phase fraction | F*, C | F*, C | `Alpha` |
```
davidscn:
```suggestion
| Volume fraction | F*, C | F*, C | `Alpha` |
```
or is Phase clear here?

davidscn (changes requested):
I didnt check all the `C`s in the FF module

docs/README.md
vidulejs: Only saw your review now. I was also confused about mesh connectivity on `F*` on read side. 

In Adapter.C, line 100, we allow mesh connectivity for faceNodes only. 
```cpp
                if (interfaceConfig.meshConnectivity && (interfaceConfig.locationsType == "faceCenters" || interfaceConfig.locationsType == "volumeCenters" || interfaceConfig.locationsType == "volumeCentres"))
                {
                    DEBUG(adapterInfo("Mesh connectivity is not supported for faceCenters or volumeCenters. \n"
                                      "Please configure the desired interface with the locationsType faceNodes. \n"
                                      "Have a look in the adapter documentation for detailed information.",
                                      "error"));
                    return;
                }
```
Also in Interface.C we only do 
`precice_.setMeshTriangles(meshName_, triVertIDs);`  if location is faceNodes.

vidulejs: But then `F*` on write side is also impossible to configure.

davidscn:
> In Adapter.C, line 178, it says FSI with meshConnectivity is unsupported and will throw an error.

I think this has similar historical reasons: we started with conventional FSI, i.e., reading displacement and writing forces. In such cases, the solid solver needs to provide the connectivity, so there was no real point in conisdering this scenario. Since we now also have solid solvers in OpenFOAM, it could make sense to allow the displacement with connectivity case.

docs/README.md
MakisH: What I want to emphasize is that writing forces for an OpenFOAM structure solver (e.g., solidDisplacementFoam or solids4Foam) is not supported.

vidulejs:
I've created a new PR to fix the outdated mesh connectivity statements in all of the modules.

https://github.com/precice/openfoam-adapter/pull/393

MakisH:
Self-reminder: Check and update the documentation with a clarification:

- Mesh connectivity only makes sense for the writing side
- If a field supports connectivity, this is only for the nodes
- Temperature is available on the face centers, but if connectivity is requested (i.e., on the nodes only), we interpolate from the faces to the nodes

This is already there, but maybe we can make it easier/clearer earlier.

davidscn:
> Mesh connectivity only makes sense for the writing side

for consistent data, for coservative data, it makes only sense for the reading side

vidulejs:
Let's wrap this PR up.

1. Mesh connectivity (*) is only relevant for the Write side currently. Even though we had allowed mesh connectivity on Read, there is actually no implementation on the adapter side [(comment)](https://github.com/precice/openfoam-adapter/pull/393#discussion_r2832409906). We do not currently distinguish between read and write in `isLocationTypeSupported` either.

2. Mesh connectivity (*) is impossible for anything other than `locations: faceNodes`. There is a global block in `Adapter.C`, which will throw an error if it's anything other than Face Nodes. In `Interface.C`, we provide preCICE with the mesh connectivity data only in this case. 

3. FSI now allows mesh connectivity since PR [393](https://github.com/precice/openfoam-adapter/pull/393).

docs/README.md
```diff
@@ -12,32 +12,58 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
+
+| Field | Write | Read | Config prefix |
+| --- | --- | --- | --- |
+| Displacement: absolute | N*, [F*](https://github.com/precice/openfoam-adapter/issues/153) | N*, F* | `Displacement` |
+| Displacement: relative | N*, F* | N*, F* | `DisplacementDelta` |
+| Force | F* (flow) | F* | `Force` |
+| Stress | F* (flow) | F* | `Stress` |
+
+Displacement reading and writing is supported both for flow (compressible or incompressible) and structure solvers.
+Force and stress writing in only supported for flow solvers, reading for both.
```
MakisH:
```suggestion
Force and stress writing is only supported for flow solvers, reading for both.
```

docs/README.md
MakisH: - Removed the connectivity from Force and Stress writing (in the code, we now have `if (meshConnectivity) return false;` since #393).
- Removed the reading of stresses (implemented in https://github.com/precice/openfoam-adapter/pull/417).
- Clarified `(flow solvers)` instead of `(flow)`.
- Also removed the reading of DisplacementDelta (implemented in https://github.com/precice/openfoam-adapter/pull/418).

docs/README.md
MakisH: I removed the connectivity from most fields (which even had `false` in the `if (meshConnectivity_)`).

For the `AlphaGradient` and `Phi`, the `isLocationTypeSupported` was not distinguishing for `meshConnectivity`, but this looks wrong. I added the guard in the code and removed the connectivity from the docs. These fields were not modified in #393 (iirc, these fields were already there).

docs/README.md
```diff
@@ -15,32 +15,65 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
+| Field | Write | Read | Config prefix |
+| --- | --- | --- | --- |
+| Drag force | F, C | F, C | `DragForce` |
+| Momentum: explicit | F, C | F, C | `ExplicitMomentum` |
+| Momentum: implicit | F, C | F, C | `ImplicitMomentum` |
+| Phase flux | F | F | `Phi` |
+| Volume fraction | F, C | F, C | `Alpha` |
+| Phase fraction gradient | F | F | `AlphaGradient` |
+| Pressure | F, C | F, C | `Pressure` |
+| Pressure: full gradient | F, C* | F | `PressureGradientFull` (TODO: read-only) |
```
MakisH: I am no longer sure what I meant with the `(TODO: read-only)`. Do you have any idea?

I removed the `C*` (write), which does not make sense to me.


```suggestion
| Pressure: full gradient | F, C | F | `PressureGradientFull` |
```

docs/README.md
MakisH: I removed the nodes reading from all CHT fields, converted all `F*` to `F`, and added a note on how we interpolate from faces to nodes when writing.

MakisH (commented):
I went again through the complete discussion, modified all tables, and added notes. This now makes much more sense to me, even though the `isLocationTypeSupported` mechanism remains confusing. #393 has been very helpful in settling this (thanks again, @vidulejs).

@davidscn do you see anything else wrong before we can merge this? As more PRs that add fields start to appear, having this table would be helpful.

docs/README.md
```diff
@@ -15,32 +15,65 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
+| Phase fraction gradient | F | F | `AlphaGradient` |
+| Pressure | F, C | F, C | `Pressure` |
+| Pressure: full gradient | F, C | F | `PressureGradientFull` (TODO: read-only) |
+| Pressure: surface-normal gradient | F | F | `PressureGradient` |
+| Temperature | F | F | `FlowTemperature` |
+| Temperature surface-normal gradient | F | F | `FlowTemperatureGradient` |
+| Velocity | F, C | F, C | `Velocity` |
+| Velocity surface-normal gradient | F | F | `VelocityGradient` |
+
+All fields assume a flow solver, and some of these fields are mainly meant for a CFD-DEM coupling (`DragForce`, `ExplicitMomentum`, `ImplicitMomentum`, `PressureGradientFull`).
```
MakisH: We probably don't need to add `PressureGradientFull` in this list:

```suggestion
All fields assume a flow solver, and some of these fields are mainly meant for a CFD-DEM coupling (`DragForce`, `ExplicitMomentum`, `ImplicitMomentum`).
```

docs/README.md
```diff
@@ -15,32 +15,65 @@ This preCICE adapter is a plug-in (function object) for OpenFOAM, which can work
...
+| --- | --- | --- | --- |
+| Displacement: absolute | N*, [F*](https://github.com/precice/openfoam-adapter/issues/153) | N, F | `Displacement` |
+| Displacement: relative | - | N, F | `DisplacementDelta` |
+| Force | F (flow solvers) | F | `Force` |
+| Stress | F (flow solvers) | - | `Stress` |
+
+Displacement reading and writing are supported for both flow (compressible or incompressible) and structure solvers.
+For mesh nodes, the `pointDisplacement` field is used; for face centers, the `cellDisplacement` field is used.
+
+Only flow solvers can write forces or stresses. Solid solvers can read forces, but not stresses.
```
MakisH, line 56:
```suggestion
Only flow solvers can write forces or stresses. Solid solvers can read forces, but not stresses.

Originally, only flow OpenFOAM solvers have been considered. Reading more fields and providing connectivity by solid solvers can still be implemented.
```

davidscn: > Originally, only flow OpenFOAM have been considered

flow OpenFOAM sounds a bit strange to me

docs/README.md
davidscn: No clue. I think we planned to rename this at some point to Pressure Gradient? I have never tested reading the full gradient either

davidscn (approved):
Looks good now

PR #393, opened 2026-02-19
Update meshConnectivity claims

vidulejs:
### Updated `isLocationTypeSupported` across all modules
The `isLocationTypeSupported` methods were not entirely correct w.r.t. mesh connectivity support (implementation did not match the statements) and inconsistent across the modules. 

Previously we allowed mesh connectivity for face centers when reading, however that is wholly irrelevant, because only the writing participant determines mesh connectivity. See [David's comment](https://github.com/precice/openfoam-adapter/pull/392#discussion_r2817737302):
> That said, for reading data, connectivity is not relevant, one would configure the face centers and leave the connectivity definition to the writing side,

Furthermore, this condition was actually unreachable by the adapter, because in Adapter.C, line 100, we allow mesh connectivity only when location is face nodes.

```cpp
if (interfaceConfig.meshConnectivity && (interfaceConfig.locationsType == "faceCenters" || interfaceConfig.locationsType == "volumeCenters" || interfaceConfig.locationsType == "volumeCentres"))
{
    DEBUG(adapterInfo("Mesh connectivity is not supported for faceCenters or volumeCenters. \n"
                      "Please configure the desired interface with the locationsType faceNodes. \n"
                      "Have a look in the adapter documentation for detailed information.",
                      "error"));
    return;
}
```

### Updated support

- CHT module: mesh connectivity supported for writing faceNodes.
- FF module: no mesh connectivity support.
- FSI module:  Since we now have solid solvers in OpenFOAM, we can allow mesh connectivity for displacement, see [David's comment](https://github.com/precice/openfoam-adapter/pull/392#issuecomment-3919455527). 

TODO list:

### DIFF
--- a/FSI/Displacement.C
+++ b/FSI/Displacement.C
@@ -142,7 +142,15 @@ void preciceAdapter::FSI::Displacement::read(double* buffer, const unsigned int 
 
 bool preciceAdapter::FSI::Displacement::isLocationTypeSupported(const bool meshConnectivity) const
 {
-    return (this->locationType_ == LocationType::faceCenters || this->locationType_ == LocationType::faceNodes);
+    // Solid solver can allow connectivity for writing displacement
+    if (meshConnectivity)
+    {
+        return (this->locationType_ == LocationType::faceNodes);
+    }
+    else
+    {
+        return (this->locationType_ == LocationType::faceCenters || this->locationType_ == LocationType::faceNodes);
+    }
 }
 
 std::string preciceAdapter::FSI::Displacement::getDataName() const
--- a/changelog-entries/394.md
+++ b/changelog-entries/394.md
@@ -1,0 +1,1 @@
+Refactored `isLocationTypeSupported` across all modules. FSI now allows mesh connectivity when writing Displacement.
--- a/changelog-entries/394.md
+++ b/changelog-entries/394.md
@@ -1,1 +1,1 @@
-Refactored `isLocationTypeSupported` across all modules. FSI now allows mesh connectivity when writing Displacement.
+Updated `isLocationTypeSupported` across all modules to remove unreachable conditions and to better reflect the implementation. FSI now allows mesh connectivity when writing Displacement.
--- a/Adapter.C
+++ b/Adapter.C
@@ -234,20 +234,6 @@ void preciceAdapter::Adapter::configFileRead()
     // FSI-specific options and configure it.
     if (FSIenabled_)
     {
-        // Check for unsupported FSI with meshConnectivity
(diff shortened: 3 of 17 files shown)


changelog-entries/394.md
```diff
@@ -0,0 +1 @@
+Refactored `isLocationTypeSupported` across all modules. FSI now allows mesh connectivity when writing Displacement.
```
MakisH:
```suggestion
Updated `isLocationTypeSupported` across all modules to remove unreachable conditions and to better reflect the implementation. FSI now allows mesh connectivity when writing Displacement.
```

MakisH (approved):
Live discussion with @vidulejs, went through all the files together, looks good to me (and documentation will follow in #392).

@davidscn do you see anything wrong here?

FSI/Displacement.C
```diff
@@ -142,7 +142,15 @@ void preciceAdapter::FSI::Displacement::read(double* buffer, const unsigned int
 
 bool preciceAdapter::FSI::Displacement::isLocationTypeSupported(const bool meshConnectivity) const
 {
-    return (this->locationType_ == LocationType::faceCenters || this->locationType_ == LocationType::faceNodes);
+    // Solid solver can allow connectivity for writing displacement
+    if (meshConnectivity)
+    {
+        return (this->locationType_ == LocationType::faceNodes);
```
davidscn, line 148: From an implementation perspective, both displacement fields would actually allow using `faceCenters` with connectivity for reading data, although this would still be pointless from the (conventional) mapping perspective. However, maybe useful for `scaled-consistent` and similar preCICE features.

davidscn (commented):
I think it would be valuable to state the preCICE requirements when erroring out due to an unsupported locations type, i.e., when should a user care about which connectivity.

vidulejs:
> I think it would be valuable to state the preCICE requirements when erroring out due to an unsupported locations type, i.e., when should a user care about which connectivity.

Added more specific error message.

FSI/Displacement.C
vidulejs: As far as I understand, in order to support mesh connectivity on consistent read mapping, we would have to implement the interpolation from `faceNodes` back to `faceCenters`. The location type would then be configured `faceCenters`. However, we only provide preCICE with the connectivity information in Interface.C when location type is `faceNodes`.

davidscn: Ah ok, I was not aware of that, then the implementation doesn't add up, but ok, my comment doesnt apply then.

MakisH:
I understand that this is now ready to merge.

PR #394, opened 2026-02-24
Initialize meshPoints_ to null pointers

vidulejs:
Fixed bug where `storeMeshPoints()` would not store points because meshPoints_ pointer had garbage non-zero value.

```cpp
// Adapter.C
void preciceAdapter::Adapter::storeMeshPoints()
{
    if (!meshPoints_)
    {
        DEBUG(adapterInfo("Storing mesh points..."));
        // Add points and oldPoints
        meshPoints_ = new Foam::pointField(mesh_.points());
        meshOldPoints_ = new Foam::pointField(mesh_.oldPoints());
    }
    //... 
```

The fix was to initialize with null pointers:
```cpp
// Adapter.H
    //- Checkpointed mesh points, oldPoints
    Foam::pointField* meshPoints_ = NULL;
    Foam::pointField* meshOldPoints_ = NULL;
```
TODO list:

### DIFF
--- a/Adapter.H
+++ b/Adapter.H
@@ -155,8 +155,8 @@ private:
     // Vectors of pointers to the checkpointed mesh fields and their copies
 
     //- Checkpointed mesh points, oldPoints
-    Foam::pointField* meshPoints_;
-    Foam::pointField* meshOldPoints_;
+    Foam::pointField* meshPoints_ = NULL;
+    Foam::pointField* meshOldPoints_ = NULL;
 
     //- Checkpointed surfaceScalarField mesh fields (phi)
     std::vector<Foam::surfaceScalarField*> meshSurfaceScalarFields_;
--- a/Adapter.H
+++ b/Adapter.H
@@ -155,8 +155,8 @@ private:
     // Vectors of pointers to the checkpointed mesh fields and their copies
 
     //- Checkpointed mesh points, oldPoints
-    Foam::pointField* meshPoints_ = NULL;
-    Foam::pointField* meshOldPoints_ = NULL;
+    Foam::pointField* meshPoints_ = nullptr;
+    Foam::pointField* meshOldPoints_ = nullptr;
 
     //- Checkpointed surfaceScalarField mesh fields (phi)
     std::vector<Foam::surfaceScalarField*> meshSurfaceScalarFields_;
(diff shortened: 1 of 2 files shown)


Adapter.H
```diff
@@ -155,8 +155,8 @@ private:
     // Vectors of pointers to the checkpointed mesh fields and their copies
 
     //- Checkpointed mesh points, oldPoints
-    Foam::pointField* meshPoints_;
-    Foam::pointField* meshOldPoints_;
+    Foam::pointField* meshPoints_ = NULL;
+    Foam::pointField* meshOldPoints_ = NULL;
```
MakisH: Should be `nullptr`, no?


```suggestion
    Foam::pointField* meshPoints_ = nullptr;
    Foam::pointField* meshOldPoints_ = nullptr;
```

vidulejs: Right, `nullptr` is preferred to `NULL` as far as I can gather online.

We can replace also the other instances of `NULL` in Adapter.H, right above these lines?

```cpp
    //- preCICE solver interface
    precice::Participant* precice_ = nullptr;

    //- preCICE solver interface initialized
    bool preciceInitialized_ = false;

    //- Conjugate Heat Transfer module object
    CHT::ConjugateHeatTransfer* CHT_ = NULL;

    //- Fluid-Structure Interaction module object
    FSI::FluidStructureInteraction* FSI_ = NULL;

    //- Fluid-Fluid module object
    FF::FluidFluid* FF_ = NULL;

    //- Generic module object
    Generic::GenericInterface* Generic_ = NULL;
```

vidulejs: https://stackoverflow.com/questions/53963646/what-is-the-difference-between-null-and-null-in-c#53963689

MakisH: Even better: there is a core guideline https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es47-use-nullptr-rather-than-0-or-null

changelog-entries/394.md
MakisH, line 1: The respective points first appeared with https://github.com/precice/openfoam-adapter/pull/387, which has not been released. This PR does not need a changelog entry.

MakisH (approved):
Merging to get the system tests working again. I can update the rest of the pointers in another PR.

PR #395, opened 2026-02-24
Implement `flip-normal` operation adapter-wide inside `CouplingDataUser`

vidulejs:
I decided that the most appropriate place to implement the new `flip-normal` operation is inside of the `CouplingDataUser`, because it 'owns' the data and is associated with that specific data. Therefore the operation is decoupled from the modules and supported by all the modules.

There looks to be a lot of changes in a lot of files, but it's actually quite simple if you look commit by commit. I had to refactor the `Interface::addCouplingDataReader` and `Interface::addCouplingDataWriter` methods to accept `struct FieldConfig` instead of just `string dataName`, because I need to read the `flip-normal` flag per each `CouplingDataUser`.

```cpp
// Interface.C
void preciceAdapter::Interface::addCouplingDataReader(
    const FieldConfig& fieldConfig,
    preciceAdapter::CouplingDataUser* couplingDataReader)
{
    // Set the patchIDs of the patches that form the interface
    couplingDataReader->setDataName(fieldConfig.name);

    // Set the flip normal option
    couplingDataReader->setFlipNormal(fieldConfig.flip_normal);

// ...
```

Now the `CouplingDataUser` has new methods `setFlipNormal` and `applyFlipNormal`.

```cpp
// CouplingDataUser.C
void preciceAdapter::CouplingDataUser::setFlipNormal(bool flipNormal)
{
    flipNormal_ = flipNormal;
}

void preciceAdapter::CouplingDataUser::applyFlipNormal(double* dataBuffer, std::size_t size)
{
    if (flipNormal_)
    {
        for (std::size_t i = 0; i < size; ++i)
        {
            dataBuffer[i] *= -1.0;
        }
    }
}
```

Additionally, I noticed that I had named instances of the `FieldConfig` struct starting with uppercase, so the first commit is lowercasing them.

TODO list:

### DIFF
--- a/CouplingDataUser.C
+++ b/CouplingDataUser.C
@@ -24,6 +24,22 @@ const std::string& preciceAdapter::CouplingDataUser::dataName()
     return dataName_;
 }
 
+void preciceAdapter::CouplingDataUser::setFlipNormal(bool flipNormal)
+{
+    flipNormal_ = flipNormal;
+}
+
+void preciceAdapter::CouplingDataUser::applyFlipNormal(double* dataBuffer, std::size_t size)
+{
+    if (flipNormal_)
+    {
+        for (std::size_t i = 0; i < size; ++i)
+        {
+            dataBuffer[i] *= -1.0;
+        }
+    }
+}
+
 void preciceAdapter::CouplingDataUser::setPatchIDs(std::vector<int> patchIDs)
 {
     patchIDs_ = patchIDs;
--- a/CouplingDataUser.C
+++ b/CouplingDataUser.C
@@ -29,13 +29,13 @@ void preciceAdapter::CouplingDataUser::setFlipNormal(bool flipNormal)
     flipNormal_ = flipNormal;
 }
 
-void preciceAdapter::CouplingDataUser::applyFlipNormal(double* dataBuffer, std::size_t size)
+void preciceAdapter::CouplingDataUser::applyFlipNormal(precice::span<double> dataBuffer)
 {
     if (flipNormal_)
     {
-        for (std::size_t i = 0; i < size; ++i)
+        for (double& val : dataBuffer)
         {
-            dataBuffer[i] *= -1.0;
+            val *= -1.0;
         }
     }
 }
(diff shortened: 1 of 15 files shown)


CouplingDataUser.C
```diff
@@ -24,6 +24,22 @@ const std::string& preciceAdapter::CouplingDataUser::dataName()
     return dataName_;
 }
 
+void preciceAdapter::CouplingDataUser::setFlipNormal(bool flipNormal)
+{
+    flipNormal_ = flipNormal;
+}
+
+void preciceAdapter::CouplingDataUser::applyFlipNormal(double* dataBuffer, std::size_t size)
```
MakisH: Since this is a new function, let's directly use a container here (e.g., `std::span`).

Interface.C
```diff
@@ -552,6 +558,9 @@ void preciceAdapter::Interface::readCouplingData(double relativeReadTime)
             relativeReadTime,
             {dataBuffer_.data(), nReadData});
```
MakisH: I think this can be simply: `dataBuffer_`. An `std::vector` is accepted as a `std::span` without any explicit conversion.

vidulejs: In this case we should not generally pass the whole vector, because we use the same `dataBuffer_` for scalar and vector data (3x larger than scalar), and therefore we need to specify the size `nReadData` or `nWrittenData`.

MakisH: Could you please open an issue about that? Sounds like a potential source of future trouble. Maybe we need to define our own container as an interface.

CouplingDataUser.C
vidulejs: Resolved.

Interface.C
vidulejs: I don't think it's a problem. We can just provide functions the readable/writable part of the buffer using a span. 

By the way, if you look at `couplingDataReader->read(dataBuffer_.data(), dim_);`, you can see that it already provided the full buffer to all readers, irregardless of if they are scalar or vector readers. But there were no issues, so I guess it's fine. The readers anyway iterate over the length of the patch elements, not the length of the dataBuffer.

Anyway, along the way I thought I can modernize all buffers to use spans instead of pointers, so I opened a new PR for that [#397](https://github.com/precice/openfoam-adapter/pull/397).

MakisH (approved):
All good, and [the system tests pass](https://github.com/precice/openfoam-adapter/actions/runs/22399609761/job/64842290736?pr=395), nice!

PR #396, opened 2026-02-25
Change NULL pointer values to C++11 nullptr

MakisH:
Following an outdated practice that accumulated over the years, our pointers were initialized or checked against `NULL`. Following the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es47-use-nullptr-rather-than-0-or-null), I replaced all of them with `nullptr`.

This pointer literal is [available since C++11](https://en.cppreference.com/w/cpp/language/nullptr.html), so we are fine in terms of support in recent OpenFOAM versions. If an older one does not support it, we should just add a fix in the respective patch.

### DIFF
--- a/Adapter.C
+++ b/Adapter.C
@@ -649,7 +649,7 @@ void preciceAdapter::Adapter::initialize()
 
 void preciceAdapter::Adapter::finalize()
 {
-    if (NULL != precice_ && preciceInitialized_ && !isCouplingOngoing())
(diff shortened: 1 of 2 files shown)


vidulejs (approved):
Straightforward approval. I double checked (searched) whether any `NULL` were left on this branch. Compiles and works. Functionally there should be no changes, except if users had another compiler, like the Intel oneAPI for example (?), in which case this is a compatibility improvement.

PR #397, opened 2026-02-25
Modernize codebase - use span instead of pointer for buffer

vidulejs:
While we're on the topic of modernizing the codebase, I found it not too difficult to replace all instances of `double* buffer` with `precice::span<double> dataBuffer`. Since preCICE implemented/backported its own version of `std::span` (since preCICE version 3.xx?), I've decided to use that instead, because `std:span` was only implemented in C++ 20.

To implement this change I had to import the preCICE header in `CouplingDataUser.H` and change the function signature of `CouplingDataUser::write` and `CouplingDataUser::read`. Since this is a base class, I've had to make changes to all the derived classes in all of the modules. 

TODO list:

### DIFF
--- a/CHT/HeatFlux.C
+++ b/CHT/HeatFlux.C
@@ -18,7 +18,7 @@ preciceAdapter::CHT::HeatFlux::HeatFlux(
     dataType_ = scalar;
 }
 
-std::size_t preciceAdapter::CHT::HeatFlux::write(double* buffer, bool meshConnectivity, const unsigned int dim)
+std::size_t preciceAdapter::CHT::HeatFlux::write(precice::span<double> buffer, bool meshConnectivity, const unsigned int dim)
 {
     int bufferIndex = 0;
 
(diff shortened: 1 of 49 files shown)


MakisH:
[System tests pass](https://github.com/precice/openfoam-adapter/actions/runs/22399588792/job/64842188646?pr=397) in the state before merging #395.

They also pass [after merging](https://github.com/precice/openfoam-adapter/actions/runs/22402394104/job/64852449452?pr=397).

MakisH:
Some first thoughts:

- Generally nice, but the motivation is a bit unclear.
- Impact: Many signatures.
- This changes a bit some calling code in `Interface.C`, but without bringing any additional safety, since we anyway don't bound the `dataBuffer` (and maybe that's a problem). But I guess it could provide some potential in the future (not sure in which direction)
- We are injection `precice::span` everywhere, which is not guaranteed to stay when preCICE itself does not need it anymore. Not sure if we would consider that a breaking change, but it could be removed in a v4 (or earlier).
- Clearer types are anyway nice than raw `double*`. Maybe then introduce our own `DataSpan` as an alias to `precice::span`, and then we can change it easily to an `std::span` in the future? 
- Note that there is a [`stdFoam::span`](https://api.openfoam.com/2506/classstdFoam_1_1span.html) in OpenFOAM.com, which OpenFOAM.org does not define (and I guess foam-extend also not). I would prefer using such an interface type from preCICE than from OpenFOAM.

MakisH:
While this is in first view the right approach and a good way forward, I think the time is not right: We need to touch every reader/writer that other people will need to port to their customized versions, without a clear benefit.

I would close this today, and reconsider it when we remove most of these boilerplate readers/writers.

PR #398, opened 2026-02-26
Implement operation gradient in Generic module

vidulejs:
Implemented `operation gradient;` for volume & surface scalar fields in the new Generic module. The code closely follows `PressureGradientFull.C`. The gradient operation on a scalar field will result in a vector field. Therefore, when `operation == 'gradient'`, the `dataType_` will become vector. The user has to make sure it's also the case in the preCICE config.

For that reason I implemented a new check in `Interface::addCouplingDataWriter`, where it will find if there's a vector/scalar data mismatch between the preCICE config and the adapter.  If the user configured operation gradient, they will get a more specific error:

```log
Error in the preCICE adapter: 
Data dimension mismatch for field "PressureGradient". The field is defined as
scalar data in the preCICE configuration, but the data is vector in the adapter.
Please check your preCICE configuration. If you are trying to couple the gradient
of a scalar field, make sure the data is defined vector in preCICE config.
```

Previously the error would be thrown by preCICE and say that the number of vertices is incorrect.

TODO list:

### DIFF
--- a/Interface.C
+++ b/Interface.C
@@ -460,6 +460,27 @@ void preciceAdapter::Interface::addCouplingDataWriter(
 
     // Add the CouplingDataUser to the list of writers
     couplingDataWriters_.push_back(couplingDataWriter);
+
+    // Check whether config mismatch scalar or vector data
+    unsigned int preciceDataDim = precice_.getDataDimensions(meshName_, fieldConfig.name);
+    bool isVectorDataConfig = (dim_ == preciceDataDim); // 2d or 3d vectors
+    bool isVectorDataAdapter = couplingDataWriter->hasVectorData();
+
+    // Taking gradient of a scalar field results in a vector field
+    bool isScalarConfiguredGradient = (!isVectorDataConfig) && (fieldConfig.operation == "gradient");
+
+    if (isVectorDataConfig != isVectorDataAdapter)
+    {
+        adapterInfo("Data dimension mismatch for field \"" + fieldConfig.name + "\". "
+                        + "The field is defined as " + (isVectorDataConfig ? "vector" : "scalar")
+                        + " data in the preCICE configuration, but the data is "
+                        + (isVectorDataAdapter ? "vector" : "scalar")
+                        + " in the adapter. Please check your preCICE configuration."
+                        + (isScalarConfiguredGradient ? " If you are trying to couple the gradient of a scalar field, \
+                        make sure the data is defined vector in the preCICE config."
+                                                      : ""),
+                    "error");
+    }
 }
 
 
@@ -490,6 +511,21 @@ void preciceAdapter::Interface::addCouplingDataReader(
 
     // Add the CouplingDataUser to the list of readers
     couplingDataReaders_.push_back(couplingDataReader);
+
+    // Check whether config mismatch scalar or vector data
+    unsigned int preciceDataDim = precice_.getDataDimensions(meshName_, fieldConfig.name);
+    bool isVectorDataConfig = (dim_ == preciceDataDim); // 2d or 3d vectors
+    bool isVectorDataAdapter = couplingDataReader->hasVectorData();
+
+    if (isVectorDataConfig != isVectorDataAdapter)
+    {
+        adapterInfo("Data dimension mismatch for field \"" + fieldConfig.name + "\". "
+                        + "The field is defined as " + (isVectorDataConfig ? "vector" : "scalar")
+                        + " data in the preCICE configuration, but the data is "
+                        + (isVectorDataAdapter ? "vector" : "scalar")
+                        + " in the adapter. Please check your preCICE configuration.",
+                    "error");
+    }
 }
 
 void preciceAdapter::Interface::createBuffer()
--- a/docs/config.md
+++ b/docs/config.md
@@ -304,7 +304,7 @@ When reading on the OpenFOAM side, the Generic module automatically detects the 
 
 Limitations:
 
-- The operation `gradient` is currently only supported for scalar fields (resulting to a vector field).
+- The operation `gradient` is currently only supported for writing scalar fields (resulting to a vector field).
 - The operation `surface-normal-gradient` is currently only supported for surface scalar field writing (resulting to a scalar field).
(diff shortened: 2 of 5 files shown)


module-generic/ReadWrite.C
```diff
@@ -30,18 +37,92 @@ void preciceAdapter::Generic::ScalarFieldCoupler::initialize()
...
 std::size_t preciceAdapter::Generic::ScalarFieldCoupler::write(double* buffer, bool meshConnectivity, const unsigned int dim)
 {
     int bufferIndex = 0;
 
+    if (fieldConfig_.operation == "gradient")
+    {
+        // Calculate the full gradient (volVectorField)
+        // Temporary field discarded afterwards
+        Foam::tmp<volVectorField> tmpObject(fvc::grad(*scalarField_));
+        const volVectorField& gradScalarField = tmpObject(); // dereference
```
vidulejs: Unlike `PressureGradientFull`, I don't store the resulting `volVectorField` as a member variable. I'm using `Foam::tmp<volVectorField>` to temporarily allocate it and automatically deallocate it once it's out of scope. Would this be a performance problem - allocating and deallocating memory for each write? Alternatively, we could create a shared buffer to store the scalar field gradients if configured.

module-generic/ReadWrite.C
```diff
@@ -125,6 +206,11 @@ void preciceAdapter::Generic::ScalarFieldCoupler::read(double* buffer, const uns
 {
     int bufferIndex = 0;
 
+    if (fieldConfig_.operation == "gradient" || fieldConfig_.operation == "surface-normal-gradient")
+    {
+        adapterInfo("Generic module: The gradient operation is only supported for writing.", "error");
+    }
+
```
vidulejs, line 222: Should we throw this error when reading a field with operation defined gradient? I feel like this behavior is undefined, therefore I would throw an error.

Interface.C
```diff
@@ -460,6 +460,27 @@ void preciceAdapter::Interface::addCouplingDataWriter(
...
+        adapterInfo("Data dimension mismatch for field \"" + fieldConfig.name + "\". "
+                        + "The field is defined as " + (isVectorDataConfig ? "vector" : "scalar")
+                        + " data in the preCICE configuration, but the data is "
+                        + (isVectorDataAdapter ? "vector" : "scalar")
+                        + " in the adapter. Please check your preCICE configuration."
+                        + (isScalarConfiguredGradient ? " If you are trying to couple the gradient of a scalar field, \
+                        make sure the data is defined vector in the preCICE config."
+                                                      : ""),
+                    "error");
+    }
```
vidulejs, line 483: New check and error message. Variable names might not be the best.

module-generic/ReadWrite.C
MakisH: Discussed in a meeting: This should indeed have performance implications, better allocate in the constructor and reuse.

Interface.C
MakisH: Indeed not the clearest variable names, but it also doesn't really matter, as they are only locally used.

MakisH (commented):
Looks good overall, I have not tested it.

vidulejs:
Finally got around to implementing this. Now we also allocate memory for the gradient of the scalar field (only if the gradient operation is specified). However, I don't think this solves the problem we thought it does. The `Foam::tmp<T>` is just a smart pointer, and OpenFOAM will still do the memory allocation under the hood. `fvc::grad` will not do the calculation in-place if I understand correctly. Maybe this version of the code is cleaner, but I don't think it makes a difference performance wise. 

The only difference is that we *could* now reuse the gradient field multiple times per time step if there are multiple writes. *Could* in italics, because right now it will re-calculate it on each write regardless. To avoid that, we'd have to check if we already calculated it for that timestep in that time window. Anyway, I think this is very niche and not worth looking into.

Ready to merge. Didn't test but it's functionally the same as last version.

vidulejs:
I just realized that allocating memory for a fixed size field is more dangerous than using a `tmp` field, because then we don't support changing mesh topology. I.e., the mesh is refined or otherwise dynamic. The same problem will occur for `PressureGradientFull.C`. This might be an existing problem in other parts of the codebase too. I might open an issue for this.

docs/config.md
```diff
@@ -302,7 +302,11 @@ Explicitly setting `locations volumeCenters;` in preciceDict is required for vol
 
 When reading on the OpenFOAM side, the Generic module automatically detects the boundary condition type of the coupled patch to apply the received data. The boundary condition must be respected, therefore the data received is applied either as a `fixedValue` or `fixedGradient` boundary condition, determined by the type set in the `0/` files.
 
-The `operation surface-normal-gradient;` is currently only supported for scalar field surface writing. The operation `gradient` as well as the option `flip-normal` are not yet supported by the adapter.
+Limitations:
+
+- The operation `gradient` is currently only supported for scalar fields (resulting to a vector field).
```
```suggestion
- The operation `gradient` is currently only supported for writing scalar fields (resulting to a vector field).
```

Following https://github.com/precice/openfoam-adapter/pull/398#discussion_r2864205973

MakisH (commented):
Thanks for updating!

> The only difference is that we _could_ now reuse the gradient field multiple times per time step if there are multiple writes. _Could_ in italics, because right now it will re-calculate it on each write regardless. To avoid that, we'd have to check if we already calculated it for that timestep in that time window. Anyway, I think this is very niche and not worth looking into.
> 
> Ready to merge. Didn't test but it's functionally the same as last version.

I would say let's bring the feature to the users and get feedback about it. We currently only have very limited use cases, and we could look into such aspects once we have more examples.

> I just realized that allocating memory for a fixed size field is more dangerous than using a `tmp` field, because then we don't support changing mesh topology. I.e., the mesh is refined or otherwise dynamic. The same problem will occur for `PressureGradientFull.C`. This might be an existing problem in other parts of the codebase too. I might open an issue for this.

I took a note in https://github.com/precice/openfoam-adapter/issues/382#issuecomment-4367044594

module-generic/ReadWrite.C
MakisH: See further discussion in https://github.com/precice/openfoam-adapter/pull/398#issuecomment-4151028002 and below.

MakisH (approved):

PR #400, opened 2026-03-08
Implement mixed boundary condition (Robin) in Generic module

vidulejs:
Implemented mixed boundary condition (Robin) in Generic module. This allows to replicate, for example, CHT cases where there are `Sink-Temperature` and `Heat-Transfer-Coefficient` coupling data. I've been able to validate the implementation on the tutorial `heat-exchanger` (pictures to follow).

There are new `operation` options available to set the mixed boundary conditions when reading data from preCICE:
- `ref-value` 
- `ref-gradient`
- `value-fraction`,
which will set the respective fields in the [OpenFOAM mixed BC](https://www.openfoam.com/documentation/guides/latest/doc/guide-bcs-mixed.html).

To test the implementation standalone from the CHT module, I created a new tutorial case `partitioned-heat-transfer-robin` on my [fork of the tutorials](https://github.com/vidulejs/tutorials/tree/robin-robin/partitioned-heat-conduction-robin). 

Here's the example preciceDict configuration for that case:
```preciceDict
participant Left;

modules (generic);

interfaces
{
  Interface1
  {
    mesh              Left-Mesh;
    patches           (interface);
    
    readData
    (
      Sink-Temp
      {
        name        Temperature-Right;
        solver_name T;
        operation   ref-value;
      }
      HTC
      {
        name        Gradient-Right;
        solver_name T;
        operation   ref-gradient;
        flip-normal true;
      }
    );
    
    writeData
    (
      Write-Temp
      {
        name        Temperature-Left;
        solver_name T;
        operation   value;
      }
      Write-Grad
      {
        name        Gradient-Left;
        solver_name T;
        operation   surface-normal-gradient;
      }
    );
  };
};
```

TODO list:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -302,7 +302,7 @@ Explicitly setting `locations volumeCenters;` in preciceDict is required for vol
 
 When reading on the OpenFOAM side, the Generic module automatically detects the boundary condition type of the coupled patch to apply the received data. The boundary condition must be respected, therefore the data received is applied either as a `fixedValue` or `fixedGradient` boundary condition, determined by the type set in the `0/` files.
 
-Additionally, the Robin or mixed boundary condition is now supported for surface coupling. The OpenFOAM boundary type `mixed` must be set for the coupled patch in the `0/` files like so:
+Additionally, the Robin or mixed boundary condition is now supported for scalar surface coupling. The OpenFOAM boundary type `mixed` or `mixedCoded` must be set for the coupled patch in the `0/` files like so:
 
 ```cpp
 // File 0/T
@@ -318,7 +318,11 @@ boundaryField
 }
 ```
 
-Any of the three fields `refValue`, `refGradient`, and `valueFraction` can be written or read using the adapter (the initial values will be overwritten by the adapter). To do so, specify the `operation` for each coupling data in `preciceDict` as `ref-value`, `ref-gradient`, or `value-fraction`. Be careful when using the `value-fraction` operation, as it is not a physical quantity and must be within the range of (0, 1). Please note that if you intend to use the `mixed` boundary condition for a heat transfer problem, the `valueFraction` is not the same as the heat transfer coefficient and it's not possible to explicitly set the heat transfer coefficient as a parameter.
+Any of the three fields `refValue`, `refGradient`, and `valueFraction` can be read from another participant (the initial values will be overwritten by the adapter). To do so, specify the `operation` for each coupling data in `preciceDict` as `ref-value`, `ref-gradient`, or `value-fraction`. Be careful when using the `value-fraction` operation, as it is not a physical quantity and must be within the range of [0, 1].
+
+These three operations are only supported for the read side. To write coupling data from a mixed boundary patch, just use the standard `value` or `surface-normal-gradient` operation. The `valueFraction` field of the mixed boundary condition is not exposed by the adapter for writing.
+
+Please note that if you intend to use the `mixed` boundary condition for a heat transfer problem, the `valueFraction` is not the same as the heat transfer coefficient and it's not possible to explicitly set the heat transfer coefficient as a parameter.
 
 Limitations:
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -318,7 +318,7 @@ boundaryField
 }
 ```
 
-Any of the three fields `refValue`, `refGradient`, and `valueFraction` can be written or read using the adapter (the initial values will be overwritten by the adapter in case of write). To do so, specify the `operation` for each coupling data in `preciceDict` as `ref-value`, `ref-gradient`, or `value-fraction`. Be careful when using the `value-fraction` operation, as it is not a physical quantity and must be within the range of (0, 1). Please note that if you intend to use the `mixed` boundary condition for a heat transfer problem, the `valueFraction` is not the same as the heat transfer coefficient and it's not possible to explictly set the heat transfer coefficient as a parameter.
+Any of the three fields `refValue`, `refGradient`, and `valueFraction` can be written or read using the adapter (the initial values will be overwritten by the adapter). To do so, specify the `operation` for each coupling data in `preciceDict` as `ref-value`, `ref-gradient`, or `value-fraction`. Be careful when using the `value-fraction` operation, as it is not a physical quantity and must be within the range of (0, 1). Please note that if you intend to use the `mixed` boundary condition for a heat transfer problem, the `valueFraction` is not the same as the heat transfer coefficient and it's not possible to explicitly set the heat transfer coefficient as a parameter.
 
 Limitations:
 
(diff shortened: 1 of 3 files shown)


module-generic/ReadWrite.C
MakisH, line 280: We could add similar checks and error messages here: if not a supported operation for this BC type, throw an error.

module-generic/ReadWrite.C
```diff
@@ -173,14 +182,35 @@ void preciceAdapter::Generic::ScalarFieldCoupler::read(double* buffer, const uns
                 boundaryPatch.gradient()[i] = buffer[bufferIndex++];
             }
         }
+        else if (isA<mixedFvPatchScalarField>(bc))
+        {
+            auto& boundaryPatch = refCast<mixedFvPatchScalarField>(bc);
+            if (fieldConfig_.operation == "ref-value" || fieldConfig_.operation == "value")
```
MakisH: Maybe add a comment for the motivation for both checks here.

module-generic/ReadWrite.C
MakisH, line 1: Needs a changelog entry as well.

MakisH (commented):
Looks good overall, some minor suggestions.

From our discussion: The [robin test case](https://github.com/vidulejs/tutorials/tree/robin-robin/partitioned-heat-conduction-robin) would also be nice to have as a tutorial, with the new schema, to use as a test case.

Reply by @vidulejs: We don't have the complete support yet, as we are missing the conductivity calculation. In this case, it happens to work because the coefficient is 1. In that sense, adding this tutorial wouldbe misleading at the moment.

MakisH:
The merge conflict in `ReadWrite.C` originates from merging #398 and should be easy to resolve.

vidulejs:
Resolved merge conflict and wrote a little bit of documentation. Please check if the documentation looks alright to you and suggest and / or commit any changes.

module-generic/ReadWrite.C
vidulejs: Right, we should throw an error if the operation is not compatible with the BC type.  This would be only a few lines of code, but I can't finish it tonight. I will come back to this.

docs/config.md
```diff
@@ -302,6 +302,24 @@ Explicitly setting `locations volumeCenters;` in preciceDict is required for vol
...
+    {
+        type            mixed;
+        refValue        uniform 1;
+        refGradient     uniform 1;
+        valueFraction   uniform 0.5;
+    }
+}
+```
+
+Any of the three fields `refValue`, `refGradient`, and `valueFraction` can be written or read using the adapter (the initial values will be overwritten by the adapter in case of write). To do so, specify the `operation` for each coupling data in `preciceDict` as `ref-value`, `ref-gradient`, or `value-fraction`. Be careful when using the `value-fraction` operation, as it is not a physical quantity and must be within the range of (0, 1). Please note that if you intend to use the `mixed` boundary condition for a heat transfer problem, the `valueFraction` is not the same as the heat transfer coefficient and it's not possible to explictly set the heat transfer coefficient as a parameter.
```
MakisH: The overwriting is in case of `read`, no? In any case, this is something we already have in all boundary conditions.

```suggestion
Any of the three fields `refValue`, `refGradient`, and `valueFraction` can be written or read using the adapter (the initial values will be overwritten by the adapter). To do so, specify the `operation` for each coupling data in `preciceDict` as `ref-value`, `ref-gradient`, or `value-fraction`. Be careful when using the `value-fraction` operation, as it is not a physical quantity and must be within the range of (0, 1). Please note that if you intend to use the `mixed` boundary condition for a heat transfer problem, the `valueFraction` is not the same as the heat transfer coefficient and it's not possible to explictly set the heat transfer coefficient as a parameter.
```

module-generic/ReadWrite.C
```diff
@@ -47,6 +48,14 @@ void preciceAdapter::Generic::ScalarFieldCoupler::initialize()
             adapterInfo("Generic module: The surface-normal-gradient operation is only supported for faceCenters location type.", "error");
         }
     }
+
+    if (fieldConfig_.operation == "ref-value" || fieldConfig_.operation == "ref-gradient" || fieldConfig_.operation == "value-fraction")
```
MakisH, line 52: If we are to accept both `ref-value` and `value` below, we should also do it here and everywhere else. But while it would be nice to have such an alias, I think it can also be a source of confusion and bugs. I would rather only support `ref-value`. This would also simplify such error checks.

(we have a similar support for `face-center` and `face-centre`, which I think has not really helped so far)

MakisH (commented):
Thanks for the update! Looking forward to merging this once ready.

module-generic/ReadWrite.C
vidulejs: Removed.

module-generic/ReadWrite.C
vidulejs: Added the error messages!

vidulejs:
Hey! 

I looked at this PR again and was at first confused about something, namely the lack of write side changes. I went off on a tangent and tried writing `mixedBoundaryPatch.refValue()` and `.refGrad()` only to find out that those are the static values set in the boundary condition ( simply the values from the `0/` files), i.e., they are not solved for and it doesn't make sense to write them. I of course knew this in the past when I wrote this pull request, but I forgot since then :)

Instead we should write `value` or `surface-normal-gradient`, just like I already did in my test case on my fork of the tutorials. Therefore, to avoid this confusion in the future, I added a clarifying sentence to the docs and an error in case the user would try writing `ref-gradient`, `ref-value` or `value-fraction`. 

Additionally, I tested these changes once more with a new tutorial case [partitioned-heat-conduction-mixedbc](https://github.com/vidulejs/tutorials/tree/robin-robin/partitioned-heat-conduction-mixedbc), where I set the `valueFraction` field of the mixed BC to 0 on one side and 1 on the other, which reduces the mixed BC case to simply Neumann-Dirichlet. It works! I get good error compared to analytical, so it should be good to go.

```log
Error metrics at t = 1s:
l2 error (sqrt(sum(err^2)/n)):		6.27862e-05
Maximum absolute error (max(err)):	7.49135e-05
Global absolute error (sum(err^2)):	3.94211e-05
```
image

I think this PR is complete. Take a look, only thing is maybe you'll want to modify the wording in the docs.

MakisH (approved):
Thanks a lot for the updates! This looks indeed ready to merge.

> Additionally, I tested these changes once more with a new tutorial case [partitioned-heat-conduction-mixedbc](https://github.com/vidulejs/tutorials/tree/robin-robin/partitioned-heat-conduction-mixedbc), where I set the `valueFraction` field of the mixed BC to 0 on one side and 1 on the other, which reduces the mixed BC case to simply Neumann-Dirichlet. It works! I get good error compared to analytical, so it should be good to go.

Could you please quickly open a PR for that tutorial? I can then directly modify it and add it to the system tests. This will be quite useful as a test for this feature.

PR #405, opened 2026-06-05
Add support for implicit coupling of Lagrangian particles

mike-tree-corvid:
Added checkpointing for Lagrangian particles within the preCICE adaptor. When a time-rewind is detected the adapter clears the kinematicCloud state from the mesh at the end of the time window and restores the kinematicCloud state from the beginning of the time window to the mesh.

However, the preCICE adapter is a function object that operates at the global mesh level and cannot access the protected memory accumulators within OpenFOAM's native InjectionModel that govern if and how many particles are injected. So, in order to completely reset the sim and properly adjust active particle injection an additional library was created. The library, implicitPatchInjection, was created to gain protected access to these variables and override the native prepareForNextTimeStep method allowing for safely wiping the injector's memory and state caches to reset injection. To employ this properly the user will need to load the implicitPatchInjection library in system/controlDict and then swap their injectionModel type to "implicitPatchInjection" in constant/kinematicCloudProperties

This addition only supports basicthermoCloud types and patchInjection -> implicitPatchInjection injection types. Other cloud types not supported include: kinematicCloud, reactingCloud, reactingMultiphaseCloud, collidingCloud, MPPICCloud. Other injection models that cannot be replaced with implicitPatchInjection include: coneInjection, solidConeInjection, cellZoneInjection, manualInjection, surfaceFilmInjection, and fieldActivatedInjection.

TODO list:


implicitPatchInjection/Make/linux64GccDPInt32Opt/makeImplicitPatchInjection.C.dep
MakisH, line 1: This folder sneaked into the branch. We could define such common patterns in `.gitignore`.

implicitPatchInjection/Make/options
MakisH, line 1: Do you have an impression at the moment of which versions of OpenFOAM this currently builds with? We currently document support of the adapter back to v1812: https://precice.org/adapter-openfoam-support.html

We could raise that to v1912 (Ubuntu 22.04) or v2212 (Ubuntu 24.04) soon.

We would later also need to port to the latest Foundation version (but we are currently behind on that matter).

implicitPatchInjection/implicitPatchInjection.H
```diff
@@ -0,0 +1,242 @@
...
+  \\      /  F ield         | OpenFOAM: The Open Source CFD Toolbox
+   \\    /   O peration     |
+    \\  /    A nd           | www.openfoam.com
+     \\/     M anipulation  |
+-------------------------------------------------------------------------------
+    Copyright (C) 2011-2017 OpenFOAM Foundation
+    Copyright (C) 2021 OpenCFD Ltd.
+-------------------------------------------------------------------------------
+License
+    This file is part of OpenFOAM.
```
MakisH: Is it, or is it your code? We would need some history (e.g., based on ... from version ..., with extensions ...).
That would also make reviewing easier.

MakisH (commented):
Thank you for the PR! I know this is still a draft, but you might be interested in some comments already.

How could I reproduce the original issue and verify the fix? Could we also add the respective case as a tutorial, so that we can integrate it into the system tests?

Happy to integrate this in a next release. We are very close to releasing v1.4.0, so I would rather postpone it. But we could make another release for this feature once it is integrated.

Note that you can use pre-commit to format the code: https://github.com/precice/openfoam-adapter/blob/develop/CONTRIBUTING.md#code-formatting

mike-tree-corvid:
> Thank you for the PR! I know this is still a draft, but you might be interested in some comments already.
> 
> How could I reproduce the original issue and verify the fix? Could we also add the respective case as a tutorial, so that we can integrate it into the system tests?
> 
> Happy to integrate this in a next release. We are very close to releasing v1.4.0, so I would rather postpone it. But we could make another release for this feature once it is integrated.
> 
> Note that you can use pre-commit to format the code: https://github.com/precice/openfoam-adapter/blob/develop/CONTRIBUTING.md#code-formatting

The sim I'm currently running this on is coupling two 60 million cell meshes and uses an in-house solver. I don't think it's something that will work as a tutorial. It appears that another preCICE user may have a case that is more suitable for tutorial use:

https://precice.discourse.group/t/does-openfoam-adapter-support-lagrangian-particle-checkpointing/2947/5?u=mike_tree

If the adapted perpendicular flap case isn't shareable for any reason, I'll see if I can drum up a much smaller and completely open test case that we can use for a tutorial and continuous code verification.

implicitPatchInjection/implicitPatchInjection.H
mike-tree-corvid: You're right. It is my code, so i added a copyright to my company and changed the .H file description to no longer be a direct copy of OpenFOAM's native PatchInjection.H

implicitPatchInjection/Make/options
mike-tree-corvid: I'm using v2412, so I know it works with that. I've never ported anything to the Foundation version, but I know the differences between the two code bases has grown recently. I hope this doesn't prove too much of a struggle.

implicitPatchInjection/Make/linux64GccDPInt32Opt/makeImplicitPatchInjection.C.dep
mike-tree-corvid: added to .gitignore

mike-tree-corvid:
I added a draft tutorial pull request to the precice/tutorials repo here:

https://github.com/precice/tutorials/pull/834

Allwmake
```diff
@@ -13,6 +13,16 @@ ADAPTER_CFLAGS="${PRECICE_OPENFOAM_CFLAGS:-}"
 # e.g., add "export WM_NCOMPPROCS=4" to your ~/.bashrc
 # Make sure that these options are supported by your OpenFOAM version.
 adapter_build_command(){
+#    log "Building custom Lagrangian particle patch injector..."
+#    wmake libso implicitPatchInjection/
+#
+#    log "Building custom Lagrangian particle cone nozzle injector..."
+#    wmake libso implicitConeNozzleInjection/
+
```
MakisH: Why keep these commented-out lines here, if the directories are not provided?

Adapter.C
```diff
@@ -887,6 +887,48 @@ void preciceAdapter::Adapter::setupCheckpointing()
...
+    {
+        backupKinematicParcelsLists_.resize(activeKinematicClouds_.size());
+        backupThermoParcelsLists_.resize(activeThermoClouds_.size());
+        backupReactingParcelsLists_.resize(activeReactingClouds_.size());
+        backupReactingMultiphaseParcelsLists_.resize(activeReactingMultiphaseClouds_.size());
+        backupKinematicCollidingParcelsLists_.resize(activeKinematicCollidingClouds_.size());
+    }
+
     // Add fields in the checkpointing list - sorted for parallel consistency
     DEBUG(adapterInfo("Adding in checkpointed fields..."));
```
MakisH, line 936: It would make sense to have this or a similar output message before the first `setupCloudType` call.

.gitignore
```diff
@@ -38,6 +38,8 @@
 lnInclude/
 Make/linux64GccDPInt32Opt/
 Make/darwin64ClangDPInt32Opt/
+*/Make/linux64GccDPInt32Opt/
+*/Make/darwin64ClangDPInt32Opt/
```
MakisH: This should be a more general rule:


```suggestion
**/Make/linux64GccDPInt32Opt/
**/Make/darwin64ClangDPInt32Opt/
```

Make/options
```diff
@@ -22,5 +35,15 @@ LIB_LIBS = \
...
+    -llagrangian \
+    -llagrangianIntermediate \
+    -lspecie \
+    -lSLGThermo \
+    -lregionModels \
+    -lsurfaceFilmModels \
+    -lreactionThermophysicalModels \
+    -lthermophysicalProperties \
+    -lregionFaModels \
+    -lfiniteArea \
```
MakisH: I am afraid I know the answer, but: Are all of these includes and linking options really needed in the main `Make/options`, or some of them only in the `implicitInjectionWrapper/Make/options`?

If we could modularize this feature (ideally with macros to enable/disable it) and make the dependencies optional, I would feel a bit more confident that we can maintain it in the long term and for different versions.

implicitInjectionWrapper/Make/files
MakisH, line 1: Since this is a complete library, you could also add a very short `README.md` giving some context: what is the purpose, what is the entry point, what are the different files for.

implicitInjectionWrapper/implicitInjectionWrapper.H
```diff
@@ -0,0 +1,196 @@
...
+    Wrapper template class for overriding the prepareForNextTimeStep function
+    from the InjectionModel class for every InjectionModel type.
+
+    This creates "implicit" versions of the native OpenFOAM InjectionModel
+    types (implicitPatchInjection, implicitConeNozzleInjection,
+    implicitManualInjection, etc.) that support time-rewind from preCICE
+    implicit coupling.
+
+    Derived from native InjectionModels (v2412) to support Eulerian-Lagrangian 
+    implicit coupling. Explicitly handles memory lobotomization,
```
MakisH: Is this a technical term? 👀

implicitInjectionWrapper/implicitInjectionWrapper.H
```diff
@@ -0,0 +1,196 @@
...
+
+    Wrapper template class for overriding the prepareForNextTimeStep function
+    from the InjectionModel class for every InjectionModel type.
+
+    This creates "implicit" versions of the native OpenFOAM InjectionModel
+    types (implicitPatchInjection, implicitConeNozzleInjection,
+    implicitManualInjection, etc.) that support time-rewind from preCICE
+    implicit coupling.
+
+    Derived from native InjectionModels (v2412) to support Eulerian-Lagrangian
```
MakisH, line 44: Could you please add a link to the exact file on GitLab?

https://gitlab.com/openfoam/core/openfoam/-/tree/OpenFOAM-v2412

implicitInjectionWrapper/implicitInjectionWrapper.H
```diff
@@ -0,0 +1,196 @@
...
+#define implicitInjectionWrapper_H
+
+#include "InjectionModel.H"
+#include <vector>
+
+// * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * //
+
+namespace Foam
+{
+#define DEFINE_IMPLICIT_INJECTOR(WrapperName, BaseInjector)                       \
```
MakisH: Could this long and complicated macro be a template that takes a template as an argument? Or does this pattern originate from the original file?

I understand why this might be a convenient solution, and I have no strong objection if there are good reasons for it, but I have to challenge such an overuse of macros.

Utilities.C
MakisH, line 1: Unrelated change

MakisH (commented):
Thank you for all the updates! I am slowly catching up with the context. Having two motivated users needing this feature is already quite a good reason to integrate this feature.

I am still just thinking how we could make it easier to disable it, in case we cannot maintain it or port it to different versions in the future, since it is adding quite some dependencies.

[All system tests but one pass](https://github.com/precice/openfoam-adapter/actions/runs/29340801559?pr=405): The [elastic-tube-3d (OpenFOAM-FEniCS)](https://github.com/precice/tutorials/tree/develop/elastic-tube-3d) reports some regressions in the first iterations, but this seems to be a known unrelated issue (https://github.com/precice/tutorials/issues/393). The OpenFOAM-CalculiX case of the same tutorial passes.

Allwmake
mike-tree-corvid: these were left over from before refactoring to build a single library for all implicit injectors. I removed these lines.

Adapter.C
mike-tree-corvid: Added a similar output message for checkpointing clouds

.gitignore
mike-tree-corvid: made this .gitignore rule more general, as suggested

Make/options
mike-tree-corvid: Yes, unfortunately they are all necessary. Lagrangian particle considerations touch a lot. I'll see if I can find an easy way to enable / disable these additions...

implicitInjectionWrapper/implicitInjectionWrapper.H
mike-tree-corvid: not technical in this context, no. I removed it in place of a different, simpler term.

implicitInjectionWrapper/implicitInjectionWrapper.H
mike-tree-corvid: link added

implicitInjectionWrapper/implicitInjectionWrapper.H
mike-tree-corvid: Good catch. I reformatted this to a template-template parameter class, like you suggested. I did keep a small macro for boilerplate creation because OpenFOAM's run-time selection mechanism requires TypeName string declaration to match the class name for dictionary look-ups.

implicitInjectionWrapper/Make/files
mike-tree-corvid: added a README.md for the implicitInjectionWrapper

Make/options
mike-tree-corvid: I used #ifdef statements to modularize the Lagrangian particle support within the Adapter.C and Adapter.H files. I then added variables so the Make/options files is dynamically modified. Allwmake is updated to look for the presence of an "ADAPTER_ENABLE_LAGRANGIAN" environment variable. If ADAPTER_ENABLE_LAGRANGIAN=1 then -DADAPTER_ENABLE_LAGRANGIAN is added to ADAPTER_CFLAGS and the implicitInjectionWrapper code will be built. If ADAPTER_ENABLE_LAGRANGIAN does not exist or is equal to something other than 1, then neither of these things happen.

implicitInjectionWrapper/README.md
```diff
@@ -0,0 +1,30 @@
...
+makeImplicitInjectorsKinematicColliding.C - uses macro defined in implicitInjectionWrapper.H to create wrapped template classes for implicitPatchInjection, implicitConeNozzleInjection, and implicitManualInjection types registered to the basicKinematicCollidingCloud type
+makeImplicitInjectorsReacting.C - uses macro defined in implicitInjectionWrapper.H to create wrapped template classes for implicitPatchInjection, implicitConeNozzleInjection, and implicitManualInjection types registered to the basicReactingCloud type
+makeImplicitInjectorsReactingMultiphase.C - uses macro defined in implicitInjectionWrapper.H to create wrapped template classes for implicitPatchInjection, implicitConeNozzleInjection, and implicitManualInjection types registered to the basicReactingMultiphaseCloud type
+makeImplicitInjectorsThermo.C - uses macro defined in implicitInjectionWrapper.H to create wrapped template classes for implicitPatchInjection, implicitConeNozzleInjection, and implicitManualInjection types registered to the basicThermoCloud type
```
MakisH: Markdown formatting:

```suggestion
- `implicitInjectionWrapper.H` - a wrapper template class used to override the `prepareForNextTimeStep` function from the native OpenFOAM `InjectionModel` class for every `InjectionModel` type. This creates "implicit" versions of the native OpenFOAM InjectionModel types (`implicitPatchInjection`, `implicitConeNozzleInjection`, `implicitManualInjection`, etc.) that support checkpoint logic from preCICE implicit coupling.
- `makeImplicitInjectorsKinematic.C` - uses macro defined in `implicitInjectionWrapper.H` to create wrapped template classes for `implicitPatchInjection`, `implicitConeNozzleInjection`, and `implicitManualInjection` types registered to the `basicKinematicCloud` type
- `makeImplicitInjectorsKinematicColliding.C` - uses macro defined in `implicitInjectionWrapper.H` to create wrapped template classes for `implicitPatchInjection`, `implicitConeNozzleInjection`, and `implicitManualInjection` types registered to the `basicKinematicCollidingCloud` type
- `makeImplicitInjectorsReacting.C` - uses macro defined in `implicitInjectionWrapper.H` to create wrapped template classes for `implicitPatchInjection`, `implicitConeNozzleInjection`, and `implicitManualInjection` types registered to the `basicReactingCloud` type
- `makeImplicitInjectorsReactingMultiphase.C` - uses macro defined in `implicitInjectionWrapper.H` to create wrapped template classes for `implicitPatchInjection`, `implicitConeNozzleInjection`, and `implicitManualInjection` types registered to the `basicReactingMultiphaseCloud` type
- `makeImplicitInjectorsThermo.C` - uses macro defined in `implicitInjectionWrapper.H` to create wrapped template classes for `implicitPatchInjection`, `implicitConeNozzleInjection`, and `implicitManualInjection` types registered to the `basicThermoCloud` type
```

implicitInjectionWrapper/README.md
```diff
@@ -0,0 +1,29 @@
...
+makeImplicitInjectorsKinematicColliding.C - uses macro defined in implicitInjectionWrapper.H to create wrapped template classes for implicitPatchInjection, implicitConeNozzleInjection, implicitCellZoneInjection, and implicitManualInjection types registered to the basicKinematicCollidingCloud type
+makeImplicitInjectorsReacting.C - uses macro defined in implicitInjectionWrapper.H to create wrapped template classes for implicitPatchInjection, implicitConeNozzleInjection, implicitCellZoneInjection, and implicitManualInjection types registered to the basicReactingCloud type
+makeImplicitInjectorsReactingMultiphase.C - uses macro defined in implicitInjectionWrapper.H to create wrapped template classes for implicitPatchInjection, implicitConeNozzleInjection, implicitCellZoneInjection, and implicitManualInjection types registered to the basicReactingMultiphaseCloud type
+makeImplicitInjectorsThermo.C - uses macro defined in implicitInjectionWrapper.H to create wrapped template classes for implicitPatchInjection, implicitConeNozzleInjection, implicitCellZoneInjection, and implicitManualInjection types registered to the basicThermoCloud type
```
MakisH, line 11: Why not using Markdown formatting here? I converted this to a list with inline code elements in https://github.com/precice/openfoam-adapter/pull/405/changes/ac3467380e01e2fce761820f16a2d3a3fe6e8f0e, but it was reverted in https://github.com/precice/openfoam-adapter/pull/405/changes/e7e6c6ef3be8d687775480bc95f17bedd0c7f801. I can apply that patch again.

mike-tree-corvid: my mistake, I didn't intend to revert your changes. Yes, please apply your patch again.

PR #412, opened 2026-06-29
Prepare documentation for rendering in Hugo (native)

MuhammadAashirAslam:
Preserve the existing openfoam-adapter documentation URL for the Hugo migration
TODO list:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,10 +1,10 @@
 ---
 title: The OpenFOAM adapter
 permalink: adapter-openfoam-overview.html
-url: /adapter-openfoam-overview.html
-redirect_from: adapter-openfoam.html
 aliases:
+  - /adapter-openfoam-overview.html
   - /adapter-openfoam.html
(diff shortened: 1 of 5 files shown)


MakisH:
@MuhammadAashirAslam could you please explain the main idea and why this is different that the approach you originally had in https://github.com/precice/code_aster-adapter/pull/31? Shouldn't the `alias` include the sections in the end?

MuhammadAashirAslam:
We dropped the Hugo URL field and now use the old Jekyll URLs as aliases pointing to Hugo’s canonical section URLs, e.g. /docs/adapters/openfoam/.

> Shouldn't the `alias` include the sections in the end

Sections are made by hugo.toml during the module mount mapping (https://github.com/MuhammadAashirAslam/precice.github.io/blob/hugo-test/hugo.toml#L95C1-L98C52)

MakisH (approved):

PR #417, opened 2026-08-28
Support reading Stress in solid participants

### LINKED ISSUE #240: Reading stress from solid solvers
https://github.com/precice/openfoam-adapter/pull/236 added the possibility to read forces, effectively enabling structure solvers (such as solidDisplacementFoam or solids4Foam) to use the adapter.

It would still be nice if the adapter could also read stresses (`Stress`). This should hopefully not be so complicated to implement, assuming it is relatively similar to the writing part.

philipcardiff:
## Summary

Closes #240.

`ForceBase::readFromBuffer()` was a stub that raised an adapter error (*"Reading stresses is not supported."*), so a `solverType solid` participant could not read `Stress` — even though `docs/config.md` already lists `Stress` as read data for solid participants.

This PR:

* implements `ForceBase::readFromBuffer(buffer, field, dim)`, mirroring the existing `Force::read()` (patch loop, `faceCenters` only, `notImplemented` for `faceNodes`), and returning the buffer index so it is symmetric with `writeToBuffer`;
* gives `Stress` the same **bind-or-create** field handling that `Force` already has, so the buffer can be written into a field owned by the solid solver rather than into a shadow field. The raw `new`/`delete` and the explicit destructor are replaced by a `std::unique_ptr`, matching `Force::ForceOwning_`;
* adds a `nameStress` entry to the `FSI` sub-dictionary of `preciceDict` (default `Stress`), alongside the existing `nameForce`.

No refactor of the `ForceBase`/`Force`/`Stress` hierarchy — the in-code TODO suggesting that is left for a separate change, to keep this diff reviewable.

For solids4foam, this lets a `solidTraction` boundary condition consume the coupled traction directly through its `tractionField` option:

```c++
// 0/D
interface
{
    type            solidTraction;
    tractionField   solidTraction;
    pressure        uniform 0;
    value           uniform (0 0 0);
}
```

```c++
// system/preciceDict
readData ( Stress );

FSI
{
    solverType    solid;
    nameStress    solidTraction;
}
```

## Validation

Tested with OpenFOAM v2606, preCICE 3.4.1, solids4foam (development).

`perpendicular-flap-stress` is the same physical scenario as `perpendicular-flap` — the two `fluid-openfoam/system/blockMeshDict` files are byte-identical, and only the exchanged data differs (`Stress` with consistent mapping vs `Force` with conservative mapping). A solids4foam solid participant was added to `perpendicular-flap-stress` and run for the full 5 s (500 time windows), then compared against the existing `perpendicular-flap/solid-solids4foam` case:

| t [s] | Force path | Stress path |
|---|---|---|
| 0.10 | +1.836160e-02 | +1.837240e-02 |
| 0.50 | +1.598337e-01 | +1.593658e-01 |
| 2.00 | +1.562350e-01 | +1.564509e-01 |
| 5.00 | +1.392887e-01 | +1.383371e-01 |

Peak flap-tip amplitude over the run: **0.166294 (Force) vs 0.165038 (Stress), i.e. 0.76%**. The residual difference is the expected consequence of the different mapping constraints, not of the read path. Both runs report *"The momentum equation converged in all time-steps"*.

Regression check: `perpendicular-flap` (Force/Displacement) and `breaking-dam-2d` are unchanged by this PR — the `Force` and `Displacement` code paths are untouched.

A follow-up tutorials PR adds the `perpendicular-flap-stress/solid-solids4foam` case used here.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SBFTMffUN7qDigrDQrPVb2

### DIFF
--- a/FSI/FSI.C
+++ b/FSI/FSI.C
@@ -86,6 +86,10 @@ bool preciceAdapter::FSI::FluidStructureInteraction::readConfig(const IOdictiona
     nameForce_ = FSIdict.lookupOrDefault<word>("nameForce", "Force");
     DEBUG(adapterInfo("    force field name : " + nameForce_));
 
+    // Read the name of the stress field (if different)
+    nameStress_ = FSIdict.lookupOrDefault<word>("nameStress", "Stress");
+    DEBUG(adapterInfo("    stress field name : " + nameStress_));
+
     return true;
 }
 
@@ -162,7 +166,7 @@ bool preciceAdapter::FSI::FluidStructureInteraction::addWriters(const preciceAda
     {
         interface->addCouplingDataWriter(
             fieldConfig,
-            new Stress(mesh_, solverType_) /* TODO: Add any other arguments here */
+            new Stress(mesh_, solverType_, nameStress_) /* TODO: Add any other arguments here */
         );
         DEBUG(adapterInfo("Added writer: Stress."));
     }
@@ -212,7 +216,7 @@ bool preciceAdapter::FSI::FluidStructureInteraction::addReaders(const preciceAda
     {
         interface->addCouplingDataReader(
             fieldConfig,
-            new Stress(mesh_, solverType_) /* TODO: Add any other arguments here */
+            new Stress(mesh_, solverType_, nameStress_) /* TODO: Add any other arguments here */
         );
         DEBUG(adapterInfo("Added reader: Stress."));
     }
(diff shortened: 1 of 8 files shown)


philipcardiff:
Pinging @MakisH in case you are not pinged by default (I suspect you are).

Issue #153, opened 2021-03-24
Duplicate coupling nodes at processor interfaces


DavidSCN:
When running cases in parallel and a distributed interface, we currently have duplicated definition of interface nodes at processor boundaries across the coupling interface. The duplication occurs for the location `faceNodes` as well as `faceCenters`. As a consequence, we cannot compute any RBF mappings in cases where the mapping is generated from the (distributed) OpenFOAM side (we had similar issues in our quasi 2D mappings). However, for `faceNodes` this won't be relevant at the moment since we only map `consistent` data across the  `faceNodes` interface and the mapping needs to be executed on the other participant. As a remedy, one could choose a nearest-neighbor mapping (I guess nearest-projection works as well) or avoid the partitioning of the interface in OpenFOAM. Resolving the issue here will probably fix #40 as well/ the error is probably the same. 

![Screenshot from 2021-03-24 13-28-06](https://user-images.githubusercontent.com/33414590/112311579-251c3f00-8ca6-11eb-8559-e23760df3efb.png)
![Screenshot from 2021-03-24 13-27-48](https://user-images.githubusercontent.com/33414590/112311583-25b4d580-8ca6-11eb-8374-c6e72848e51f.png)
![Screenshot from 2021-03-24 13-30-16](https://user-images.githubusercontent.com/33414590/112311568-23527b80-8ca6-11eb-8a98-a42391418c66.png)

DavidSCN:
I was actually too fast here. The issue occurs only for the location `faceNodes` and I visualized the wrong meshes here (both refer to the `faceNodes`). In the end, I was able to successfully run my case in parallel. @MakisH should we keep it here as a low priority issue or directly close it. As described above, we currently don't have any affected cases for `faceNodes`.

MakisH:
That was also my expectation and I am not sure if we should consider this a preCICE, adapter, or OpenFOAM issue.

Could you please add a note in the adapter documentation? I guess it does make sense for 3D cases (e.g. a valve), where the domain boundary cuts the interface. I would expect this to be a relatively common issue in practice.

DavidSCN:
> Could you please add a note in the adapter documentation? I guess it does make sense for 3D cases (e.g. a valve), where the domain boundary cuts the interface. I would expect this to be a relatively common issue in practice.

[...]

PR #418, opened 2026-08-28
Support writing DisplacementDelta in solid participants

philipcardiff:
## Summary

`DisplacementDelta::write()` was a stub that raised an adapter error (*"Writing displacementDelta is not supported."*), so a `solverType solid` participant could only write `Displacement`. This blocked coupling against fluid participants that read `DisplacementDelta` — for example the `elastic-tube-3d` tutorial.

This PR implements `write()` following the structure of `Displacement::write()`, writing the displacement increment over the current coupling time window:

* `faceCenters`: `D - D.oldTime()` on the interface patches;
* `faceNodes`: the same difference on the point field, keeping the existing #153 parallel-`faceNodes` warning.

The reference state is the stored old-time field rather than a class-owned snapshot. This is safe under implicit coupling because the adapter already checkpoints and restores `oldTime()` and `oldTime().oldTime()` for `volVectorField`s (`Adapter.C`, `readCheckpoint`/`writeCheckpoint`), so re-iterating a time window does not shift the reference. At `t = 0` the increment is legitimately zero. It also avoids adding a "time-window complete" hook to `CouplingDataUser`, which has none today.

`docs/config.md` now lists `DisplacementDelta` under solid `writeData`.

## Validation

Tested with OpenFOAM v2606, preCICE 3.4.1, solids4foam (development).

**1. A/B against a known-good baseline.** `perpendicular-flap` with a solids4foam solid was run twice: once unmodified (solid writes `Displacement`), once with the config changed so the solid writes `DisplacementDelta` instead. Everything else is identical. The cumulative sum of the written increments must reproduce the `Displacement` trajectory:

| t [s] | `Displacement` | `cumsum(DisplacementDelta)` | diff |
|---|---|---|---|
| 0.10 | +1.83615979e-02 | +1.83600145e-02 | 1.6e-06 |
| 1.00 | +7.38292638e-02 | +7.38424688e-02 | 1.3e-05 |
| 3.00 | +2.71168093e-02 | +2.71251167e-02 | 8.3e-06 |
| 5.00 | +1.39288688e-01 | +1.39275641e-01 | 1.3e-05 |

Over the full 500-window run the maximum absolute difference is **2.29e-05, i.e. 1.4e-04 relative to the peak amplitude** — the size of the coupling convergence tolerance (`5e-3` relative measure with IQN-ILS), not an error in the increment. Peak amplitudes: 0.1662942 vs 0.1663001.

**2. `elastic-tube-3d`.** A solids4foam solid participant (annulus, r = 5→6 mm, L = 50 mm, E = 300 kPa, nu = 0.3, rho = 1200, matching the CalculiX case) was run for the full 100 time windows against `fluid-openfoam`, converging in ~4 coupling iterations per window. The tube-midpoint `DisplacementDelta` matches the published CalculiX/FEniCS reference plot in both shape and magnitude:

| component | this run | reference plot |
|---|---|---|
| axial | dip to -2.1e-06 at t ≈ 0.004, rise to +3.0e-06 at t ≈ 0.007 | dip to ≈ -2.3e-06, rise to ≈ +3.0e-06 |
| radial | peak +3.9e-06 at t ≈ 0.008, then sharp drop | peak ≈ +3.7e-06, then sharp drop |
| circumferential | ≈ 0 (max magnitude 4.6e-07) | ≈ 0 |

Note that in that tutorial the face-centred OpenFOAM interface mesh carries no connectivity, so preCICE reports `3D Mesh "Solid-Mesh" does not contain triangles` and the two `nearest-projection` mappings fall back to nearest-neighbour. The agreement above is with that fallback in effect.

A follow-up tutorials PR adds the `elastic-tube-3d/solid-solids4foam` case used here.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SBFTMffUN7qDigrDQrPVb2

### DIFF
--- a/FSI/DisplacementDelta.C
+++ b/FSI/DisplacementDelta.C
@@ -36,13 +36,62 @@ void preciceAdapter::FSI::DisplacementDelta::initialize()
 
 std::size_t preciceAdapter::FSI::DisplacementDelta::write(double* buffer, bool meshConnectivity, const unsigned int dim)
 {
-    /* TODO: Implement
-     * We need two nested for-loops for each patch,
-     * the outer for the locations and the inner for the dimensions.
-     * See the preCICE writeBlockVectorData() implementation.
-     */
-    adapterInfo("Writing displacementDelta is not supported.", "error");
-    return 0;
+    // Copy the displacement increment over the current time window from
+    // OpenFOAM to the buffer. The reference state is the stored old-time
+    // field: the adapter checkpoints oldTime() (see Adapter::readCheckpoint),
+    // so re-iterating an implicit time window does not shift the reference.
+
+    int bufferIndex = 0;
+    if (this->locationType_ == LocationType::faceCenters)
+    {
+        // For every boundary patch of the interface
+        for (const label patchID : patchIDs_)
+        {
+            const vectorField& displacement =
+                cellDisplacement_->boundaryField()[patchID];
+            const vectorField& displacementOld =
+                cellDisplacement_->oldTime().boundaryField()[patchID];
+
+            // Write the displacement increment to the preCICE buffer
+            // For every cell of the patch
+            forAll(displacement, i)
+            {
+                for (unsigned int d = 0; d < dim; ++d)
+                    buffer[bufferIndex++] =
+                        displacement[i][d] - displacementOld[i][d];
+            }
+        }
+    }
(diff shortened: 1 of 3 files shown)


philipcardiff:
Pinging @MakisH in case you are not pinged by default (I suspect you are).

PR #419, opened 2026-09-01
docs/README.md: Move disclaimer up to fix footnotes

MakisH:
Fixes a rendering issue on the website:

Screenshot from 2026-09-01 10-23-30

Extracted from #392, to not hold this back any longer.


MakisH:
This was not enough, and footnotes are really rendered as footnotes in the layout used by the upcoming Hugo port. Let's not mix them up. Replaced by a list in https://github.com/precice/openfoam-adapter/commit/f9683ce7f8fa3f8915f4f7fdc53b92a74c7839e0

